PR #8, opened 2025-09-03
add google drive file/folder selector

mfortman11:


src/main.py
```diff
@@ -589,6 +589,17 @@ async def create_app():
...
+            "/connectors/{connector_type}/token",
+            require_auth(services["session_manager"])(
+                partial(
+                    connectors.connector_token,
+                    connector_service=services["connector_service"],
+                    session_manager=services["session_manager"],
+                )
+            ),
+            methods=["GET"],
+        ),
```
phact, line 602: Hm do we really want access tokens in the frontend?

PR #9, opened 2025-09-04
Add ingestion flow and ingestion integration with Langflow

ogabrielluiz:
This commit modifies both docker-compose.yml and docker-compose-cpu.yml to update the environment variable from FLOW_ID to LANGFLOW_CHAT_FLOW_ID, ensuring consistency across configurations. These changes contribute to a more robust and well-documented codebase.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "openrag"
-version = "0.1.1"
+version = "0.1.2"
 description = "Add your description here"
 readme = "README.md"
 requires-python = ">=3.13"


Makefile
```diff
@@ -0,0 +1,210 @@
...
+	@echo "🧹 Cleaning up containers and volumes..."
+	docker-compose down -v --remove-orphans
+	docker-compose -f docker-compose-cpu.yml down -v --remove-orphans 2>/dev/null || true
+	docker system prune -f
+
+# Local development
+backend:
+	@echo "🐍 Starting backend locally..."
+	@if [ ! -f .env ]; then echo "⚠️  .env file not found. Copy .env.example to .env first"; exit 1; fi
+	cd src && uv run python main.py
```
edwinjosechittilappilly: for me this gave an error! earlier that util.py is missing we might need to set the pythonpath

ogabrielluiz: That may be the ingestion_flow that was missing (was ignored because it is a json file)

docker-compose.yml
```diff
@@ -97,7 +97,8 @@ services:
       - LANGFLOW_SECRET_KEY=${LANGFLOW_SECRET_KEY}
       - JWT="dummy"
       - OPENRAG-QUERY-FILTER="{}"
-      - LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT=JWT,OPENRAG-QUERY-FILTER
+      - OPENSEARCH_PASSWORD=${OPENSEARCH_PASSWORD}
```
phact, line 102: Langflow doesn't need the opensearch admin password. Users should not have access to these creds.

src/services/langflow_file_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+    async def _get_api_key(self) -> Optional[str]:
+        """Get Langflow API key, ensuring it's generated if needed"""
+        from config.settings import generate_langflow_api_key
+
+        api_key = await generate_langflow_api_key()
+        print(f"[LF] _get_api_key returning: {'present' if api_key else 'None'}")
+        if api_key:
+            print(f"[LF] API key prefix: {api_key[:8]}...")
+        return api_key
```
phact: why not use LANGFLOW_KEY set in config.settings?

src/services/langflow_file_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+    async def _headers(self, extra: Optional[Dict[str, str]] = None) -> Dict[str, str]:
+        api_key = await self._get_api_key()
+        headers = {"x-api-key": api_key} if api_key else {}
+        if extra:
+            headers.update(extra)
+        return headers
```
phact: We should figure out a central place to do this rather than a central place in the code base.

src/services/langflow_file_service.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        # NOTE: base_url points to /api/v1; v2 endpoints must not be prefixed with /api/v1
+        url = f"{LANGFLOW_URL}/api/v2/files"
+        api_key = await self._get_api_key()
+        self.logger.debug("[LF] Upload (v2) -> %s (key_present=%s)", url, bool(api_key))
+        if api_key:
+            self.logger.debug(f"[LF] Using API key: {api_key[:12]}...")
+        else:
+            self.logger.error("[LF] No API key available for upload!")
+        async with httpx.AsyncClient(timeout=60.0) as client:
```
phact: maybe we can have a langflow _http client in settings that everyone can use, the same way we handle all the other clients.

docker-compose.yml
ogabrielluiz: If the component in Langflow doesn't have that I get a 401 when building it.

src/services/langflow_file_service.py
```diff
@@ -0,0 +1,153 @@
...
+            files = {"file": file_tuple}
+            headers = await self._headers()
+            print(f"[LF] Upload headers: {headers}")
+            # Note: jwt_token is for OpenSearch, not for Langflow API - only use x-api-key
```
phact: but don't you need the user jwt token so you know what user to authenticate with to open search?

src/services/langflow_file_service.py
ogabrielluiz: I'll refactor this a bit but the idea is to use a safer way to get it just so we don't accidentally get it and it hasn't been generated yet.

src/services/langflow_file_service.py
ogabrielluiz: Sounds good.

src/services/langflow_file_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+        url = f"{self.base_url}/run/{self.flow_id_ingest}"
+
+        payload: Dict[str, Any] = {
+            "input_value": "Ingest files",
+            "input_type": "chat",
+            "output_type": "text",  # Changed from "json" to "text"
+        }
+
+        # Pass files via tweaks to File component (File-PSU37 from the flow)
```
phact, line 87: we should come up with a cleaner API that doesn't require passing component IDs. What if we added support for openai's file API? 

https://platform.openai.com/docs/api-reference/files/create

src/services/langflow_file_service.py
```diff
@@ -0,0 +1,153 @@
...
+            list(tweaks.keys()) if isinstance(tweaks, dict) else None,
+            bool(jwt_token),
+        )
+
+        # Log the full payload for debugging
+        self.logger.debug("[LF] Request payload: %s", payload)
+
+        extra_headers = {}
+        # Note: Ingestion flow doesn't need JWT authentication context
+        # Removed X-LANGFLOW-GLOBAL-VAR-JWT header
```
phact: the point about langflow needing the user jwt applies for all of these calls.

src/main.py
```diff
@@ -317,13 +334,18 @@ async def initialize_services():
                 "Failed to load persisted connections on startup", error=str(e)
             )
     else:
-        logger.info("Skipping connector loading in no-auth mode")
+        print("[CONNECTORS] Skipping connection loading in no-auth mode")
```
phact: use logger

.env.example
```diff
@@ -1,9 +1,11 @@
-# flow id from the the openrag flow json
-FLOW_ID=1098eea1-6649-4e1d-aed1-b77249fb8dd0
-# Set a strong admin password for OpenSearch; a bcrypt hash is generated at
-# container startup from this value. Do not commit real secrets.
+# make one like so https://docs.langflow.org/api-keys-and-authentication#langflow-secret-key
+LANGFLOW_SECRET_KEY=
+# flow ids for chat and ingestion flows
+LANGFLOW_CHAT_FLOW_ID=1098eea1-6649-4e1d-aed1-b77249fb8dd0
+LANGFLOW_INGEST_FLOW_ID=5488df7c-b93f-4f87-a446-b67028bc0813
+# must match the hashed password in secureconfig, must change for secure deployment!!!
```
phact, line 11: We'll have to propagate these to the TUI

docker-compose-cpu.yml
```diff
@@ -96,7 +96,8 @@ services:
       - LANGFLOW_SECRET_KEY=${LANGFLOW_SECRET_KEY}
       - JWT="dummy"
       - OPENRAG-QUERY-FILTER="{}"
-      - LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT=JWT,OPENRAG-QUERY-FILTER
+      - OPENSEARCH_PASSWORD=${OPENSEARCH_PASSWORD}
```
phact, line 102: Remove this

src/services/langflow_file_service.py
ogabrielluiz: We can achieve that but it is going to be limited and hacky in Langflow. It won't work if there's a ChatInput and a File component in the same flow, for exmaple.

src/services/langflow_file_service.py
ogabrielluiz: We are using the API key for that, no?

docker-compose.yml
phact: yeah you need to pass the jwt

src/services/langflow_file_service.py
phact: thanks, I think there's a way to ensure the key is there already.

src/services/langflow_file_service.py
phact: the API key is for langflow, the jwt is for open search user auth (document level security)

src/services/langflow_file_service.py
phact: if you look at the responses endpoint in langflow it has validation for chatinput and throws error if there isn't one.  we could do something similar.

The main issue IMO is how do we pass the flow ID.

ogabrielluiz: I think the files are global so all we need is the user id

src/services/langflow_file_service.py
ogabrielluiz: Done

PR #19, opened 2025-09-05
Fix: Persist Function Call Data Across Conversation Switching

Cristhianzl:
Function calls (tool calls) were not persisting when users switched between chat conversations. The blue "Function Call" UI components would disappear when navigating away from a conversation and returning to it, even though the function calls had been executed successfully during the original chat session.

Root Cause
The issue occurred at multiple levels in the data pipeline:

  1. Backend Storage: Function call metadata was only being stored as text content, losing structured tool call information
  2. Backend Retrieval: History APIs were filtering out function call data fields during conversation retrieval
  3. Frontend Processing: Conversation loading logic wasn't extracting function call data from backend responses

https://www.loom.com/share/a2119d78a86e4f88bc8c8c820a5ee7b6?sid=39250c2c-0f97-46d4-8a7c-e11e586a8cbc


src/services/langflow_history_service.py
```diff
@@ -0,0 +1,205 @@
...
+                )
+                
+                if response.status_code == 200:
+                    result = response.json()
+                    self.auth_token = result.get('access_token')
+                    print(f"Successfully authenticated with Langflow for history retrieval")
+                    return self.auth_token
+                else:
+                    print(f"Langflow authentication failed: {response.status_code}")
+                    return None
```
phact, line 48: is this actually needed?

src/services/langflow_history_service.py
```diff
@@ -0,0 +1,205 @@
...
+            return []
+            
+        try:
+            headers = {"Authorization": f"Bearer {token}"}
+            params = {}
+            
+            if flow_id:
+                params["flow_id"] = flow_id
+                
+            async with httpx.AsyncClient() as client:
```
phact, line 70: gabriel is making a shared langflow http client session. Let's switch to that in a future PR.

src/services/conversation_persistence_service.py
```diff
@@ -0,0 +1,126 @@
...
+"""
+
+import json
+import os
+from typing import Dict, Any
+from datetime import datetime
+import threading
+
+
+class ConversationPersistenceService:
```
phact, line 13: Why do we save the conversations themselves to disk if we can get them from langflow?

Cristhianzl: that's redundant. I will fix it.

Cristhianzl: i do need to track at least the metadata of the messages in that session to organize between the users.
and not store everything again in the memory.

src/services/langflow_history_service.py
Cristhianzl: yeah the monitor endpoint needs a jwt auth..

PR #39, opened 2025-09-12
tui docling serve button

phact:


pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "python-dotenv>=1.0.0",
     "textual-fspicker>=0.6.0",
     "structlog>=25.4.0",
+    "docling-serve>=0.16.1",
```
dolfim-ibm: let's pin the v1 API specs. (assuming this PR is used https://github.com/langflow-ai/langflow/pull/9702)

```suggestion
    "docling-serve>=1.4.1",
```

dolfim-ibm: The `docling` entry should also be changed to

```
docling[ocrmac,vlm]
```

PR #54, opened 2025-09-19
Move knowledge filters list to left nav

colegoldsmith:
Screenshot 2025-09-19 at 3 23 51 PM

### DIFF
--- a/frontend/components/knowledge-actions-dropdown.tsx
+++ b/frontend/components/knowledge-actions-dropdown.tsx
@@ -12,6 +12,7 @@ import { Button } from "./ui/button";
 import { DeleteConfirmationDialog } from "./confirmation-dialog";
 import { useDeleteDocument } from "@/app/api/mutations/useDeleteDocument";
 import { toast } from "sonner";
+import { useRouter } from "next/navigation";
 
 interface KnowledgeActionsDropdownProps {
   filename: string;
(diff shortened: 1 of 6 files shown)


frontend/src/app/api/[...path]/route.ts
Comment (colegoldsmith) [inline frontend/src/app/api/[...path]/route.ts:1]: The changes here fix an issue with proxying DELETE requests.

src/services/knowledge_filter_service.py
colegoldsmith, line 1: adding `refresh="wait_for"` to various create, delete, update calls to the opensearch client. This is 100% AI helping me out, i've got no idea what I'm doing here. Please tell me if this is dumb 🙏

zznate: This flag blocks the response until the index update comes back with an ACK that it's been applied (any CRUD op). Think of it like a database transaction, but without the strong guarantee of concurrency, more just an "i applied your op, not really sure what everyone else is doing". 
It can take a couple of seconds for this op to complete on a reasonably busy system, so weigh that in with the user experience you want.

colegoldsmith: Ok thank you for the info! I needed these ops to complete so the UI would update immediately/be in sync after CRUD actions on filters. I'm guessing these won't be created super often so it's probably fine to have these calls be a little bit slow sometimes. If it does become a bottleneck we could try implementing a client side state lib to allow the UI to reflect actions immediately and have some way of keeping that in sync with the server state every so often, not sure if that's a good move or not

PR #61, opened 2025-09-23
docs: install python wheel or docker compose

mendonk:
[Preview](https://d5rxiv0do0q3v.cloudfront.net/langflow-drafts/docs-installation/index.html)
Add `install.mdx` page under `docs/docs/get-started/` with instructions for installing OpenRAG using either the Python wheel (with TUI) or clone repo+Docker Compose, including setup steps, environment variables, and explanations for both basic and advanced configurations.


docs/docs/get-started/install.mdx
```diff
@@ -0,0 +1,159 @@
...
+- [Docker Compose](https://docs.docker.com/compose/install/) installed
+- For GPU support: (TBD)
+
+1. Clone the OpenRAG repository.
+   ```bash
+   git clone https://github.com/langflow-ai/openrag.git
+   cd openrag
+   ```
+
+2. Create a `.env` file in the repository root.
```
zznate: This should probably be from the project root:`cp .env.example .env` as the example env file has a bunch of useful comments and a couple of defaults.

mendonk: Good call @zznate , this is now added

PR #187, opened 2025-10-03
chore: fix typo and clarify comments in .env.example

dvncvn:
- Fixed a typo in comments
- Clarified instructions
- Added spacing for better readability
- No variable values changed

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,7 +20,7 @@ NUDGES_FLOW_ID=ebc01d31-1976-46ce-a385-b0240327226c
 # Set a strong admin password for OpenSearch.
 # A bcrypt hash is generated at container startup from this value.
 # Do not commit real secrets.
-# Must match the hashed password in secureconfig. Must be changed for secure deployments.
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -1,42 +1,60 @@
...
 
-# Set a strong admin password for OpenSearch; a bcrypt hash is generated at
-# container startup from this value. Do not commit real secrets.
-# must match the hashed password in secureconfig, must change for secure deployment!!!
+
+# OpenSearch Auth
+# Set a strong admin password for OpenSearch.
+# A bcrypt hash is generated at container startup from this value.
+# Do not commit real secrets.
+# Must match the hashed password in secureconfig. Must be changed for secure deployments.
```
phact: > # Must be changed for secure deployments.

we actually do this bit automatically now so the user doesn't have to worry about secure config at this point. Let's delete this sentence.

phact:
```suggestion
# Must be changed for secure deployments.
```

PR #194, opened 2025-10-03
fix: make tasks be displayed after failing, solve knowledge duplication issues

lucaseduoli:
This pull request introduces duplicate file handling to the document upload workflow in the `KnowledgeDropdown` component, improving user experience and data integrity. When a user attempts to upload a file with a name that already exists, a dialog prompts them to confirm overwriting the existing document. The implementation also refactors related code to support this new flow and ensures that UI state and search caches are updated appropriately.

**Duplicate file handling and upload workflow improvements:**

* Added a new `DuplicateHandlingDialog` component to prompt users when uploading a file with a duplicate name, allowing them to confirm overwriting the existing document.
* Updated the file upload logic in `knowledge-dropdown.tsx` to check for duplicate filenames before uploading, show the duplicate dialog if needed, and handle the overwrite action by updating search caches and proceeding with the upload. [[1]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40R42-R57) [[2]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40L169-R246) [[3]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40L251-R343) [[4]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40R686-R693)
* Integrated `react-query`'s `useQueryClient` to remove old file entries from the search cache before overwriting, ensuring UI consistency after a file is replaced. [[1]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40R30) [[2]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40L251-R343)
* Refactored folder and S3 upload flows to use `refetchTasks()` for refreshing UI state after successful uploads, replacing previous event-based refreshes. [[1]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40L304-R378) [[2]](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40L357-R421)
* Minor code quality improvements, such as adding missing `type="button"` attributes and updating key props for better React list rendering. ([frontend/components/knowledge-dropdown.tsxR502](diffhunk://#diff-62ce844c2ab79fff41954cc3d4b66de45ef3af6c7c6b4b237c0f5a73d9d28d40R502), F7fe8425L576R624)

These changes collectively improve document upload reliability and user experience, especially when handling files with duplicate names.

### DIFF
--- a/frontend/src/app/api/queries/useGetSearchQuery.ts
+++ b/frontend/src/app/api/queries/useGetSearchQuery.ts
@@ -29,6 +29,7 @@ export interface ChunkResult {
   owner_email?: string;
   file_size?: number;
   connector_type?: string;
+  index?: number;
 }
 
 export interface File {
@@ -55,7 +56,7 @@ export interface File {
 export const useGetSearchQuery = (
   query: string,
   queryData?: ParsedQueryData | null,
-  options?: Omit<UseQueryOptions, "queryKey" | "queryFn">,
+  options?: Omit<UseQueryOptions, "queryKey" | "queryFn">
 ) => {
   const queryClient = useQueryClient();
 
@@ -184,7 +185,7 @@ export const useGetSearchQuery = (
       queryFn: getFiles,
       ...options,
     },
-    queryClient,
+    queryClient
   );
 
   return queryResult;
--- a/frontend/src/app/knowledge/chunks/page.tsx
+++ b/frontend/src/app/knowledge/chunks/page.tsx
@@ -1,8 +1,11 @@
 "use client";
 
-import { ArrowLeft, Check, Copy, Loader2, Search } from "lucide-react";
+import { ArrowLeft, Check, Copy, Loader2, Search, X } from "lucide-react";
 import { useRouter, useSearchParams } from "next/navigation";
 import { Suspense, useCallback, useEffect, useMemo, useState } from "react";
+// import { Label } from "@/components/ui/label";
+// import { Checkbox } from "@/components/ui/checkbox";
+import { filterAccentClasses } from "@/components/knowledge-filter-panel";
 import { ProtectedRoute } from "@/components/protected-route";
 import { Button } from "@/components/ui/button";
 import { Checkbox } from "@/components/ui/checkbox";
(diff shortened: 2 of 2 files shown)


src/services/task_service.py
```diff
@@ -82,18 +84,31 @@ async def create_langflow_upload_task(
...
         """Create a new task with custom processor for any type of items"""
+        import os
         # Store anonymous tasks under a stable key so they can be retrieved later
         store_user_id = user_id or AnonymousUser().user_id
         task_id = str(uuid.uuid4())
+
+        # Create file tasks with original filenames if provided
+        file_tasks = {}
+        for i, item in enumerate(items):
+            if original_filenames and i < len(original_filenames):
```
edwinjosechittilappilly: should we validate this check? or should the original file name be key value pairs?

src/models/processors.py
```diff
@@ -55,6 +55,108 @@ async def check_document_exists(
...
+        max_retries = 3
+        retry_delay = 1.0
+
+        for attempt in range(max_retries):
+            try:
+                # Search for any document with this exact filename
+                search_body = {
+                    "query": {
+                        "term": {
+                            "filename.keyword": filename
```
zznate: This is different functionality than in `documents.py` above. Let's encapsulate the `search_body` construction for a filename in one place so it consistently uses `filename` and `filename.keyword` (it looks like that is what we want?) for any operation against the index.

src/models/processors.py
```diff
@@ -55,6 +55,108 @@ async def check_document_exists(
...
+        Delete all chunks of a document with the given filename from OpenSearch.
+        """
+        from config.settings import INDEX_NAME
+
+        try:
+            # Delete all documents with this filename
+            delete_body = {
+                "query": {
+                    "term": {
+                        "filename.keyword": filename
```
zznate: same as above. encapsulate the query construction for consistency.

src/api/documents.py
```diff
@@ -6,14 +6,73 @@
...
+            user.user_id, jwt_token
+        )
+
+        # Search for any document with this exact filename
+        # Try both .keyword (exact match) and regular field (analyzed match)
+        search_body = {
+            "query": {
+                "bool": {
+                    "should": [
+                        {"term": {"filename.keyword": filename}},
```
zznate: see comment below about consistent naming.

zznate: Just checked `process_upload_document` in `document_service.py` and it doesnt look like we are capturing `filename.keyword` as part of the index data. Is there another place docs (in the OpenSearch sense) get written to the index?

PR #307, opened 2025-10-24
Feat/278 Edit current model provider settings

colegoldsmith:
Screenshot 2025-10-24 at 4 54 25 PM
Screenshot 2025-10-24 at 4 54 37 PM

### DIFF
--- a/frontend/src/app/settings/components/model-selectors.tsx
+++ b/frontend/src/app/settings/components/model-selectors.tsx
@@ -78,7 +78,7 @@ export function ModelSelectors({
             )}
           />
         </LabelWrapper>
-        {errors[embeddingModelName] && (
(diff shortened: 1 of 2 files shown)


src/api/settings.py
```diff
@@ -396,22 +402,117 @@ async def update_settings(request, session_manager):
                 # Don't fail the entire settings update if flow update fails
                 # The config will still be saved
 
+        # Update provider settings
```
colegoldsmith, line 405: I will likely remove these changes to the /settings POST in favor of re-using the /onboarding POST as long as I can verify there are no issues with that.

lucaseduoli: You can keep them in /settings

colegoldsmith: 👍

PR #329, opened 2025-10-29
docs: clarify installation steps

mendonk:
1. Only include uvx option in README.md for a quickstart. Clarify the difference between uvx and uv add.
2. Remove outdated "Python wheel" language.
3. Remove uv sync step because uv add does this.
4. Added a troubleshooting entry for missing package and uvx/uv venv issue fix.

### DIFF
--- a/src/services/models_service.py
+++ b/src/services/models_service.py
@@ -64,7 +64,7 @@ class ModelsService:
                             {
                                 "value": model_id,
                                 "label": model_id,
-                                "default": model_id == "gpt-5",
+                                "default": model_id == "gpt-4o",
                             }
                         )
 


docs/docs/support/troubleshoot.mdx
```diff
@@ -38,6 +38,33 @@ This example increases the machine size to 8 GB of RAM, which should be sufficie
 
 Ensure ports 3000, 7860, 8000, 9200, 5601 are available.
 
+## OCR ingestion fails (easyocr not installed)
+
+If Docling ingestion fails with an OCR-related error and mentions `easyocr` is missing, install `easyocr` in your OpenRAG environment.
```
phact: why would we see this? do we just need to add that dependency in the project? doesn' t make sense to make the user do it.

mendonk: Specifically for uvx cache - i added more info to explain

PR #366, opened 2025-11-06
dont show banner when down

phact:


colegoldsmith (approved):
Looks good! I pulled the branch and it seems to work well locally

PR #399, opened 2025-11-12
docs: use installer script in readme

mendonk:
When the repo is public again, this is the fastest method of getting started.


README.md
```diff
@@ -26,13 +26,26 @@ OpenRAG is a comprehensive Retrieval-Augmented Generation platform that enables
 
 ## Quickstart
```
aimurphy, line 27: Can we use this option on the main quickstart?

Also, isn't Python still a prerequisite?

mendonk: uv/uvx will check for Python and install it [uv docs](https://docs.astral.sh/uv/guides/install-python/#automatic-python-downloads)
But i added a note that this happens, and that it's better to manage your own

PR #662, opened 2025-12-11
fix: pass .env to commands being executed to always match new changes

### LINKED ISSUE #659: TUI: Docker pulls wrong image version when starting for the first time after an update
When starting the TUI for the first time, it sets OPENRAG_VERSION correctly with the TUI version, but doesn't refresh the environment, so the docker commands called still have the old version, so the old version is pulled.

lucaseduoli:
This pull request improves how environment variables are handled when running Docker Compose commands in the `container_manager.py` module. The main enhancement is ensuring that values from the `.env` file always take precedence over the current process environment, so Docker Compose commands use the most up-to-date configuration. The changes also standardize this behavior across all relevant async command execution methods.

**Environment variable handling improvements:**

* Added a new `_get_env_from_file` method that reads environment variables from the `.env` file, overriding any existing values from `os.environ`, and uses a sanitizer for values. This ensures Docker Compose commands always use the latest `.env` values.
* Updated all async methods that run Docker Compose commands (`_run_compose_command`, `_run_compose_command_streaming`, `_stream_compose_command`, and `follow_service_logs`) to use the new environment loading logic, so `.env` changes are respected everywhere. [[1]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eR277-R285) [[2]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eR323-R331) [[3]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eR396-R404) [[4]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eR1153-R1161)

**Code cleanup and imports:**

* Removed unnecessary or redundant imports of `re` and `subprocess` in favor of module-level imports and streamlined code. [[1]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eL5-R6) [[2]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eL156) [[3]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eL929)

Closes #659

### DIFF
--- a/src/tui/managers/env_manager.py
+++ b/src/tui/managers/env_manager.py
@@ -12,7 +12,6 @@ from dotenv import load_dotenv
 from utils.logging_config import get_logger
 
 from ..utils.validation import (
-    sanitize_env_value,
     validate_documents_paths,
     validate_google_oauth_client_id,
     validate_non_empty,
@@ -548,17 +547,12 @@ class EnvManager:
             
             # Check if OPENRAG_VERSION is already set in .env
             if self.env_file.exists():
-                env_content = self.env_file.read_text()
-                if "OPENRAG_VERSION" in env_content:
-                    # Already set, check if it needs updating
-                    for line in env_content.splitlines():
-                        if line.strip().startswith("OPENRAG_VERSION"):
-                            existing_value = line.split("=", 1)[1].strip()
-                            existing_value = sanitize_env_value(existing_value)
-                            if existing_value == current_version:
-                                # Already correct, no update needed
-                                return
-                            break
+                # Load .env file using load_dotenv
+                load_dotenv(dotenv_path=self.env_file, override=False)
+                existing_value = os.environ.get("OPENRAG_VERSION", "")
+                if existing_value and existing_value == current_version:
+                    # Already correct, no update needed
+                    return
             
             # Set or update OPENRAG_VERSION
             self.config.openrag_version = current_version
--- a/src/tui/utils/validation.py
+++ b/src/tui/utils/validation.py
@@ -96,21 +96,6 @@ def validate_non_empty(value: str) -> bool:
     return bool(value and value.strip())
 
 
-def sanitize_env_value(value: str) -> str:
-    """Sanitize environment variable value."""
-    # Remove leading/trailing whitespace
-    value = value.strip()
-
-    # Remove quotes if they wrap the entire value
-    if len(value) >= 2:
-        if (value.startswith('"') and value.endswith('"')) or (
-            value.startswith("'") and value.endswith("'")
-        ):
-            value = value[1:-1]
-
-    return value
-
-
 def validate_documents_paths(paths_str: str) -> tuple[bool, str, list[str]]:
     """
     Validate comma-separated documents paths for volume mounting.


src/tui/managers/container_manager.py
```diff
@@ -121,6 +122,35 @@ def _find_compose_file(self, filename: str) -> Path:
...
+                from ..utils.validation import sanitize_env_value
+                with open(env_file, "r") as f:
+                    for line in f:
+                        line = line.strip()
+                        if not line or line.startswith("#"):
+                            continue
+                        if "=" in line:
+                            key, value = line.split("=", 1)
+                            key = key.strip()
+                            value = sanitize_env_value(value)
```
edwinjosechittilappilly: we might need to improve this parsing for standard scenario.
also cant we use load dotnenv from python dotenv?

what are your thoughts on it?

phact: hm we already use dotenv in the tui here https://github.com/langflow-ai/openrag/blob/a78926dd3761c8abef823a3204df6f388d8f7a14/src/tui/screens/welcome.py#L44

phact: maybe what you're looking for is load_dotenv(override=True) ?

src/tui/managers/env_manager.py
```diff
@@ -1,16 +1,16 @@
...
 from pathlib import Path
 from typing import Dict, List, Optional
 
+from dotenv import load_dotenv
 from utils.logging_config import get_logger
 
-logger = get_logger(__name__)
-
 from ..utils.validation import (
     sanitize_env_value,
```
phact: do we still do this sanitize thing? Was it doing anything different than what load_dotenv does? Would be good to test w/ different quoting examples.

if we don't need it let's remove the import and the function all together.

src/tui/screens/welcome.py
```diff
@@ -41,7 +41,8 @@ def __init__(self):
         self.has_env_file = self.env_manager.env_file.exists()
 
         # Load .env file if it exists
-        load_dotenv()
+        # override=True ensures .env file values take precedence over existing environment variables
+        load_dotenv(override=True)
```
phact, line 45: I'm a bit confused on why we do this both in welcome AND env manager AND now in container manager.

lucaseduoli: It's necessary because the env manager updates the env variables, the container manager needs to load the env to pass it to the command so that the openrag version is updated. Only putting it at welcome wouldnt update the values if they're changed by the TUI after the welcome.

src/tui/managers/env_manager.py
lucaseduoli: load_dotenv does the same thing that we did with the sanitize. We dont need the sanitize function anymore

Issue #732, opened 2026-01-02
Knowledge-filter "create" command causes front-end application error


SonicDMG:
OpenRAG v0.2.0
While using and testing the knowledge-filter endpoints for my app I've run into what looks like a critical issue when using the "create" operation.

For context, I can get, update, and search just fine using the knowledge-filter endpoint with cURL. I have also confirmed that any changes I've made to existing filters created via the OpenRAG web app are, in fact, applied to my filter and reflect back in the webapp on refresh.

However, when I attempt to create a new filter I consistently get the following error:
<img width="1263" height="921" alt="Image" src="https://github.com/user-attachments/assets/5c4e39e5-528a-4417-8188-e2fe65ec7604" />

This issue completely breaks the "knowledge" section of front-end web app. I can thankfully still query data via the SDK and use the chat and settings pages. 
I tried a multitude of approaches to address the issue:
- Deleted local storage, cleared cache, refreshed
- Stopped and restarted all services
- Pruned all containers and images
- Wiped all containers, images, re-downloaded

Finally, the only way I could get the app running again was a full "Factory Reset".

Here's the cURL command used:
```bash
curl -X POST "http://localhost:3000/api/v1/knowledge-filters" \
  -H "Authorization: Bearer orag_4v514..." \
  -H "Content-Type: application/json" \
  -d '{
    "name": "Technical Docs",
    "description": "Filter for technical documentation",
    "queryData": "{\"query\": \"technical\", \"filters\": {\"document_types\": [\"application/pdf\"]}, \"limit\": 10, \"scoreThreshold\": 0.5}"
  }'
```

edwinjosechittilappilly:
Testing

PR #745, opened 2026-01-05
Fix harden sdk filters

phact:
Makes it so that:

1) sdk doesn't allow invalid query filters

and

2) ui doesn't blow up with the invalid query filters that were allowed by the sdk endpoint previously.


phact:
will fix https://github.com/langflow-ai/openrag/issues/732

PR #751, opened 2026-01-07
Update docker-compose.yml

ticlazau:
The backend service was running on port 8000 internally but had no port exposed to the host machine, making it inaccessible


phact:
yes, it's by design. why do you need the backed port exposed?

PR #753, opened 2026-01-07
Docs: Enable Algolia search

### LINKED ISSUE #519: Docs: Search functionality
We need to figure out how to enable search for the OpenRAG docs, either by extending Algolia coverage from Langflow or creating a new Algolia plan for OpenRAG.

If it's not possible to enable search in a reasonable amount of time, the pages need to be broken down granularly (one main idea per page). Typically, we group a few related ideas on one page (ex. _Manage widgets_ would cover all CRUD operations for widgets). This only works well when search is available so users can find niche subjects like troubleshooting or special configurations. Without search, we need to separate the pages so that all main ideas bubble up in the left nav (ex. create widgets, list widgets, edit widgets, delete widgets as 4 separate pages).

aimurphy:
Closes #519 

This PR adds the Algolia block to the Docusaurus config.

❓ I see Algolia dependencies in LF's [package-lock.json](https://github.com/langflow-ai/langflow/blob/main/docs/package-lock.json#L43) and yarn.lock but no where else. @mendonk - Do I need to do something to update package-lock.json and yarn.lock?

⚠️ Don't merge until I have configured the Algolia crawler; otherwise, the search will be useless.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -17,14 +17,6 @@ const config = {
   favicon: 'img/favicon.ico',
 
   headTags: [
-    // Algolia site verification meta tag
-    {
-      tagName: "meta",
-      attributes: {
-        name: "algolia-site-verification",
-        content: "424339D27FB7921F",
-      },
-    },
     ...(isProduction
       ? [
           // Google Consent Mode - Set defaults before Google tags load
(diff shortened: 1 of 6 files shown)


mendonk (approved):
Looks good, approving for when the index is hooked up

docs/docusaurus.config.js
```diff
@@ -244,6 +244,15 @@ const config = {
           },
         ],
       },
+      algolia: {
+        appId: "SMEA51Q5OL",
+        // public key, safe to commit
+        apiKey: "b2ec302e9880e8979ad6a68f0c36271e",
+        indexName: "openrag",
```
aimurphy:
```suggestion
        indexName: "openrag-algolia",
```

PR #806, opened 2026-01-19
feat: Add an option to delete an existing index in onboarding

matanor:
- To be used in automatic benchmarking, when it is desired to start from an empty index:
- added a new boolean option of `delete_existing_index` to the onboarding endpoint. 
- when the `delete_existing_index` option is set, the existing index is first deleted, and a new clean index is created


ricofurtado (approved):

lucaseduoli (approved):
LGTM

PR #886, opened 2026-01-30
feat: Connectors Update and stability improvements

edwinjosechittilappilly:
Frontend: add isDisconnecting state, handleDisconnect function and UI to allow disconnecting connectors (with Reconnect + Disconnect buttons), guard Add Knowledge visibility by connectionId, and disable conflicting actions while connecting/disconnecting. Backend: verify connector authentication in connector_status (reporting client_id and is_authenticated, and only treat connections as active if authenticated) to avoid stale active flags. Add new connector_disconnect endpoint to remove user connections, cleanup subscriptions, and return deleted count. App bootstrap: load .env early via python-dotenv and register the new DELETE /connectors/{connector_type}/disconnect route.


mfortman11:
image

PR #889, opened 2026-02-02
fix: allow overwrite during onboarding

phact:
We added overwrite check by filename in the UI [which I'm not sure is very useful to be honest] but it sometimes breaks onboarding flow so we should at least disable it for onboarding. 

TODO - look into how the overwrite check affects connector sync flows.


phact:
but it does affect sdk calls, fixing that here too

PR #900, opened 2026-02-05
Changes in ingestion flow to support docling chunking component.

ricofurtado:
Changes in default ingestion flow so it can support docling-based chunking.


flows/ingestion_flow.json
```diff
@@ -5734,19 +5205,301 @@
...
...AITokenizer(\n                    tokenizer=tiktoken.encoding_for_model(self.openai_model_name), max_tokens=max_tokens\n                )\n            chunker = HybridChunker(\n                tokenizer=tokenizer,\n            )\n        elif self.chunker == \"HierarchicalChunker\":\n            chunker = HierarchicalChunker()\n\n        results: list[Data] = []\n        try:\n            for doc in documents:\n                for chunk in chunker.chunk(dl_doc=doc):\n                    enriched_text = chunker.contextualize(chunk=chunk)\n                    meta = DocMeta.model_validate(chunk.meta)\n\n                    results.append(\n                        Data(\n                            data={\n                                \"text\": enriched_text,\n                                \"document_id\": f\"{doc.origin.binary_hash}\",\n                                \"doc_items\": json.dumps([item.self_ref for item in meta.doc_items]),\n                            }\n                        )\n                    )\n\n        except Exception as e:\n            msg = f\"Error splitting text: {e}\"\n            raise TypeError(msg) from e\n\n        return DataFrame(results)\n"
```
mpawlow: ### (a) [Normal] `chunker` variable unbound if dropdown value is unexpected

- In ChunkDoclingDocument.chunk_documents():

```python
  chunker: BaseChunker
  if self.chunker == "HybridChunker":
      ...
      chunker = HybridChunker(tokenizer=tokenizer)
  elif self.chunker == "HierarchicalChunker":
      chunker = HierarchicalChunker()
```
-  UnboundLocalError can occur if neither branch matched
```python
  results: list[Data] = []
  try:
      for doc in documents:
          for chunk in chunker.chunk(dl_doc=doc):   # <<<<<<<<<<<<<<<<
```

- If self.chunker has an unexpected value (e.g., future schema migration or corrupted flow state), chunker is unbound and will raise UnboundLocalError inside the try block. The try block only catches generic Exception, so this would propagate as an unhandled error. 
- Potential fix

```python
  else:
      msg = f"Unknown chunker: {self.chunker}"
      raise ValueError(msg)
```

mpawlow: ### (b) [Normal] tokenizer variable unbound in HybridChunker path

- Similar to (a)
- Within the HybridChunker branch, UnboundLocalError if provider is neither value

```python
  if self.provider == "Hugging Face":
      tokenizer = HuggingFaceTokenizer.from_pretrained(...)
  elif self.provider == "OpenAI":
      tokenizer = OpenAITokenizer(...)

  chunker = HybridChunker(tokenizer=tokenizer)  # <<<<<<<<<<<
```
- If self.provider has an unexpected value, tokenizer is unbound. 
- Potential Solution: Add an else branch with an error message

mpawlow: ### (c) [Normal] Dead code: _docs_to_data method in ChunkDoclingDocument

- The component defines a helper method that is never called
```python
  def _docs_to_data(self, docs) -> list[Data]:
      return [Data(text=doc.page_content, data=doc.metadata) for doc in docs]
```

mpawlow: ### (d) [Normal] Large value 128 * 1024 set as default max_tokens for OpenAI tokenizer

- i.e.
```python
  if max_tokens is None:
      max_tokens = 128 * 1024  # context window length required for OpenAI tokenizers
```

- A 128k-token default for chunk max size is enormous and would result in very few or no splits for most documents. 
- `Question`: Is this value indeed required? (according to the comment)
- Note: The HuggingFaceTokenizer default from sentence-transformers/all-MiniLM-L6-v2 is 256 tokens (which might be more appropriate for RAG)
- Recommend verifying the value against what Docling's OpenAITokenizer actually requires vs. what makes sense for retrieval.

flows/ingestion_flow.json
```diff
@@ -4292,7 +3763,7 @@
               "api_key": {
                 "_input_type": "SecretStrInput",
                 "advanced": false,
-                "display_name": "IBM watsonx.ai API Key",
+                "display_name": "OpenAI API Key",
```
mpawlow: ### (e) [Minor] DoclingRemote field label change are a breaking rename: "IBM watsonx.ai API Key" -> "OpenAI API Key"

- Note: This isn't directly authored in this PR (it's an upstream lfx component update), but the effect for any existing WatsonX users of OpenRAG is that their configured field names would mismatch. 
- Recommend checking if FlowsService needs to handle this migration.

flows/ingestion_flow.json
```diff
@@ -4261,13 +3732,13 @@
...
-                "value": "75fd27c1-8f4b-46a1-88bb-a8a8e72719e3"
+                "value": "15acea35-7c03-488d-8702-13327d5a4cea"
               },
               "_type": "Component",
               "api_base": {
                 "_input_type": "MessageTextInput",
-                "advanced": true,
-                "display_name": "API Base URL",
+                "advanced": false,
+                "display_name": "OpenAI API Base URL",
```
mpawlow: ### (f) [Minor] DoclingRemote field label change are a breaking rename: "API Base URL" -> "OpenAI API Base URL"

- Similar to (e)
- Note: This isn't directly authored in this PR (it's an upstream lfx component update), but the effect for any existing WatsonX users of OpenRAG is that their configured field names would mismatch. 
- Recommend checking if FlowsService needs to handle this migration.

mpawlow (changes requested):
@ricofurtado 

## Code Review 1

- See PR comments: (a) to (f)
- Note: No functional review was performed as part of this code review

flows/ingestion_flow.json
ricofurtado: Good catch, fixed.

ricofurtado:
Resolved in other PR and merged due to too many conflicts in this branch.

PR #916, opened 2026-02-06
Docs: Troubleshooting for GPU mode with local Ollama and Fedora 43

aimurphy:
From a Slack thread

### DIFF
--- a/docs/docs/support/troubleshoot.mdx
+++ b/docs/docs/support/troubleshoot.mdx
@@ -358,7 +358,7 @@ The following steps explain how to troubleshoot these issues with Fedora 43, Pod
    ```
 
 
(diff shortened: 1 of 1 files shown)


docs/docs/support/troubleshoot.mdx
```diff
@@ -268,4 +268,117 @@ For more information, see [Model availability and performance issues](#model-ava
...
+
+   ```bash
+   /usr/bin/nvidia-container-runtime-hook prestart
+   ```
+
+
+10. Test the GPU with Podman, and make sure `nvidia-smi` prints your GPU information instead of an error:
+
+   ```bash
+   podman run --rm \
```
mendonk, line 364: You might want to include that this `--rm` will remove the temporary container when it's done displaying the information.

PR #924, opened 2026-02-09
fix: support file upload for filenames with unicode characters

### LINKED ISSUE #769: OpenRAG Fails to ingest files with character '\u2013
Ingestion failed: 'ascii' codec can't encode character '\u2013' in position 14: ordinal not in range(128)

Backend logs show:
UnicodeEncodeError: 'ascii' codec can't encode character '\u2013' in position 14: ordinal not in range(128)

Cause: The filenames contain an en-dash character (–, Unicode U+2013) which cannot be encoded in ASCII. The error occurs when the system tries to set HTTP headers with these filenames.

Location: The error happens in /app/.venv/lib/python3.13/site-packages/httpx/_models.py when trying to encode header values.

Result: These files fail to ingest and the tasks complete with errors.

Is there a way we can handle special characters such that the end user doesn't need to rename files that seem selectable?

philnash:
Fixes #769 

> [!Note]
> This sends the filename as a percent-encoded string to the Langflow API. Langflow does not currently decode those strings, so the filename will travel through Langflow as `super%E2%80%93interesting%E2%80%93document.pdf` instead of `super–interesting–document.pdf` (those are en-dashes) and ultimately be stored with the percent-encoded filename. This shows back up in OpenRAG as ugly looking strings. 
>
> We could decode all the strings in OpenRAG everywhere we display them, but it shouldn't be OpenRAG's job to do that. I opened [this PR on Langflow](https://github.com/langflow-ai/langflow/pull/11658) to decode the headers before they are set in the global variables. This will lead to the filename being properly decoded and stored as its original string, so that when it is read by OpenRAG it looks normal again.

### DIFF
--- a/src/services/langflow_file_service.py
+++ b/src/services/langflow_file_service.py
@@ -1,4 +1,5 @@
 from typing import Any, Dict, List, Optional
+from urllib.parse import quote
 
 from config.settings import LANGFLOW_INGEST_FLOW_ID, clients
 from utils.logging_config import get_logger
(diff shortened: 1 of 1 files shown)


philnash:
Closing this as Langflow has implemented globals in the body of the API request in https://github.com/langflow-ai/langflow/pull/13351

PR #950, opened 2026-02-11
fix: Getting OS jwt from env

ajithreji:
This pull request updates the `create_jwt_token` method in `src/session_manager.py` to allow using a JWT token from an environment variable if it is set. This provides flexibility for deployments where a pre-generated token is preferred over dynamically generating one.

Authentication and configuration improvements:

* Modified the `create_jwt_token` method to first check for a JWT token in the `OPENSEARCH_JWT_TOKEN` environment variable, and only generate a new token if the environment variable is not set.


src/services/search_service.py
```diff
@@ -331,7 +331,7 @@ async def embed_with_model(model_name):
             "aggs": {
                 "data_sources": {"terms": {"field": "filename", "size": 20}},
                 "document_types": {"terms": {"field": "mimetype", "size": 10}},
-                "owners": {"terms": {"field": "owner_name.keyword", "size": 10}},
```
edwinjosechittilappilly, line 335: May I know why this change?

ajithreji: It was to get it aligned with the index we created.
Now that the index is made more accurate with the field names, this can be changed.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "openrag"
-version = "0.2.0"
+version = "0.2.3.0.2"
```
edwinjosechittilappilly: lets not update the version since it will trigger a release, lets keep it same as main

PR #967, opened 2026-02-16
fix(tui): do not create a directory on keydown for OpenRAG documents path

### LINKED ISSUE #892: [Bug]: Directory selector creates intermediate directories while typing custom path
### OpenRAG Version

0.2.3.0.2

### Deployment Method

uvx (uvx openrag)

### Operating System

Windows 11 WSL Ubuntu 22.04

### Python Version

3.13

### Affected Area

Onboarding (setup wizard, initial configuration)

### Bug Description

When selecting a custom directory to save documents, the application creates intermediate directories for each character typed in the path input field. This results in multiple unwanted directories being created in the filesystem as the user types the complete path.

### Steps to Reproduce

1. Open the document save settings or directory selection dialog
2. Choose the option to specify a custom directory
3. Start typing a custom path (e.g., `/mnt/e/folder`)
4. Observe the filesystem while typing each character

### Expected Behavior

The application should only create the directory once the user has:
- Finished typing the complete path, AND
- Confirmed/submitted the path (by pressing Enter, clicking Save/OK, or similar action)

The directory creation should be deferred until user confirmation, not triggered on every keystroke.

### Actual Behavior

The application should only create the directory once the user has:
- Finished typing the complete path, AND
- Confirmed/submitted the path (by pressing Enter, clicking Save/OK, or similar action)

The directory creation should be deferred until user confirmation, not triggered on every keystroke.

### Relevant Logs

```shell
As the user types the path `/mnt/e/folder`, the application creates multiple directories in real-time:
- `/mnt/e/f`
- `/mnt/e/fo`
- `/mnt/e/fol`
- `/mnt/e/fold`
- `/mnt/e/folde`
- `/mnt/e/folder`

This creates unnecessary directories in the filesystem and may cause confusion or clutter in the directory structure.
```

### Screenshots

_No response_

### Additional Context

- This appears to be caused by a directory validation or auto-creation mechanism that triggers on input change events rather than on form submission
- Suggested fix: Implement debouncing or only create directories on explicit user confirmation (button click, Enter key, blur event, etc.)
- This behavior can be problematic when typing long paths or when making typos, as it pollutes the filesystem with partial/incorrect paths

### Checklist

philnash:
By default Textual runs `Input` validators on each change. The `DocumentsPathValidator` attempts to perform some actions on the directory, including creating it if it doesn't exist, to ensure that OpenRAG can write to the selected directory.

Doing this on every change means that every keystroke will attempt to find or create the directory on the filesystem, causing many new directories as reported in #892.

Running the validator on submission means that a user will only get feedback about errors in accessing the directory when they submit, but this is preferable over creating many unused directories on their filesystem.

This fixes #892 by delaying the check until the form is submitted. It is likely best to delay file system operations until the user has made a decision, rather than trying over and over as they type.


philnash:
The integration tests aren't passing, at least for some of them, because embedding fails. This PR doesn't touch embedding at all, so the tests are flaking for some reason.

Any ideas? Will this block merging?

PR #970, opened 2026-02-16
Fix dependencies (OpenRAG on OpenSearch)

TejasQ:
Please feel free to close if this is made in error, but there seems to be a misconfiguration where OpenSearch _depends on_ OpenRAG backend when in reality, the OpenRAG backend depends on OpenSearch?

I consistently receive this error and cannot use knowledge at all. Could this be why?

Screenshot 2026-02-16 at 13 15 23

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,8 +5,6 @@ services:
     #context: .
     #dockerfile: Dockerfile
     container_name: os
-    depends_on:
-      - openrag-backend
     environment:
       - discovery.type=single-node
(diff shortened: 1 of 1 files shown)


mpawlow (commented):
> there seems to be a misconfiguration where OpenSearch depends on OpenRAG backend when in reality, the OpenRAG backend depends on OpenSearch?

- Based on offline discussions, it appears to be both (a bi-directional relationship)
- Will defer to @phact here
- For full details, see: https://github.com/langflow-ai/openrag/issues/982#issuecomment-3937026866

Issue #973, opened 2026-02-17
[Feature]: Colorize the Backend application logging events


mpawlow:
### Feature Area

Backend/API (Python/Starlette)

### Problem Description

The severity of the logging events are difficult glean / parse / read at a glance

### Proposed Solution

Colorize the Backend application logging events

### Use Case

As a user, I want to read the logs

### Alternatives Considered

N/A

### Priority

Nice to have

### Contribution

- [x] I would be willing to help implement this feature.
- [x] I can help test this feature once implemented.

### Checklist

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before.

Issue #845, opened 2026-01-23
[Bug]: Data ingestion fails due to OpenSearch "documents" index not found exception


mpawlow:
### OpenRAG Version

0.2.3.0.1

### Deployment Method

Local development (make dev)

### Operating System

Ubuntu 24.04.3 LTS

### Python Version

3.13.11

### Affected Area

Infrastructure (Docker, OpenSearch, Langflow)

### Bug Description

- OpenSearch `documents` index wasn't created during onboarding (silent failure or OpenSearch not ready)
- e.g.
```
OpenSearch error: `NotFoundError(404, 'index_not_found_exception', 'no such index [documents]', documents, index_or_alias)`. Error appears in backend logs at `src/services/search_service.py:429`. 
```
- Unknown why the index wasn't created
- Suspect it was related to the **Langflow** containerized service not being fully initialized and ready during the initial OpenRAG setup process

### Steps to Reproduce

- I don't know the exact set of steps to reproduce.
- Note: Others have encountered similar issues.

1. Follow development environment setup instructions from https://github.com/langflow-ai/openrag/blob/main/CONTRIBUTING.md#local-development-workflow

### Expected Behavior

- OpenSearch document index should be automatically created during initial OpenRAG initial setup

### Actual Behavior

- OpenSearch document index not found

### Work-around

- I needed to manually create the OpenSearch document index

```bash
source /home/mpawlow/git/openrag/.env && curl -sk -u "admin:$OPENSEARCH_PASSWORD" -X PUT https://localhost:9200/documents -H "Content-Type: application/json" -d '{
  "settings": {"index": {"knn": true, "knn.algo_param.ef_search": 512}, "number_of_shards": 1, "number_of_replicas": 0},
  "mappings": {"properties": {
    "chunk_embedding_text_embedding_3_small": {"type": "knn_vector", "dimension": 1536, "method": {"name": "disk_ann", "space_type": "l2", "engine": "jvector", "parameters": {"ef_construction": 512, "m": 16}}},
    "embedding_model": {"type": "keyword"}, "embedding_dimensions": {"type": "integer"},
    "document_id": {"type": "keyword"}, "filename": {"type": "keyword"}, "mimetype": {"type": "keyword"},
    "page": {"type": "integer"}, "text": {"type": "text"}, "source_url": {"type": "keyword"},
    "connector_type": {"type": "keyword"}, "owner": {"type": "keyword"},
[...]

PR #974, opened 2026-02-17
fix: Data ingestion fails due to OpenSearch "documents" index not found exception

mpawlow:
### Issues

- #845
- #973

### Summary

Fix data ingestion failures caused by a race condition where the backend attempted to interact with Langflow and OpenSearch before those services were ready, and improve logging and vector index configuration.

### Bug Fix: Langflow and OpenSearch Readiness

- Add new wait_for_langflow() utility with exponential backoff and jitter to poll Langflow's /health endpoint before proceeding with API key generation and client initialization (src/utils/langflow_utils.py)
- Add LangflowNotReadyError exception raised when Langflow fails to become ready within the retry limit
- Reorder AppClients.initialize() to create the langflow_http_client before calling wait_for_langflow() and get_langflow_api_key(), ensuring the HTTP client exists before health checks begin (src/config/settings.py)
- Move Langflow API key generation (get_langflow_api_key()) to after the Langflow readiness check instead of the very start of initialization
- Guard the onboarding endpoint with wait_for_langflow(), returning HTTP 503 if Langflow is unavailable (src/api/settings.py)
- Extract init_index_when_ready() to explicitly sequence wait_for_opensearch() before init_index(), and call it from the onboarding endpoint (src/main.py)
- Remove the redundant wait_for_opensearch() call inside init_index() so it can be reused without double-waiting
- Add post-onboarding index initialization in startup_tasks() to recreate the OpenSearch index when the volume has been deleted but onboarding was already completed

### Vector Index Configuration

- Increase ef_construction from 100 to 512 across all index creation paths for improved search recall (src/config/settings.py, src/utils/embedding_fields.py, src/utils/embeddings.py, flows/components/opensearch_multimodal.py, scripts/migrate_embedding_model_field.py)

### Logging Improvements

- Add colored log output with ANSI codes for log level and timestamp, respecting NO_COLOR env var and non-TTY environments (src/utils/logging_config.py)
- Document LOG_LEVEL and NO_COLOR environment variables in .env.example

### Code Quality

- Fix trailing whitespace throughout src/api/settings.py (whitespace-only cleanup)

### Testing

- Add comprehensive unit tests for wait_for_langflow() covering success, transient failures, retry exhaustion, backoff bounds, and edge cases (tests/unit/test_langflow_utils.py)
- Add tests/unit/conftest.py with a no-op onboard_system fixture override so unit tests don't require running infrastructure

### Screenshots

2026-02-13 20_40_19-mpawlow@OpenRAG1_ ~_git_openrag


.env.example
```diff
@@ -123,6 +123,15 @@ NEXT_ALLOWED_DEV_ORIGINS=
 # LANGFLOW_WORKERS=2    # Langflow workers (default: 1)
 # DOCLING_WORKERS=2     # Docling workers (default: 1)
 
+# OPTIONAL: Set the logging level for the Backend service
+# - Supported values: DEBUG, INFO, WARNING, ERROR, CRITICAL
+# - Default: INFO
+# LOG_LEVEL=DEBUG
```
edwinjosechittilappilly, line 129: should we add this to docker compose and helm?

mpawlow: Yes, good catch

- Added `LOG_LEVEL` to the relevant Helm / Docker configuration files
- Also, added a whole bunch of missing environment variables that were present in `.env.example` but not in the relevant Helm / Docker configuration files
- For full details, see https://github.com/langflow-ai/openrag/pull/974/changes/b877c384cce3789550c0b1bab1db6a35b2a5e603

PR #986, opened 2026-02-18
feat: Helm chart changes to import Flow and document samples files from git

rodageve:
**Description:**
- Import large Flow and Sample document files into the langflow and backend pods respectively by downloading the files from git in an initContainer
- Remove importing these files as a separate step using ConfigMaps
- Updates to the Ingestion Flow from cloud testing

Alternatives for Cloud deployment:
1. Consider importing the files from object storage such as S3
2. An operator can deploy the respective files from S3 as required into the backend and langflow pdos

### DIFF
--- a/kubernetes/helm/openrag/values.yaml
+++ b/kubernetes/helm/openrag/values.yaml
@@ -84,7 +84,7 @@ langflow:
       repo: openrag
       branch: main
       # Optional: specific git commit hash (overrides branch if set)
-      hash: f5635c9906df581384c62b570afb55ace65ad1eb
+      #hash: ""
       path: flows
       files:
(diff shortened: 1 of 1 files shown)


flows/ingestion_flow.json
```diff
@@ -4127,7 +4127,7 @@
                 "title_case": false,
                 "track_in_telemetry": false,
                 "type": "str",
-                "value": "jwt"
```
edwinjosechittilappilly, line 4130: @rodageve We may not need the ingestion flow .json since in its using the opensearchg pass instead of JWT.

rodageve: The ingestion flow has been removed from the PR.

Issue #991, opened 2026-02-19
[Feature]: Add Docker support and SSE transport for MCP


scrobot:
### Feature Area

Infrastructure (Docker, OpenSearch, Langflow)

### Problem Description

The OpenRAG MCP server currently only supports stdio transport, limiting it to being spawned as a local subprocess by an MCP client. There is no way to deploy the MCP server as a standalone networked service or include it in the standard OpenRAG Docker stack. This prevents use cases like remote MCP client connections, multi-host deployments, and containerized production environments.

### Proposed Solution

1. **SSE transport** — Add a Server-Sent Events transport mode to server.py using Starlette + Uvicorn, exposing /sse and /messages/ HTTP endpoints. Transport is selected via the MCP_TRANSPORT environment variable (stdio by default, sse for Docker/network).

2. **Dockerfile** — Add `Dockerfile.mcp` (Python 3.13-slim, uv package manager) to build a lightweight container image for the MCP server.

3. **Docker Compose integration** — Add an openrag-mcp service to `docker-compose.yml` that connects to the frontend container internally and exposes a configurable port (default 8080).

4. **Configuration** — Extend `.env.example` with new variables:
    - MCP_PORT — exposed port (default 8080)
    - OPENRAG_API_KEY — API key for MCP-to-OpenRAG authentication
    - OPENRAG_MCP_TIMEOUT — request timeout in seconds (default 60)
    - OPENRAG_MCP_MAX_RETRIES — max retry attempts (default 3)

### Use Case

- **IDE integrations** — Editors like VS Code or Cursor connect to OpenRAG tools via a remote MCP endpoint instead of spawning a local process.
- **Containerized deployments** — Run the full OpenRAG stack (including MCP) with a single docker compose up.
- **Multi-host setups** — MCP server runs on a shared server while multiple developers connect from their machines.

### Priority

Would improve my workflow

### Additional Context

- The SSE transport follows the [MCP SSE specification](https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse) and uses the official mcp Python SDK's SseServerTransport.
- The stdio transport remains the default and is unaffected by this change.
- The Dockerfile uses a two-stage copy (dependencies first, source second) to optimize layer caching during rebuilds.

### Contribution

- [x] I would be willing to help implement this feature.
- [x] I can help test this feature once implemented.

### Checklist

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before.

PR #992, opened 2026-02-19
feat: Add MCP server Docker support with SSE transport

scrobot:
# Summary

Adds Docker containerization and SSE transport support for the OpenRAG MCP server, enabling networked and containerized deployments alongside the existing stdio transport.

# Changes
- sdks/mcp/src/openrag_mcp/server.py — Add run_sse_server() using Starlette + Uvicorn with /sse and /messages/ endpoints; refactor main() to select transport via MCP_TRANSPORT env var
- Dockerfile.mcp — New Dockerfile (Python 3.13-slim, uv) with optimized layer caching for the MCP server image
- docker-compose.yml — Add openrag-mcp service with internal networking to openrag-frontend and configurable port mapping
- .env.example — Add MCP_PORT, OPENRAG_API_KEY, OPENRAG_MCP_TIMEOUT, OPENRAG_MCP_MAX_RETRIES configuration variables

# Notes
- Default transport remains stdio — no breaking changes for existing local usage
- SSE transport follows the [MCP SSE specification](https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse) using the official mcp SDK's SseServerTransport

Details proposed here: https://github.com/langflow-ai/openrag/issues/991


philnash:
SSE is a deprecated transport for MCP, it has been superseded by [Streamable HTTP](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http). SSE should only be included for backwards compatibility with old clients.

Is there a reason that this remote server capability is being added as a new process and new container, when it could be endpoints on the openrag-backend starlette server?

philnash:
> > Is there a reason that this remote server capability is being added as a new process and new container, when it could be endpoints on the openrag-backend starlette server?
> 
> This is intentional and driven by the architecture I'm targeting. The idea is that OpenRAG serves as the central knowledge core for the platform, while MCP servers should be independently deployable units. A few reasons for the separation:
> 
>     1. **Flexible deployment topology** — MCP servers can be deployed in different environments and isolation levels depending on the use case (e.g., production cluster, staging, internal tooling).

I don't see how that's different from the OpenRAG backend itself.

>     2. **Local development experience** — Any developer on the team can spin up their own local MCP server instance connected to OpenRAG for personal agents and experimentation, without running the full backend.

If OpenRAG hosted the MCP server too, then you wouldn't need to spin up anything, which would be even easier.

>     3. **Independent scaling & lifecycle** — MCP server traffic patterns are different from the main backend. Decoupling them allows independent scaling, updates, and failure isolation.

But all actions taken through the MCP server hit the OpenRAG backend via the API. By adding the MCP server as an extra service, that just results in one more service that needs scaling, but doesn't stop you having to scale the backend too.

>     4. **Composability** — Treating the MCP server as a standalone, containerized service makes it easy to plug into various agent setups and orchestration tools without coupling to the OpenRAG backend's deployment.

I don't see how this helps at all. The MCP client doesn't care whether the server is containerized and separate or part of the same backend. 
 
>     5. **Data access control (upcoming)** — I'm currently working on an issue for granular access levels within the RAG knowledge base. The idea is that different memory segments will have different visibility scopes — some data should be fully private, some shared within a team, some broadly accessible, etc. Having MCP servers as separate instances makes this much more natural: each server instance can be configured with its own access scope and permissions, rather than trying to manage all of that within a single monolithic backend.

I would have thought that access control would be controlled at a user/API key layer. 
 
> Essentially, OpenRAG is the knowledge layer, and MCP servers are lightweight, portable adapters that can live wherever agents need them.

As far as I can tell, OpenRAG is already serving itself over HTTP as a web application and API, and I see no difference in providing MCP from that too. A separate MCP server only makes starting it up and using it harder than being able to connect to the backend service you are running already.

PR #1001, opened 2026-02-19
feat: add playwright testing to frontend for end to end tests

lucaseduoli:
This pull request introduces end-to-end (E2E) testing infrastructure for the frontend using Playwright, including environment setup, configuration, and a sample test. It adds Playwright dependencies, a test environment configuration, scripts for automated infrastructure setup, and updates to the Makefile for safer factory resets. The changes are grouped below by theme.

**E2E Testing Infrastructure**

* Added Playwright as a dev dependency (`@playwright/test`, `playwright`, `playwright-core`) and configured Playwright for E2E tests in `frontend/playwright.config.ts`, including web server startup and environment variable management. [[1]](diffhunk://#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R64) [[2]](diffhunk://#diff-4a2d9aa3e849b134993936ca81b83fb139edd2b0218077ab0f403b8c4803c62aR57) [[3]](diffhunk://#diff-4a2d9aa3e849b134993936ca81b83fb139edd2b0218077ab0f403b8c4803c62aR1297-R1313) [[4]](diffhunk://#diff-4a2d9aa3e849b134993936ca81b83fb139edd2b0218077ab0f403b8c4803c62aR5464-R5510) [[5]](diffhunk://#diff-1ed7245e0c7c634928b52839600c3d493c9d29150d369e9ab20979e61e7f8f97R1-R62)
* Created a sample Playwright test (`frontend/tests/core/onboarding.spec.ts`) to verify onboarding content is visible.
* Added a global setup script (`frontend/tests/globalSetup.ts`) and shell scripts (`scripts/setup-e2e.sh`, `scripts/start-backend-e2e.sh`) to automate environment preparation and backend startup for tests. [[1]](diffhunk://#diff-6437565f991716cbd74c25f032a01a7119ef1312af846cb1d717a218c9f7d9b3R1-R16) [[2]](diffhunk://#diff-fbe4608d19692209377719610da990a0c5bfea326b75144a744df7c4a54f76beR1-R47) [[3]](diffhunk://#diff-aaf067c55866f418a97a199360b41083c84c1f73b822cfa14687c5de9822d9c1R1-R13)

**Environment and Configuration**

* Introduced `.env.test` for test-specific environment variables, and updated `.gitignore` to allow committing this file and ignore Playwright reports. [[1]](diffhunk://#diff-559409600c013973a4129ab3ba0bc5585037cc25c7295ed291a3ee935ba6ceccR1-R37) [[2]](diffhunk://#diff-8e548ede50532796f028e75d8f3384d5bdb57b9fbe2bda95521774d0c46d0f97R35-R44)
* Increased webpack chunk load timeout in `frontend/next.config.ts` to prevent timeouts during development with large bundles.

**Makefile and Miscellaneous**

* Improved `factory-reset` in `Makefile` to allow forced resets without user confirmation and removed redundant deletion of the `opensearch-data` directory. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R460-R467) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L474)
* Added a test id to the onboarding content div for easier test targeting (`frontend/app/onboarding/_components/onboarding-content.tsx`).

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -73,17 +73,31 @@ jobs:
         uses: actions/cache@v5
         with:
           path: ${{ runner.temp }}/docker-cache
-          key: docker-images-${{ runner.os }}-${{ hashFiles('docker-compose.yml') }}
+          key: docker-images-v1-${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
           restore-keys: |
-            docker-images-${{ runner.os }}-
+            docker-images-v1-${{ runner.os }}-
 
-      - name: Load Cached Docker Images
+      - name: Validate Docker Cache
+        id: validate-docker-cache
         if: steps.cache-docker.outputs.cache-hit == 'true'
         run: |
-          if [ -f ${{ runner.temp }}/docker-cache/images.tar ]; then
-            echo "Loading cached Docker images..."
-            docker load -i ${{ runner.temp }}/docker-cache/images.tar
+          TAR="${{ runner.temp }}/docker-cache/images.tar"
+          # Require at least 100 MB to consider the cache valid
+          MIN_SIZE=$((100 * 1024 * 1024))
+          if [ -f "$TAR" ] && [ "$(stat -c%s "$TAR")" -ge "$MIN_SIZE" ]; then
+            echo "Docker cache is valid ($(du -sh "$TAR" | cut -f1))"
+            echo "valid=true" >> $GITHUB_OUTPUT
+          else
+            echo "Docker cache is invalid or too small, deleting..."
+            rm -f "$TAR"
+            echo "valid=false" >> $GITHUB_OUTPUT
           fi
+
+      - name: Load Cached Docker Images
+        if: steps.cache-docker.outputs.cache-hit == 'true' && steps.validate-docker-cache.outputs.valid == 'true'
+        run: |
+          echo "Loading cached Docker images..."
+          docker load -i ${{ runner.temp }}/docker-cache/images.tar
 
       # Cache Playwright browsers using a composite key
       - name: Cache Playwright Browsers
@@ -107,9 +121,10 @@ jobs:
           ./scripts/setup-e2e.sh
 
       - name: Save Docker Images for Caching
-        if: steps.cache-docker.outputs.cache-hit != 'true'
+        if: steps.cache-docker.outputs.cache-hit != 'true' || steps.validate-docker-cache.outputs.valid == 'false'
         run: |
           echo "Saving Docker images to cache..."
+          rm -rf ${{ runner.temp }}/docker-cache
           mkdir -p ${{ runner.temp }}/docker-cache
           docker save -o ${{ runner.temp }}/docker-cache/images.tar \
             langflowai/openrag-opensearch:latest \


frontend/.gitignore
```diff
@@ -32,10 +32,14 @@ yarn-error.log*
 
 # env files (can opt-in for committing if needed)
 .env*
+!.env.test
```
mpawlow: ### (a) [Minor] Recommend following the root folder `.env.example` pattern

- i.e. Should rename to `env.test.example`
- i.e. Should keep `.env*` as the `.gitignore` pattern and not commit `.env.test`
- Also, recommend updating `setup-e2e.sh` to copy the example if `.env.test` doesn't exist.

frontend/.env.test
```diff
@@ -0,0 +1,37 @@
+# Ingestion Configuration
+DISABLE_INGEST_WITH_LANGFLOW=false
+INGEST_SAMPLE_DATA=true
+
+# Auth
+LANGFLOW_SECRET_KEY=supersecretkeyfortesting
```
mpawlow: ### (b) [Normal] Recommend leaving the LANGFLOW_SECRET_KEY value empty

- Many security scanners will falsely flag this as a security vulnerability or a leaked credential even though the value is obviously false

frontend/.env.test
```diff
@@ -0,0 +1,37 @@
...
+OPENSEARCH_PASSWORD=StrongTestPass1!
```
mpawlow: ### (c) [Normal] Recommend leaving the LANGFLOW_SECRET_KEY and OPENSEARCH_PASSWORD values empty

- Same as (b)
- Many security scanners will falsely flag this as a security vulnerability or a leaked credential even though the value is obviously false

frontend/.env.test.example
```diff
@@ -0,0 +1,37 @@
...
+
+# Langflow
+LANGFLOW_AUTO_LOGIN=True
+LANGFLOW_NO_LOGIN=True
+LANGFLOW_DEACTIVATE_TRACING=true
+LANGFLOW_LOG_LEVEL=ERROR
+DO_NOT_TRACK=true
+LANGFLOW_ENABLE_SUPERUSER_CLI=false
+LANGFLOW_NEW_USER_IS_ACTIVE=false
+
```
mpawlow, line 32: ### (d) [Question] Are any of these missing env var values required?

```
LANGFLOW_SUPERUSER / LANGFLOW_SUPERUSER_PASSWORD
LANGFLOW_CHAT_FLOW_ID / LANGFLOW_INGEST_FLOW_ID / NUDGES_FLOW_ID
```

frontend/tests/globalSetup.ts
```diff
@@ -0,0 +1,16 @@
+import { execSync } from "child_process";
```
mpawlow: ### (e) [Normal] Appears to be dead code?

- Should be removed
- Commented out in `frontend/playwright.config.ts`

frontend/tests/globalSetup.ts
```diff
@@ -0,0 +1,16 @@
...
+import path from "path";
+
+async function globalSetup() {
+  console.log("Global Setup: Running factory reset and infrastructure startup...");
+  try {
+    const scriptPath = path.resolve(__dirname, "../../scripts/setup-e2e.sh");
+    execSync(scriptPath, { stdio: "inherit" });
```
mpawlow: ### (f) [Major] Security: execSync shell injection risk

- CodeQL flagged this. 
- While scriptPath is resolved from __dirname and isn't user-controlled, using execSync with a string argument runs through /bin/sh, which interprets metacharacters. 
- The safer approach per CodeQL's suggestion:
```typescript
import { execFileSync } from "child_process";
execFileSync("bash", [scriptPath], { stdio: "inherit" });
```
- If this file is indeed dead code, you can safely just delete this file to resolve

scripts/setup-e2e.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Clean up using make
+echo "Cleaning up..."
+make factory-reset FORCE=true ENV_FILE=$E2E_ENV
+
+# Start infrastructure using make (this will use the new .env)
+echo "Starting infrastructure..."
+make dev-local-cpu ENV_FILE=$E2E_ENV
+
+echo "Waiting for OpenSearch..."
```
mpawlow, line 56: ### (g) [Normal] Infinite wait loops with no timeout 

- Loop will spin indefinitely if the services fail to start. 
- In CI this will hang the job until the runner times out (potentially wasting 30–60 min of CI time). 
- Recommend: Adding a counter or use a timeout utility:
- e.g.

```shell
RETRIES=60  # 5 min max
COUNT=0
until curl -s -k https://localhost:9200 >/dev/null || [ $COUNT -ge $RETRIES ]; do
    sleep 5
    COUNT=$((COUNT + 1))
    echo "Waiting for OpenSearch... ($COUNT/$RETRIES)"
done
[ $COUNT -ge $RETRIES ] && echo "OpenSearch failed to start" && exit 1
```

scripts/setup-e2e.sh
```diff
@@ -0,0 +1,40 @@
...
+until curl -s -k https://localhost:9200 >/dev/null; do
+    sleep 5
+    echo "Waiting for OpenSearch..."
+done
+
+echo "Waiting for Langflow..."
```
mpawlow, line 69: ### (h) [Normal] Infinite wait loops with no timeout 

- Same as (g)
- Loop will spin indefinitely if the services fail to start. 
- In CI this will hang the job until the runner times out (potentially wasting 30–60 min of CI time). 
- Recommend: Adding a counter or use a timeout utility

frontend/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+  projects: [
+    {
+      name: "chromium",
+      use: { ...devices["Desktop Chrome"] },
+    },
+  ],
+
+  webServer: [
+    {
+      command: "../scripts/start-backend-e2e.sh",
```
mpawlow: ### (i) [Normal] Full factory reset + infrastructure startup should be separated from Playwright invocation

- Question: Is a full factory reset necessary for CI?
- In CI (process.env.CI=true), Playwright will always run `start-backend-e2e.sh`, which runs `setup-e2e.sh`, which runs factory-reset. 
- This is very aggressive for a webServer command
- Playwright's webServer is meant to start a server rather than orchestrate full environment teardown and
rebuild. 
- e.g. If the factory reset or infra startup fails mid-way, the error reporting through Playwright's webServer abstraction will be opaque.
- Recommend: Putting the factory-reset + infrastructure startup in a dedicated CI step (separate make target or GitHub Actions step), and have `start-backend-e2e.sh` only start the backend assuming infrastructure is already running.

frontend/next.config.ts
```diff
@@ -32,6 +32,15 @@ const nextConfig: NextConfig = {
   },
   // Allow cross-origin requests in development
   allowedDevOrigins: getAllowedDevOrigins(),
+  // Increase chunk load timeout for large bundles in dev mode
+  // The layout.js chunk (~22MB in dev) can timeout on first compile
+  webpack: (config) => {
+    config.output = {
+      ...config.output,
+      chunkLoadTimeout: 300000, // 5 minutes (default is 120s)
```
mpawlow: ### (j) [Minor] `chunkLoadTimeout` not scoped to dev mode

- Webpack config receives dev and isServer parameters that are not used:

```typescript
webpack: (config, { dev, isServer }) => {
    if (dev && !isServer) {
        config.output = { ...config.output, chunkLoadTimeout: 300000 };
    }
    return config;
},
```

- In Production, a 5-minute chunk load timeout would mask real loading failures and degrade the user-facing performance. 
- Also, 22MB layout.js in dev should be investigated in the future  🙂  
- Note: This is a Minor severity comment, please feel free to ignore or optionally implement

frontend/playwright.config.ts
```diff
@@ -0,0 +1,62 @@
...
+      url: "http://127.0.0.1:8000/health", 
+      reuseExistingServer: !process.env.CI,
+      stdout: "pipe",
+      stderr: "pipe",
+      timeout: 300 * 1000,
+      env: {
+        // Pass env vars from .env.test
+        ...process.env,
```
mpawlow: ### (k) [Minor] Security: Spreading process.env into the webServer env is unnecessarily broad

- Spreading all environment variables passes secrets from the developer's shell into the subprocess.
- Recommend: Only the variables actually needed should be passed explicitly, which also makes the configuration self-documenting.
- Note: This is a Minor severity comment, please feel free to ignore or optionally implement

mpawlow (changes requested):
@lucaseduoli 

## Code Review 1

- See PR comments: (a) to (k)

frontend/.gitignore
lucaseduoli: but the .env.test is supposed to always be the same, so that we can have a consistent environment for tests. What should we do on that case? Assuming it is not an "example", but it is the correct value

frontend/.env.test.example
lucaseduoli: Not required, they are assigned automatically if not filled in

frontend/.gitignore
mpawlow: > but the .env.test is supposed to always be the same, so that we can have a consistent environment for tests
- Understood
- Definitely makes sense for the CI env
- I was thinking more about the ability to configure and run the automated tests in the local development environment
- Changing to **Minor** severity
- Please feel free to ignore this recommendation or optionally implement

mpawlow: 🚀

frontend/.env.test
mpawlow: 🚀

frontend/.env.test
mpawlow: 🚀

frontend/tests/globalSetup.ts
mpawlow: 🚀

frontend/tests/globalSetup.ts
mpawlow: 🚀

scripts/setup-e2e.sh
mpawlow: 🚀

scripts/setup-e2e.sh
mpawlow: 🚀

frontend/playwright.config.ts
mpawlow: 🚀

frontend/next.config.ts
mpawlow: 🚀

frontend/playwright.config.ts
mpawlow: 🚀

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Install backend dependencies
+        run: uv sync
+
+      - name: Install frontend dependencies
+        working-directory: frontend
+        run: npm ci
+
+      - name: Install Playwright browsers
+        working-directory: frontend
+        run: npx playwright install chromium
```
mpawlow: ### (2a) [Question] Sanity Confirmation: Was the change to remove `--with-deps` done on purpose?

- Not sure if this will cause a regression
```
  # Before
  run: npx playwright install --with-deps chromium
  # After
  run: npx playwright install chromium
```
- Note: --with-deps instructs Playwright to also install the OS-level system libraries that Chromium needs (libglib2.0, libnss3, libatk, etc.). Without it, only the Playwright-managed browser binary is downloaded
   - If the runner doesn't have those system packages pre-installed, Chromium will fail to launch

scripts/setup-e2e.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# Go to project root
+cd "$(dirname "$0")/.."
+
+# Environment file for E2E tests
+E2E_ENV="frontend/.env.test"
+E2E_ENV_EXAMPLE="frontend/.env.test.example"
+
+# Create .env.test from the example template (never modify the tracked example file)
+cp "$E2E_ENV_EXAMPLE" "$E2E_ENV"
```
mpawlow: ### (2b) [Minor] `.env.test` unconditionally overwriten on every run

- This runs first, before checking any existing content. If a local developer has configured .env.test with their own API keys, this silently destroys it every time they run setup-e2e.sh. 
- Recommendation
```shell
  if [ ! -f "$E2E_ENV" ]; then
      cp "$E2E_ENV_EXAMPLE" "$E2E_ENV"
  fi
```
- Issue not applicable in the CI system

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,103 @@
...
+      - '.github/workflows/test-e2e.yml'
+  workflow_dispatch:
+
+jobs:
+  e2e:
+    runs-on: [self-hosted, linux, ARM64, langflow-ai-arm64-40gb]
+    env:
+      LANGFLOW_AUTO_LOGIN: "True"
+      LANGFLOW_NEW_USER_IS_ACTIVE: "True"
+      LANGFLOW_ENABLE_SUPERUSER_CLI: "True"
```
mpawlow, line 31: ### (2c) [Minor] Should this value be "False"? In frontend/.env.test.example, the value for `LANGFLOW_ENABLE_SUPERUSER_CLI` is `false`

- Consistency Problem: The Makefile's ENV_FILE mechanism exports the file-based values into the shell, and Docker Compose inherits them
   - The false value from the file will likely take effect for the Langflow container, while the job-level env may apply to other steps

.github/workflows/test-e2e.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Upload test results
+        uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() }}
+        with:
+          name: playwright-test-results
+          path: frontend/test-results/
+          retention-days: 7
+
+      - name: Teardown infrastructure
```
mpawlow, line 174: ### (2d) [Minor] CI teardown depends on .env.test existing

- If setup-e2e.sh fails before creating .env.test (e.g., it crashes before the cp line completes), the teardown step silently fails (the || true swallows the error). 
- Subsequent runs would then have stale containers from the previous run. 
- Recommendation: A safer fallback:
```yaml
  run: |
    ENV_FILE="frontend/.env.test"
    if [ ! -f "$ENV_FILE" ]; then ENV_FILE="frontend/.env.test.example"; fi
    docker compose --env-file "$ENV_FILE" down -v --remove-orphans || true
```

mpawlow (approved):
@lucaseduoli 

## Code Review 2

- ✅  Approved / LGTM
- 1 Question + 3 Minor severity issues
   - See: PR comments (2a) to (2d)
- Also, Minor severity
   - Not sure if were going to add the short `README.md` to the `tests/` directory in this PR that outlining what is needed to run the front-end tests (e.g. running infrastructure, auto-generated `.env.test` file, etc.)

.github/workflows/test-e2e.yml
lucaseduoli: It was on purpose, since it was failing to install with it. It had problem installing it without root

scripts/setup-e2e.sh
lucaseduoli: Sounds good! do you think we can make it so that it uses .env if configured, and if not, uses .env.test?  or should we keep everything on .env.test

.github/workflows/test-e2e.yml
lucaseduoli: Good call

scripts/setup-e2e.sh
mpawlow: > do you think we can make it so that it uses .env if configured

- That's a really good question
- Ideally, the tests would load / use `.env.test` and then load / use `.env` (if it exists) as a fallback for any values not defined in `.env.test`
   - i.e. An app configuration value hierarchy:
      - Superset( .`env.test` <- `.env` )
      - And values in `env.test` take precedence
   - This would work like `nconf` in Node.js
   - However, this suggestion may be over-engineered / over-complicated for a custom solution
      - Future: Not sure if Python has a default mechanism / library for hierarchical app configurations
- Practically, it may be best to keep `.env` and `env.test` completely separate (and have the developer manually copy and paste values between them) in order to prevent any confusion and clearly separate the values

.github/workflows/test-e2e.yml
lucaseduoli: Replicated test-integration, Will keep it this way for consistency

Issue #949, opened 2026-02-11
[Feature]:  UI Update for Sync feature


edwinjosechittilappilly:
### Feature Area

Connectors (Google Drive, OneDrive, SharePoint)

### Problem Description

UIn UX for sync and allowed users need to improve.

### Proposed Solution

NA

### Use Case

NA

### Priority

Would improve my workflow

### Contribution

- [x] I would be willing to help implement this feature.
- [x] I can help test this feature once implemented.

### Checklist

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before.

prasanthcaibmcom:
@mfortman11 Please take this one and work with @Analuisa1a on design

Analuisa1a:
@mfortman11 I had suggested this to edwin and lucas 

https://github.com/user-attachments/assets/2bdc739a-aead-40fe-945e-28d21d8ddc24

Analuisa1a:
@prasanthcaibmcom : Design signs off on the design after review 

The model provider will also need to have an update to the styling in a new issue

PR #1011, opened 2026-02-23
refactor: Update connector card styles

mfortman11:
Addresses https://github.com/langflow-ai/openrag/issues/949
figma: https://www.figma.com/design/8BmENbRoNQ0yKpOL32L7Q4/UI-sync_-949?node-id=1-8403&m=dev


lucaseduoli (changes requested):
I believe the Connect button should not have this hover state, changing the color to gray, since it's not consistent with the other buttons. Code-wise, this is perfect!

Issue #989, opened 2026-02-19
[Bug]: Anthropic validation uses deprecated model ID and agent crashes when response.output is None


scrobot:
### OpenRAG Version

0.2.4

### Deployment Method

Docker

### Operating System

macOS 26.1

### Affected Area

Chat (chat interface, conversations, AI responses)

### Bug Description

There are two related issues with Anthropic provider integration:

1. Agent crashes on empty API responses
In `src/agent.py`, response.output_text is a property that internally iterates over response.output. When response.output is None, it raises a TypeError — not an AttributeError. Since the current code uses getattr(response, "output_text", None), the TypeError is uncaught and propagates as an unhandled exception.

2. Deprecated model ID used for API key validation
Both `src/api/provider_validation.py` and `src/services/models_service.py` reference `claude-3-5-haiku-latest` for lightweight API key validation. This model alias is deprecated and may stop resolving, causing validation to fail even with a valid API key.

Additionally, the Anthropic models list is outdated and missing the current Claude 4.5/4.6 model family.

### Steps to Reproduce

1. choose Claude model
2. paste Claude API key
3. get the error

### Expected Behavior

Agent should gracefully handle None responses from the Anthropic API without crashing.
API key validation should use a current, stable model ID.
The models list should include the latest available Claude models.

### Actual Behavior

1. choose Claude model
2. paste Claude API key
3. success

### Relevant Logs

```shell

```

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

PR #1024, opened 2026-02-25
fix: anthropic validation and refactor model constants

edwinjosechittilappilly:
Add src/config/model_constants.py containing curated model lists and default model constants for OpenAI, Anthropic, and Ollama. Update ModelsService to import and use these constants (validation lists, embedding prefix, and defaults), detect embedding models by prefix, and build Anthropic language models from the /v1/models response. Switch Anthropic lightweight validation to use the /v1/models endpoint (no token consumption) in both ModelsService and provider_validation, remove the unused messages payload, and include response.text in error messages. Also apply minor formatting cleanups.

closes issue: https://github.com/langflow-ai/openrag/issues/989

### DIFF
--- a/.github/workflows/add-labels.yml
+++ b/.github/workflows/add-labels.yml
@@ -12,7 +12,7 @@ jobs:
       issues: write
     steps:
       - name: Manage LGTM Review Label
-        uses: actions/github-script@v8.0
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
           script: |
             try {
--- a/.github/workflows/auto-delete-branch.yml
+++ b/.github/workflows/auto-delete-branch.yml
@@ -12,7 +12,7 @@ jobs:
       contents: write
     steps:
       - name: Delete merged branch
-        uses: actions/github-script@v8.0
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
           script: |
             const branchName = context.payload.pull_request.head.ref;


src/config/model_constants.py
```diff
@@ -0,0 +1,39 @@
...
+    "o4-mini-high",
+]
+
+OPENAI_DEFAULT_LANGUAGE_MODEL = "gpt-4o"
+OPENAI_DEFAULT_EMBEDDING_MODEL = "text-embedding-3-small"
+OPENAI_EMBEDDING_MODEL_PREFIX = "text-embedding"
+
+ANTHROPIC_DEFAULT_LANGUAGE_MODEL = "claude-sonnet-4-5-20250929"
+
+OLLAMA_DEFAULT_LANGUAGE_MODEL_PATTERN = "gpt-oss"
```
mpawlow, line 39: ### (a) [Question] Sanity Confirmation: Was this pattern purposely made to be so specific / granular?

- Potential Problem: If no installed Ollama model has "gpt-oss" in its name, no default is ever set and the UI will show all models without a selected default.

src/config/model_constants.py
```diff
@@ -0,0 +1,39 @@
...
+    "o3-pro",
+    "o4-mini",
+    "o4-mini-high",
+]
+
+OPENAI_DEFAULT_LANGUAGE_MODEL = "gpt-4o"
+OPENAI_DEFAULT_EMBEDDING_MODEL = "text-embedding-3-small"
+OPENAI_EMBEDDING_MODEL_PREFIX = "text-embedding"
+
+ANTHROPIC_DEFAULT_LANGUAGE_MODEL = "claude-sonnet-4-5-20250929"
```
mpawlow, line 37: ### (b) [Minor] Default appears to be outdated

- The current latest Sonnet is claude-sonnet-4-6. 
- If this constant isn't in ANTHROPIC_VALIDATION_MODELS, the default will never match anything returned by the API, meaning no Anthropic model will be marked as default in the UI.

src/services/models_service.py
```diff
@@ -121,35 +97,32 @@ async def get_anthropic_models(self, api_key: str) -> Dict[str, List[Dict[str, s
...
 
-                for model_id in self.ANTHROPIC_MODELS:
-                    language_models.append(
-                        {
-                            "value": model_id,
-                            "label": model_id,
-                            "default": model_id == "claude-sonnet-4-5-20250929",
-                        }
-                    )
+                for model in models:
```
mpawlow, line 128: ### (c) [Normal] Silent empty model list

- Problem: If Anthropic returns a valid 200 response but none of the returned model IDs appear in ANTHROPIC_VALIDATION_MODELS (e.g., after a naming scheme change), language_models will be empty and the function silently returns 
```json 
{"language_models": [], "embedding_models": []}
```
   - No error is raised, no warning is logged. 
   - The UI will present an empty model list with no explanation.
- Recommendation: Add a warning log when the filtered list is empty, similar to what is done for IBM

src/config/model_constants.py
edwinjosechittilappilly: This was just a refactor of the old code.

src/config/model_constants.py
edwinjosechittilappilly: didnt update since langflow doesnt seems to be updated yet with latest model.
Hence aligned it with the langflow image.

src/services/models_service.py
edwinjosechittilappilly: Also added it for open ai too.
Good Suggestion mike. Thank you

src/config/model_constants.py
mpawlow: 🚀

src/config/model_constants.py
mpawlow: 🚀

src/services/models_service.py
```diff
@@ -97,6 +73,19 @@ async def get_openai_models(self, api_key: str) -> Dict[str, List[Dict[str, str]
                     key=lambda x: (not x.get("default", False), x["value"])
                 )
 
+                if not language_models:
+                    logger.warning(
```
mpawlow, line 77: ### (2a) [Normal] %-style format specifiers likely doesn't work with this project's structlog setup

- `Problem`: I believe a structlog processor is missing to perform %-style substitution
- The log message will render literally as "...The API returned %d models..." 
   - e.g. The %d placeholder is never filled.
   - e.g. The positional_args will not appear inline
- Recommendation: Use f-strings

src/services/models_service.py
```diff
@@ -97,6 +73,19 @@ async def get_openai_models(self, api_key: str) -> Dict[str, List[Dict[str, str]
...
+                        "The API returned %d models, none matched OPENAI_VALIDATION_MODELS. "
+                        "This may indicate a model naming scheme change.",
+                        len(models),
+                    )
+                if not embedding_models:
+                    logger.warning(
+                        "OpenAI API key is valid but no embedding models were found matching prefix '%s'.",
+                        OPENAI_EMBEDDING_MODEL_PREFIX,
+                    )
+
```
mpawlow, line 88: ### (2b) [Minor] Success log fires even when models list is empty

- Recommendation: Put the info statement under an `else` block

src/services/models_service.py
mpawlow: 🚀

Issue #1016, opened 2026-02-24
[Bug]: Deleted document stays in the list


RadekHavelka:
### OpenRAG Version

0.2.4

### Deployment Method

uvx (uvx openrag)

### Operating System

Windows 11, WSL

### Python Version

3.13

### Affected Area

Frontend (Next.js UI, components)

### Bug Description

When the ingestion of document failed, I tried to delete it. I checked the box, hit DELETE button, recieved confirmation in the bottom of the screen. But the file remains in the list. Hitting SYNC doesnt solve it. Refreshing the page does. This is very confusing.

### Steps to Reproduce

as above

### Expected Behavior

Deleted document will disappear from the list

### Actual Behavior

Deleted document is still visible

### Relevant Logs

```shell

```

### Screenshots

<img width="2499" height="825" alt="Image" src="https://github.com/user-attachments/assets/94cc0ed7-9b07-44ab-9472-6cd9951a155c" />

<img width="3096" height="1648" alt="Image" src="https://github.com/user-attachments/assets/1ea4fa97-c3dd-4939-bc6e-daf48ffa5dea" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

jlyoungblood:
@Wallgau @Analuisa1a, this seems similar to #869? It seems like the design pattern for that issue - giving the user the ability to delete the item - should cover this.

PR #1055, opened 2026-03-02
fix: delete table refresh and normalize long filename formatting.

Wallgau:
issue https://github.com/langflow-ai/openrag/issues/1016
File deleted do not stay longer in the table
and fix format for deletion confirmation

https://github.com/user-attachments/assets/5068b0c4-b9a3-46b2-a045-02325ec518bf


mfortman11 (approved):
Makes sense to me

Issue #575, opened 2025-12-02
Address unsustainable OpenRAG docs PDF build for "Ask OpenRAG about itself".


aimurphy:
Currently, the process to generate the OpenRAG docs PDF is manual and tedious:

* Even with Agent assistance, the output it unreliable and requires editing by hand.
* It will become unsustainable as the docs grow and more contributors join.
* It doesn't stay in-sync with the actual latest docs because it requires a non-trivial amount of work to regenerate.
* End users only get a new version when they update the Python package (_If_ the PDF is actually re-ingested on upgrade).
* The PDF is generated from the order of the files in the directory, not the order in sidebar.json. This means that usage topics are placed before installation topics. This could cause context issues when ingested by an LLM because LLMs thrive on hierarchy.

This is not a sustainable solution for the long term. Some other ideas:

* Pass the raw mdx files from `/docs/` to OpenRAG's self-knowledge instead of a PDF. Drawbacks: Still loses the hierarchy set in sidebar.js, and loses the context of partial files.
* Auto-generate the OpenRAG docs PDF whenever `*/docs/*` changes. Drawbacks: Users only get the new version when they upgrade their OpenRAG version.
* Ingest the HTML by crawling the live website (when it is no longer private), possibly creating a flow to crawl the site and ingest the content. This helps enforce the hierarchy and it lets users rerun the flow at will to get updated documents.

I don't have the expertise to do this.

Questions:
* How to maintain the documentation included in the package
* How to distribute changes between releases
* consider airgapped vs nonairgapped

aimurphy:
## Background

OpenRAG includes a PDF of the OpenRAG docs site. This is ingested during Application Onboarding so users can ask OpenRAG about itself. Currently, after the first ingestion, this PDF is never updated again unless the user deletes and reinstalls a newer version of OpenRAG.

Additionally, there are [issues with the maintenance and utility of the docs PDF](https://github.com/langflow-ai/openrag/issues/575):

* The PDF is manually created. The .mdx files must be edited to remove syntax that isn’t compatible with PDF (`<Tabs>` and `<details>`). This must be done manually to preserve the information architecture; even with AI assistance, manual checks are required to ensure the AI executed the request correctly.
* The PDF cuts off parts of some pages, such as wide tables, resulting in lost context.
* Testing revealed that PDF ingestion is subpar to HTML ingestion in this case.
[...]

PR #1063, opened 2026-03-04
feat: Pdf doc build

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/575 
notes on the changes: https://ibm.ent.box.com/notes/2155130295925?s=u5jc90l84zbtouxfsk23bgq4cex168ue

https://github.com/user-attachments/assets/3b569b92-a3a1-48bb-8b54-bfe59340bc07


src/main.py
```diff
@@ -426,6 +505,358 @@ async def _ingest_default_documents_langflow(
...
+            anonymous_user.user_id, effective_jwt
+        )
+        if hasattr(session_manager, "_anonymous_jwt"):
+            effective_jwt = session_manager._anonymous_jwt
+
+    payload = {
+        "input_value": docs_url,
+        "input_type": "chat",
+        "output_type": "text",
+        "tweaks": {
```
edwinjosechittilappilly: We need to move away from using Tweaks.

src/main.py
```diff
@@ -426,6 +505,358 @@ async def _ingest_default_documents_langflow(
...
+        ),
+    }
+    add_provider_credentials_to_headers(headers, config)
+
+    logger.info(
+        "Running default URL docs ingestion flow",
+        docs_url=docs_url,
+        crawl_depth=crawl_depth,
+    )
+    resp = await clients.langflow_request(
```
edwinjosechittilappilly: can we use langflow service from openrag?

sdks/python/openrag_sdk/documents.py
```diff
@@ -136,11 +137,17 @@ async def delete(self, filename: str) -> DeleteDocumentResponse:
         Returns:
             DeleteDocumentResponse with deleted chunk count.
         """
-        response = await self._client._request(
-            "DELETE",
-            "/api/v1/documents",
-            json={"filename": filename},
-        )
+        try:
```
edwinjosechittilappilly, line 140: Not a change required in this PR, but it is a valid change though.

src/api/settings.py
```diff
@@ -1252,10 +1277,12 @@ async def _create_openrag_docs_filter(
     query_data = json.dumps({
         "query": "",
         "filters": {
-            "data_sources": ["openrag-documentation.pdf"],
+            # URL-based docs ingestion produces many source URLs.
+            # Filter by connector type to target the OpenRAG system docs set.
+            "data_sources": ["*"],
```
edwinjosechittilappilly: This is a wild card filter and would not be able to filter the documentation ingested.

src/api/settings.py
```diff
@@ -1252,10 +1277,12 @@ async def _create_openrag_docs_filter(
...
             "document_types": ["*"],
             "owners": ["*"],
-            "connector_types": ["*"],
+            "connector_types": ["system_default"],
```
edwinjosechittilappilly: This would filter not just openrag document but any default documents ingested during onbaording.

src/models/processors.py
```diff
@@ -803,3 +817,150 @@ async def process_item(
             file_task.updated_at = time.time()
             upload_task.failed_files += 1
             raise
+
+
+class LangflowUrlProcessor(TaskProcessor):
```
edwinjosechittilappilly: Good Class, we can reuse this in future for general URL ingestion also. I would suggest to move it to a seperate file itself like URL.py etc.

src/api/settings.py
edwinjosechittilappilly: may be we can add conector type as "openrag_docs" in settings default filter as well as in langflow url ingestion service instead of system_default.

Issue #1058, opened 2026-03-02
[Feature]: Add support for Python 3.14


mpawlow:
### Feature Area

Backend/API (Python/Starlette)

### Problem Description

(from @jlyoungblood)

If we support 3.11-3.14 that will be helpful for users. I suspect supporting only 3.13 will create friction for some users / potential users who:

- Are pinned by corporate IT (e.g., already moved to 3.14)
- Are inside legacy projects (e.g., trying to connect with Django app on 3.11)
- Use CI images that lag (e.g., Ubuntu LTS default Python)
- Depend on libraries not yet upgraded

See: https://github.com/langflow-ai/openrag/issues/1023#issuecomment-3976398300

#### Related Issues

- https://github.com/langflow-ai/openrag/issues/1023

### Proposed Solution

Add support for Python 3.14

### Alternatives Considered

N/A

### Priority

Would improve my workflow

### Additional Context

N/A

### Contribution

- [x] I would be willing to help implement this feature.
- [x] I can help test this feature once implemented.

### Checklist

- [x] I have searched existing issues and discussions to ensure this feature hasn't been requested before.

mpawlow:
@jlyoungblood 
- Currently, only adopting **Python 3.14** is tracked here
- However, we can expand the scope of this issue accordingly - depending on how much work is involved in adding support for **Python 3.11** and **Python 3.12**
   - I think an initial assessment is required here from the Development side
- Also, the main inhibitor for adopting **Python 3.11** and **Python 3.12** in https://github.com/langflow-ai/openrag/issues/1023 was the lack of explicit test coverage in our CI build pipeline
   - @prasanthcaibmcom was concerned about claiming support for those Python versions without explicit test coverage

PR #1078, opened 2026-03-09
feat: Set python-preference to only-managed

edwinjosechittilappilly:
Add python-preference = "only-managed" to the [tool.uv] section in pyproject.toml to prefer managed Python interpreters (rather than system/global Python). This helps ensure builds and packaging use the project's managed Python environment.
_______
This pull request makes a minor configuration change to the `pyproject.toml` file, specifying that only managed Python versions should be used for the project.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,5 +40,6 @@ openrag = "tui.main:run_tui"
 
 [tool.uv]
 package = true
+python-preference = "only-managed"
 
 


pyproject.toml
ricofurtado, line 10: Hi Edwing, there was a problem with PyArrow wheel not present in 3.14. Wouldn't be better to limit it to 3.13 to avoid the issue?

jlyoungblood: I've also run into other Docling dependency issues when trying to use 3.14 e.g. pydantic-core. Would be good to limit to 3.13 until this issue is addressed: #1058

mpawlow: ### (1a) [Normal] Should fix the version to Python 3.13

- Agree with @ricofurtado and @jlyoungblood here

pyproject.toml
```diff
@@ -40,5 +40,6 @@ openrag = "tui.main:run_tui"
 
 [tool.uv]
 package = true
+python-preference = "only-managed"
```
mpawlow, line 43: ### (1b) [Normal] Consider the additional disk size footprint / image size (for Cloud)

- A drawback is the increase in disk size requirements
- e.g.

Image

- If this an acceptable increase, then please feel free to ignore this concern

mpawlow (changes requested):
@edwinjosechittilappilly 

## Code Review 1

- See: PR comments: (1a), (1b)

PR #1090, opened 2026-03-10
feat: Add IBM Watsonx Data (AMS) auth support

edwinjosechittilappilly:
Introduce support for IBM Watsonx Data (AMS) embedded authentication via the ibm-lh-console-session cookie. Add IBM_AUTH_ENABLED and IBM_JWT_PUBLIC_KEY_URL settings and .env.example entries; pre-fetch and cache IBM JWT public key at startup. Implement JWT validation and key rotation handling in new auth.ibm_auth module. Wire IBM auth into dependencies (extract/validate cookie, set request.state.user), main service initialization, auth_service (prevent Google OAuth login when IBM auth is active and expose ibm_auth_mode in user info), and session_manager (use IBM JWT as-is). Update frontend auth context and login page to recognize IBM auth mode and disable local login/logout and redirect accordingly. Adjust logout endpoint to no-op when IBM auth is enabled.


mfortman11 (approved):

src/api/auth.py
```diff
@@ -108,15 +110,54 @@ async def auth_logout(
...
+    """
+    from config.settings import IBM_AUTH_ENABLED, IBM_SESSION_COOKIE_NAME
+
+    if not IBM_AUTH_ENABLED:
+        raise HTTPException(status_code=404, detail="IBM auth is not enabled")
+
+    response = JSONResponse({"status": "ok"})
+
+    # Local dev fallback only — in production Traefik sets the session cookie.
+    if not request.cookies.get(IBM_SESSION_COOKIE_NAME):
```
mpawlow, line 165: ### (1a) [Critical] Security: Authentication bypass via ibm-auth-basic cookie

- `Problem`: The `ibm_login` endpoint stores the raw `Authorization: Basic <base64>` header in the `ibm-auth-basic cookie. _get_ibm_user` then reads that cookie, decodes only the username (discarding the password entirely at decoded.split(":", 1)[0]), and constructs a fully authenticated User. 
- There doesn't appears to be any credential validation 
   - i.e. Any caller can POST a crafted Basic header to `/auth/ibm/login` and authenticate as any arbitrary username.

src/dependencies.py
```diff
@@ -85,6 +85,68 @@ def get_flows_service(services: dict = Depends(get_services)):
...
+                name=claims.get("display_name", claims.get("username", claims["sub"])),
+                picture=None,
+                provider="ibm_ams",
+                jwt_token=ibm_token,
+            )
+            request.state.user = user
+            return user
+
+    # ── Option 2: ibm-auth-basic cookie (local dev, no Traefik) ─────────
+    auth_header = request.cookies.get("ibm-auth-basic", "")
```
mpawlow, line 124: ### (1b) [Critical] Authentication bypass via ibm-auth-basic cookie

- See PR comment: **(1a) [Critical] Security: Authentication bypass via ibm-auth-basic cookie**

src/auth/ibm_auth.py
```diff
@@ -0,0 +1,74 @@
...
+    validated the JWT. Returns the claims dict, or None if decoding fails.
+    """
+    try:
+        return jwt.decode(token, options={"verify_signature": False})
+    except jwt.InvalidTokenError as exc:
+        logger.warning("IBM JWT decode failed", error=str(exc))
+        return None
+
+
+async def fetch_ibm_public_key(url: str):
```
mpawlow, line 34: ### (1c) [Critical] Security: validate_ibm_jwt and fetch_ibm_public_key appear to be dead code

- `Problem`: IBM JWTs are never cryptographically verified
- The module docstring advertises RS256 validation via `IBM_JWT_PUBLIC_KEY_URL`, and `validate_ibm_jwt` + `fetch_ibm_public_key` are fully implemented. 
   - However, the only function called in the actual request path (`_get_ibm_user` Option 1) is `decode_ibm_jwt`, which uses `options={"verify_signature": False}`. 
   - The validation infrastructure appears to be completely disconnected.
   - A forged or expired IBM JWT would be accepted verbatim if it bypasses Traefik and reaches the backend directly. 
   - `IBM_JWT_PUBLIC_KEY_URL` misleads operators into believing verification is
  active.

src/auth/ibm_auth.py
```diff
@@ -0,0 +1,74 @@
...
+
+from utils.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+# Module-level cache; populated by fetch_ibm_public_key() if called.
+_cached_public_key = None
+
+
+def decode_ibm_jwt(token: str) -> dict | None:
```
mpawlow, line 21: ### (1d) [Critical] Security: validate_ibm_jwt and fetch_ibm_public_key appear to be dead code

- See: PR comment: **(1c) [Critical] Security: validate_ibm_jwt and fetch_ibm_public_key appear to be dead code**

src/dependencies.py
```diff
@@ -85,6 +85,68 @@ def get_flows_service(services: dict = Depends(get_services)):
...
+    If *required* is False, returns None instead of raising.
+    """
+    import base64
+    import auth.ibm_auth as ibm_auth
+    from config.settings import IBM_SESSION_COOKIE_NAME
+
+    # ── Option 1: ibm-openrag-session cookie (production via Traefik) ───
+    ibm_token = request.cookies.get(IBM_SESSION_COOKIE_NAME)
+    if ibm_token:
+        claims = ibm_auth.decode_ibm_jwt(ibm_token)
```
mpawlow, line 110: ### (1e) [Critical] Security: validate_ibm_jwt and fetch_ibm_public_key appear to be dead code

- See: PR comment: **(1c) [Critical] Security: validate_ibm_jwt and fetch_ibm_public_key appear to be dead code**

src/api/auth.py
```diff
@@ -108,15 +110,54 @@ async def auth_logout(
...
+        auth_header = request.headers.get("Authorization", "")
+        if auth_header.startswith("Basic "):
+            response.set_cookie(
```
mpawlow, line 169: ### (1f) [Major] ibm-auth-basic cookie missing secure=True

- `Problem`: The `ibm_login` endpoint sets the ibm-auth-basic cookie with `httponly=True` and `samesite="lax"` but no `secure=True`. 
- The cookie value is `Basic <base64(username:password)>` 
   - Base64 is not encryption and the plaintext password is trivially recoverable (atob() in the browser).
   - Without secure=True, this credential cookie will be transmitted over unencrypted HTTP connections, exposing recoverable passwords to any network observer.

src/dependencies.py
```diff
@@ -85,6 +85,68 @@ def get_flows_service(services: dict = Depends(get_services)):
     return services["flows_service"]
 
 
+# ─────────────────────────────────────────────
+# IBM AMS authentication helper
+# ─────────────────────────────────────────────
+
+def _get_ibm_user(request: Request, required: bool) -> Optional["User"]:
```
mpawlow, line 92: ### (1g) [Normal] _get_ibm_user never registers the user in session_manager.users

- `Problem`: The API key auth path has an explicit comment and code: `if user.user_id not in session_manager.users: session_manager.users[user.user_id] = user` - establishing the pattern that all auth paths must register users so get_effective_jwt_token can look them up.
- `_get_ibm_user` does not appear to do this. 
- When downstream code calls `get_effective_jwt_token(ibm_user_id, None)`, the `get_user(user_id)` lookup returns `None` for IBM users, causing `get_effective_jwt_token` to return `None` - which silently produces an unauthenticated OpenSearch client rather than explicitly failing.
- `Proposed Solution`: Follow API key path pattern in src/dependencies.py lines 283–284 (?)

src/api/auth.py
```diff
@@ -108,15 +110,54 @@ async def auth_logout(
...
+    """Logout user by clearing auth cookie(s)"""
+    from config.settings import IBM_AUTH_ENABLED, IBM_SESSION_COOKIE_NAME
+
     await TelemetryClient.send_event(Category.AUTHENTICATION, MessageId.ORB_AUTH_LOGOUT)
+
     response = JSONResponse(
         {"status": "logged_out", "message": "Successfully logged out"}
     )
 
+    if IBM_AUTH_ENABLED:
```
mpawlow, line 118: ### (1h) [Normal] auth_logout deletes a Traefik-managed cookie (false logout confidence)

- `Problem`: auth_logout calls `response.delete_cookie(key=IBM_SESSION_COOKIE_NAME, ...)` to clear ibm-openrag-session. 
   - But this cookie is owned and set by Traefik, not the backend. 
   - The backend's `Set-Cookie: ibm-openrag-session=; expires=<past>` header clears it from the browser, but on the very next proxied request Traefik re-validates with AMS (which still has an active session) and re-injects the cookie. 
   - True session termination requires an AMS-side call. 
   - The logout appears to succeed but doesn't actually terminate the IBM session.
   - Impact is partially mitigated by the frontend already no-opting logout when isIbmAuthMode is true.

frontend/contexts/auth-context.tsx
```diff
@@ -146,10 +171,35 @@ export function AuthProvider({ children }: AuthProviderProps) {
...
+        Authorization: "Basic " + btoa(username + ":" + password),
+      },
+    });
+    console.log(
+      "[loginWithIbm] response status:",
+      response.status,
+      "ok:",
+      response.ok,
+    );
+    console.log(
```
mpawlow, line 188: ### (1i) [Minor] Debug console.log statements left in production, including document.cookie dump

- `Problem`: Multiple console.log debug statements were left in this file
   - Recommend evaluating all `console.log` statements and consider removing if done debugging
- `Problem`: Dumping all browser cookies (including ibm-openrag-session and any other session tokens) to the DevTools console unconditionally on every IBM login. 
- Note: No remote exploitability, but exposes session state on shared or monitored machines.
- This is a Minor severity issue. Please feel free to optionally fix or ignore

mpawlow (changes requested):
@edwinjosechittilappilly 

## Code Review 1

- See PR comments: (1a) to (1i)

src/api/auth.py
edwinjosechittilappilly: Proxy will be used to intercept the unautheticated requests. so Only authenticated requests will be available for BE.

cc. @zzzming

src/auth/ibm_auth.py
edwinjosechittilappilly: Agreed!
This was an initial approach, kept the legacy code for future tasks. we can remove it later.

src/api/auth.py
edwinjosechittilappilly: in a follow up PR we would be reading the relevant creds from aunthenticated header from Proxy. Which will be used to authenticate Opensearch.

src/api/auth.py
edwinjosechittilappilly: This flow is only for local development

edwinjosechittilappilly: # secure =True not needed for local development

src/dependencies.py
edwinjosechittilappilly: This is the second part of the Auth solution which will be in seperate PR for Accessing the IBM Login Creds from headers for an Authenticated Request from Traefic

src/api/auth.py
edwinjosechittilappilly: Added a partial logout and proper message for handling this too.

PR #1104, opened 2026-03-12
feat: Pdf doc build fix

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/575
notes: https://ibm.ent.box.com/notes/2155130295925?s=u5jc90l84zbtouxfsk23bgq4cex168ue

https://github.com/user-attachments/assets/997dafa3-540a-47ba-ba27-cf89c6455591

### DIFF
--- a/src/api/documents.py
+++ b/src/api/documents.py
@@ -92,7 +92,7 @@ async def delete_documents_by_filename_core(
                 "error": (
                     "Access denied: insufficient permissions"
                     if status_code == 403
-                    else error_str
+                    else "An internal error has occurred while deleting documents"
                 ),
             },
             status_code,


src/services/search_service.py
```diff
@@ -506,4 +506,4 @@ async def search(
         set_search_limit(limit)
         set_score_threshold(score_threshold)
 
-        return await self.search_tool(query, embedding_model=embedding_model)
+        return await self.search_tool(query, embedding_model=embedding_model)
```
mpawlow, line 509: ### (1a) [Critical] Owner filter queries non-existent field `owner_name.keyword`

- Not sure if this is a pre-existing issue; but, affects the stated goal (point 8): "filter normalization… keep payloads consistent."
- `search_service.py` maps the owners filter key to "owner_name.keyword", but the OpenSearch index schema only defines "owner" as an explicit keyword field. 
- The Langflow `opensearch_multimodal.py` component correctly uses "owner". 
- The two paths disagree - owner-based filtering through the search service will either silently return no results or rely on a dynamically-mapped field that may not exist on older indexed documents.
- Need to list the affected files / lines here; since, not in the diff of changes
   - `src/services/search_service.py`: line 80
      - BUG: "owners": "owner_name.keyword",
   - `src/services/search_service.py`: line 239
      - BUG: "owners": "owner_name.keyword",
   - `src/services/search_service.py`: line 335
      - BUG: "owners": "owner_name.keyword",
   - `flows/components/opensearch_multimodal.py`: line 1204
      - CORRECT: "owners": "owner",
   - `src/config/settings.py`: line 158
      - CORRECT: "owner": {"type": "keyword"},
         - only "owner" is an explicitly mapped keyword field
- Recommended Fix: Change all three occurrences in `search_service.py` from "owner_name.keyword" to "owner".

src/api/v1/documents.py
```diff
@@ -102,6 +102,15 @@ async def delete_document_endpoint(
 
         deleted_count = result.get("deleted", 0)
         logger.info(f"Deleted {deleted_count} chunks for filename {filename}", user_id=user.user_id)
+        if deleted_count == 0:
```
mpawlow: ### (1b) [Critical] The v1 and non-v1 delete API endpoints return inconsistent response shapes

- Both the Python and TypeScript SDKs wrap both endpoints. 
- The two backend endpoints return different fields on success and on 404. 
- The v1 endpoint omits filename and message on success; 
- Neither of the two backend endpoints includes a consistent error field across both success and failure shapes.
- e.g. `v1/documents.py`: "error" only on 404, not on 200:
```json
  {"success": False, "deleted_chunks": 0, "error": "No matching..."}   # 404
```
```json
  {"success": True,  "deleted_chunks": deleted_count}                   # 200
```
- e.g. `documents.py`: different field set entirely:
```json
  {"success": True,  "deleted_chunks": deleted_count, "filename": ..., "message": ...}
```

- SDKs' `DeleteDocumentResponse` model only defines success and deleted_chunks, silently dropping error, filename, and message. 
- A caller receiving success: False with a 200 status code (if that path is hit) has no error detail.

src/api/documents.py
```diff
@@ -96,6 +96,14 @@ async def delete_documents_by_filename(
         deleted_count = result.get("deleted", 0)
         logger.info(f"Deleted {deleted_count} chunks for filename {body.filename}", user_id=user.user_id)
 
+        if deleted_count == 0:
+            return JSONResponse({
```
mpawlow: ### (1b) [Critical] The v1 and non-v1 delete API endpoints return inconsistent response shapes

- Same feedback as & linked to PR comment for `src/api/v1/documents.py`

sdks/python/tests/test_integration.py
```diff
@@ -277,12 +277,19 @@ async def test_ingest_document(self, client, test_file: Path):
...
         """Test document deletion."""
         # First ingest (wait for completion)
-        await client.documents.ingest(file_path=str(test_file))
+        ingest_result = await client.documents.ingest(file_path=str(test_file))
 
         # Then delete
         result = await client.documents.delete(test_file.name)
 
-        assert result.success is True
+        # If ingestion produced indexed chunks, delete should succeed.
```
mpawlow, line 285: ### (1c) [Normal] No test for the idempotent-delete behavior that was explicitly added 

- Related to (point 10): "Python and TypeScript SDKs treat that 404 as non-throwing 'nothing deleted' response."
- The integration tests were updated, but neither the Python nor TypeScript test suite includes a case that deletes a document that was never ingested and asserts no exception is raised.
- The existing test only conditionally validates delete-after-ingest - it never exercises the actual idempotent guarantee.
- e.g. Test scenario only delete-after-ingest:
```python
      assert result.success is False   # this isn't really testing idempotency
```
- e.g. Test scenario is missing
```python
  result = await client.delete("document_that_was_never_ingested.pdf")
  assert result.success is False   # ← must not raise, just return False
```

sdks/typescript/tests/integration.test.ts
```diff
@@ -272,12 +272,20 @@ describe.skipIf(SKIP_TESTS)("OpenRAG TypeScript SDK Integration", () => {
...
     it("should delete a document", async () => {
       // First ingest (wait for completion)
-      await client.documents.ingest({ filePath: testFilePath });
+      const ingestResult = await client.documents.ingest({ filePath: testFilePath });
 
       // Then delete
       const result = await client.documents.delete(path.basename(testFilePath));
 
-      expect(result.success).toBe(true);
+      // If ingestion produced indexed chunks, delete should succeed.
```
mpawlow, line 280: ### (1c) [Normal] No test for the idempotent-delete behavior that was explicitly added 

- Similar feedback as & linked to PR comment for `sdks/python/tests/test_integration.py `

tests/unit/test_main_docs_signature.py
```diff
@@ -0,0 +1,77 @@
+import hashlib
```
mpawlow, line 1: ### (1d) [Normal] Missing test coverage for the new refresh endpoint

- Related to (point 6) and (point 4) in Box Note
- `Problem`: `tests/unit/test_main_docs_signature.py` correctly `tests _get_remote_docs_signature()`; but, the new `POST /openrag-docs/refresh` endpoint has zero test coverage
   - e.g. no unit test for the handler
   - e.g. no test for RefreshOpenRAGDocsResponse shape
   - e.g. no test verifying auth is required

src/api/settings.py
```diff
@@ -1756,3 +1785,38 @@ async def update_docling_preset(
     except Exception as e:
         logger.error("Failed to update docling settings", error=str(e))
         raise HTTPException(status_code=500, detail=f"Failed to update docling settings: {str(e)}")
+
+
+async def refresh_openrag_docs(
```
mpawlow, line 1806: ### (1d) [Normal] Missing test coverage for the new refresh endpoint

- Related to and linked to PR comment for `tests/unit/test_main_docs_signature.py`

src/services/langflow_file_service.py
```diff
@@ -230,6 +232,154 @@ async def run_ingestion_flow(
             raise
         return resp_json
 
+    async def run_url_ingestion_flow(
```
mpawlow, line 254: ### (1e) [Normal] Auto-heal silently swallows Langflow-down errors for the entire process lifetime

- Related to (point 3) in the Box Note
- `Problem`: If Langflow is temporarily unavailable at the moment of import, `create_resp.raise_for_status()` raises an `HTTPStatusError`.
   - Multiple locations in this code block below
- There is no retry and no re-attempt on subsequent calls 
- i.e. The URL ingest feature remains broken for the rest of that process lifetime with no way to recover short of a restart.
- The gap in the auto-heal process is that a transient Langflow outage at startup permanently disables URL ingestion
- `Recommendation`: permanent failures (e.g. 400 Bad Request on bad JSON) should be distinguished from transient ones (e.g. 502/503) so callers can retry the latter.
- Note: I'm not sure how applicable this problem is now because the **Backend** app blocks and waits for Langflow readiness at start-up

flows/components/opensearch_multimodal.py
```diff
@@ -1614,22 +1614,141 @@ def search(self, query: str | None = None) -> list[dict[str, Any]]:
...
+                return run_keyword_fallback(f"{reason}; no query embeddings available")
+
+            fallback_field = legacy_vector_field or "chunk_embedding"
+            logger.warning(
+                "KNN search failed for dynamic fields; falling back to legacy field '%s'.",
+                fallback_field,
+            )
+            self.log(f"[FALLBACK] Legacy vector fallback: field={fallback_field}; reason={reason}")
+
+            fallback_body = copy.deepcopy(body)
```
mpawlow, line 1703: ### (1f) [Normal] KNN fallback assumes body["query"]["bool"]["filter"] always exists 

- Related to (point 9) in the Box Note
- `Problem`: `run_legacy_vector_fallback` deep-copies body then directly assigns to `fallback_body["query"]["bool"]["filter"]` without verifying the key path exists. 
- If query, bool, or filter is absent (e.g. a keyword-only body), this raises KeyError and the fallback fails
```python
 fallback_body["query"]["bool"]["filter"] = [   # <--- KeyError if structure absent
```

src/models/url.py
```diff
@@ -0,0 +1,157 @@
...
+from .processors import TaskProcessor
+from .tasks import UploadTask, FileTask, TaskStatus
+
+logger = get_logger(__name__)
+
+
+class LangflowUrlProcessor(TaskProcessor):
+    """Processor for Langflow URL ingestion flow as a tracked task."""
+
+    def __init__(
```
mpawlow, line 15: ### (1g) [Minor] Missing type hints in LangflowUrlProcessor.__init__ 

- `Recommended`: Add type hints
```python
  def __init__(
      self,
      langflow_file_service,    # no type hint
      session_manager,          # no type hint
      docs_url: str,
      owner_user_id: str = None,  # should be Optional[str] = None
      jwt_token: str = None,      # should be Optional[str] = None
      ...
```
- Note: This is a Minor severity issue. Please feel free to optionally implement or ignore.

frontend/app/api/mutations/useRefreshOpenragDocs.ts
```diff
@@ -0,0 +1,39 @@
...
+const refreshOpenragDocs = async (): Promise<RefreshOpenRAGDocsResponse> => {
+  const response = await fetch("/api/openrag-docs/refresh", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+  });
+
+  if (!response.ok) {
+    const error = await response.json();
```
mpawlow: ### (1h) [Normal] useRefreshOpenragDocs error handler calls response.json() without guarding against non-JSON bodies

- `Problem`: If the server returns a non-JSON error body (HTML error page, nginx 502), `await response.json()` throws and the promise rejects with a JSON parse error instead of the intended fallback message "Failed to refresh OpenRAG docs".

frontend/app/onboarding/_components/onboarding-content.tsx
```diff
@@ -12,15 +12,17 @@ import type { Message, SelectedFilters } from "@/app/chat/_types/types";
...
 import { useChatStreaming } from "@/hooks/useChatStreaming";
+import { buildSearchPayloadFilters } from "@/lib/filter-normalization";
 
 import { OnboardingStep } from "./onboarding-step";
 import OnboardingUpload from "./onboarding-upload";
 
 // Filters for OpenRAG documentation
 const OPENRAG_DOCS_FILTERS: SelectedFilters = {
-  data_sources: ["openrag-documentation.pdf"],
+  data_sources: [],
```
mpawlow, line 23: ### (1i) [Normal] OPENRAG_DOCS_FILTERS typed as SelectedFilters but used as FilterInput 

- Related to (point 8) in the Box Note
- SelectedFilters requires all four properties. 
- buildSearchPayloadFilters() expects FilterInput (all optional).
- The structural overlap works today, but if SelectedFilters gains required fields that FilterInput doesn't share, this will silently pass wrong data.

frontend/app/api/mutations/useRefreshOpenragDocs.ts
```diff
@@ -0,0 +1,39 @@
+"use client";
+
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+
+interface RefreshOpenRAGDocsResponse {
```
mpawlow, line 5: ### (1j) [Minor] RefreshOpenRAGDocsResponse shape not validated at runtime

- `Problem`: `knowledge/page.tsx` accesses `result.message` directly after the mutation. 
   - If the backend returns a different shape (e.g. during an error condition), the access silently yields undefined with no user feedback.
- Note: This is a Minor severity issue. Please feel free to optionally implement or ignore.

src/api/settings.py
mpawlow, line 1806: ### (1k) [Minor] Refresh endpoint is web-UI only with no SDK equivalent

- Related to (point 4) in the Box Note
- Note: This discrepancy is intentional
- However, there is no SDK method and no documentation stating the endpoint is intentionally excluded from the SDK. 
- `Problem`: Anyone using the Python or TypeScript SDK who wants to trigger a refresh programmatically (e.g. in a CI pipeline after uploading docs) has no path to do so. 
- `Recommend`: Commenting or documenting somewhere whether this is expected to remain UI-only.
- Note: This is a Minor severity issue. Please feel free to optionally implement or ignore.

src/api/v1/documents.py
edwinjosechittilappilly: This is crticial
@mpawlow  See if its required to this PR if not we can remove it from this PR.
Feel free to push the changes with Olfa..
Good Catch

frontend/app/api/mutations/useRefreshOpenragDocs.ts
Wallgau: as discussed during standup ignoring minor issue for now, we can address them in another PR later

src/api/settings.py
Wallgau: as discussed during standup ignoring minor issue for now, we can address them in another PR later

src/models/url.py
Wallgau: as discussed during standup ignoring minor issue for now, we can address them in another PR later

mpawlow:
### Test Scenario: Startup refresh enabled

- ✅  PASS

#### Details

```
[2026-03-12T22:28:51.923689Z] [INFO   ] [main.py:941                   ] OpenRAG docs refresh skipped: remote signature unchanged
                                                                         - signature: etag="cc04b66fae3d9147f6c3e0b1bfd7985a"
```

PR #1113, opened 2026-03-12
feat: changes to ingestion flow - added docling chunker.

ricofurtado:
Changes in default ingestion flow so it can support docling-based chunking.

---

- Previous PR: https://github.com/langflow-ai/openrag/pull/900

### DIFF
--- a/flows/ingestion_flow.json
+++ b/flows/ingestion_flow.json
@@ -6,93 +6,6 @@
         "className": "",
         "data": {
           "sourceHandle": {
-            "dataType": "DoclingRemote",
-            "id": "DoclingRemote-Dp3PX",
-            "name": "dataframe",
-            "output_types": [
-              "DataFrame"
-            ]
-          },
-          "targetHandle": {
-            "fieldName": "data_inputs",
-            "id": "ExportDoclingDocument-zZdRg",
-            "inputTypes": [
-              "Data",
-              "DataFrame"
-            ],
-            "type": "other"
-          }
-        },
-        "id": "xy-edge__DoclingRemote-Dp3PX{œdataTypeœ:œDoclingRemoteœ,œidœ:œDoclingRemote-Dp3PXœ,œnameœ:œdataframeœ,œoutput_typesœ:[œDataFrameœ]}-ExportDoclingDocument-zZdRg{œfieldNameœ:œdata_inputsœ,œidœ:œExportDoclingDocument-zZdRgœ,œinputTypesœ:[œDataœ,œDataFrameœ],œtypeœ:œotherœ}",
-        "selected": false,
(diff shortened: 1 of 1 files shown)


flows/ingestion_flow.json
```diff
@@ -1,93 +1,6 @@
 {
```
mpawlow, line 1: ### (2a) [Blocker] flows/ingestion_flow.json is invalid JSON

- Confirmed by Python JSON parser:
```
 json.decoder.JSONDecodeError: Expecting ',' delimiter: line 2925 column 41 (char 255868)
```  
- The outputs array inside the `OpenSearchVectorStoreComponentMultimodalMultiEmbedding` node's template is unclosed
   - Fields from the ingest_data input template were spliced in, corrupting the structure. Langflow will fail to import the flow entirely.
- Root cause?: merge commit 9475c56c introduced a broken merge of the flow JSON

mpawlow, line 1: ### (2b) [Major] flows_service.py hardcodes "Split Text" display name

- The PR removes the `SplitText` node from the flow, but flows_service.py still references it by display name "Split Text". 
- Calls to `update_ingest_flow_chunk_size()` and `update_ingest_flow_chunk_overlap()` via the Settings API will silently do nothing since the node lookup will return no match.
- Affected lines: 911–916 and 918–927

mpawlow, line 1: ### (2c) [Major] langflow_file_service.py hardcodes "SplitText-QIKhg"

- The removed node SplitText-QIKhg is still referenced in langflow_file_service.py for per-run tweaks.
- Any per-ingestion chunkSize, chunkOverlap, or separator settings will be sent to a non-existent node and silently dropped by Langflow.
- Affected lines: 292–309

mpawlow, line 1: ### (2d) [Major] api/langflow_files.py hardcodes "SplitText-PC36h"

- A second stale `SplitText` node ID (`SplitText-PC36h`) is referenced in the file upload endpoint for tweaks. 
- Same silent-drop behavior as Issue (2c)
- Affected lines: 96–104

mpawlow, line 1: ### (2e) [Major] api/settings.py reads defaults from "SplitText-QIKhg"

- `GET /api/settings` parses the live Langflow flow to populate chunkSize, chunkOverlap, and separator defaults. 
- Since `SplitText-QIKhg` is gone, this code will never match and the defaults will always fall back to YAML-config values instead of live flow values.
- Affected lines: 277–290

mpawlow (changes requested):
@ricofurtado 

## Code Review 2

- See PR comments: (2a) to (2e)

PR #1149, opened 2026-03-13
fix: update OAuth parameters for improved consent handling and token …

ricofurtado:
Root cause was in auth state, not the upload page UI.

Google connection was being marked unauthenticated because token expiry was interpreted incorrectly (timezone-naive handling), so authenticate() returned False.

The current Google token file also has no refresh_token, so once access token expires it cannot auto-refresh.
I patched this end-to-end:

-Normalized stored OAuth expiry to naive UTC at load time (compatible with google-auth): oauth.py

-Saved callback expiry in UTC and preserve existing refresh_token when provider omits it on re-consent.

-Added Google OAuth params to request offline refresh token consistently (include_granted_scopes=true, prompt=consent) in both login and connector connect flows)

-Made some changes in UI, so Mike needs to validate.

### DIFF
--- a/frontend/app/api/mutations/useConnectConnectorMutation.ts
+++ b/frontend/app/api/mutations/useConnectConnectorMutation.ts
@@ -77,7 +77,8 @@ export const useConnectConnectorMutation = () => {
             result.oauth_config.redirect_uri,
           )}&` +
           `access_type=offline&` +
-          `prompt=select_account&` +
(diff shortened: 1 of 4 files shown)


edwinjosechittilappilly (changes requested):
Rico lets test this once you are free after reverting the changes in a headers changes in auth context and useConnect...

PR #1164, opened 2026-03-17
chore: Move and run Python SDK integration tests

edwinjosechittilappilly:
This pull request introduces improvements to integration and SDK testing workflows, enhances test output clarity, and adds a comprehensive QA checklist for the Python SDK. The changes also refine error handling for conversation deletion and streamline onboarding logic for SDK tests. The most significant updates are grouped below.

### Test Workflow & Output Enhancements

* Integration and SDK tests are now grouped and visually separated in the `Makefile` using GitHub Actions-style output (`::group::`), making CI logs easier to navigate. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R686-R694) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L755-R795) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R815-R824) [[4]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L871-R925) [[5]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L917-R952)
* SDK integration tests now run from `tests/integration/sdk/` instead of a single test file, and use a dedicated install step for the Python SDK. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L755-R795) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L871-R925) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L917-R952)

### SDK Integration Test Infrastructure

* Added `tests/integration/sdk/conftest.py` to provide shared fixtures, API key management, onboarding, and test file creation for SDK tests.
* Added a detailed QA checklist in `tests/integration/sdk/README.md` covering 51 tests across 8 domains for the Python SDK.
* Introduced `test_auth.py` for authentication and API key behavior tests in the SDK integration suite.
* Modified backend onboarding logic to skip in-process setup when `SDK_TESTS_ONLY=true`, preventing unwanted state resets during SDK integration tests.

### Error Handling Improvements

* Improved conversation deletion logic in the backend (`chat_service.py` and `chat.py`) to return `False` if the conversation does not exist, and to respond with HTTP 404 when appropriate. [[1]](diffhunk://#diff-c7186047feb33fdedd8a4779753b0661f3c60795c1d6707d0dedc7753aaf6656L493-R502) [[2]](diffhunk://#diff-a4f7adb8dc47fe3831c1836b8dbc26ebdde5b05d83708db56e480e063bf024c7R229-R230) [[3]](diffhunk://#diff-a656d24ea3c40e5b3ba0c8c3e11af85a45cad1ba6f2ffd9ddf83af25408a6739L610-R624)

### Test Scoping & Directory Structure

* Integration tests for the core system are now scoped to `tests/integration/core/` for both CI and local runs, separating them from SDK tests. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L668-R672) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L755-R795) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L871-R925)

---

These changes collectively improve test clarity, reliability, and coverage for both core and SDK integration workflows.

### DIFF
--- a/tests/integration/sdk/test_chat.py
+++ b/tests/integration/sdk/test_chat.py
@@ -1,6 +1,5 @@
 """Tests for the chat endpoint — non-streaming, streaming, conversations, and RAG."""
 
-import asyncio
 import os
 from pathlib import Path
 
@@ -116,8 +115,6 @@ class TestChat:
         result = await client.documents.ingest(file_path=str(test_file))
         if result.status == "failed" or result.successful_files == 0:
             pytest.skip("Document ingestion failed — cannot test RAG sources")
-
-        await asyncio.sleep(3)
 
         response = await client.chat.create(
             message="What is the color of the dancing animals mentioned in my documents?"
--- a/tests/integration/sdk/test_search.py
+++ b/tests/integration/sdk/test_search.py
@@ -1,6 +1,5 @@
 """Tests for the search endpoint."""
 
-import asyncio
 import os
 from pathlib import Path
 
@@ -19,7 +18,6 @@ class TestSearch:
     async def test_search_query(self, client, test_file: Path):
         """A basic search query returns a results list."""
         await client.documents.ingest(file_path=str(test_file))
-        await asyncio.sleep(2)
 
         results = await client.search.query("purple elephants dancing")
         assert results.results is not None
@@ -32,7 +30,6 @@ class TestSearchExtended:
     async def test_search_with_limit(self, client, test_file: Path):
         """limit parameter caps the number of results returned."""
         await client.documents.ingest(file_path=str(test_file))
-        await asyncio.sleep(2)
 
         results = await client.search.query("test", limit=1)
         assert results.results is not None
@@ -42,7 +39,6 @@ class TestSearchExtended:
     async def test_search_with_high_score_threshold_returns_empty(self, client, test_file: Path):
         """A score_threshold of 0.99 should filter out most or all results."""
         await client.documents.ingest(file_path=str(test_file))
-        await asyncio.sleep(2)
 
         results = await client.search.query("test", score_threshold=0.99)
         assert results.results is not None
@@ -66,9 +62,8 @@ class TestSearchExtended:
 
     @pytest.mark.asyncio
     async def test_search_returns_result_fields(self, client, test_file: Path):
-        """Each search result must have content populated as a string."""
+        """Each search result must have text populated as a string."""
         await client.documents.ingest(file_path=str(test_file))
-        await asyncio.sleep(2)
 
         results = await client.search.query("purple elephants dancing", limit=5)
         for result in results.results:
--- a/tests/integration/sdk/test_e2e.py
+++ b/tests/integration/sdk/test_e2e.py
@@ -1,6 +1,5 @@
 """End-to-end tests covering full multi-step SDK workflows."""
 
-import asyncio
 import os
 import uuid
 
@@ -29,8 +28,6 @@ class TestEndToEnd:
         ingest_result = await client.documents.ingest(file_path=str(file_path))
         if ingest_result.successful_files == 0:
             pytest.skip("Ingestion produced no indexed chunks — skipping E2E RAG test")
-
-        await asyncio.sleep(3)
 
         search_results = await client.search.query("flamingo Zephyr planet Xylox")
         assert search_results.results is not None
@@ -62,8 +59,6 @@ class TestEndToEnd:
         ingest_result = await client.documents.ingest(file_path=str(file_path))
         if ingest_result.successful_files == 0:
             pytest.skip("Ingestion produced no indexed chunks — skipping multiturn test")
-
-        await asyncio.sleep(3)
 
         r1 = await client.chat.create(message="What is the capital of Valdoria?")
         assert r1.chat_id is not None


src/services/chat_service.py
```diff
@@ -607,20 +607,21 @@ async def delete_session(self, user_id: str, session_id: str):
             from agent import delete_user_conversation
             local_deleted = await delete_user_conversation(user_id, session_id)
 
-            # Delete from Langflow using the monitor API
-            langflow_deleted = await self._delete_langflow_session(session_id)
-
-            success = local_deleted or langflow_deleted
-            error_msg = None
+            if not local_deleted:
```
mpawlow, line 610: ### (1a) [Critical] Backend errors silently mis-reported as HTTP 404

- `Problem`: `delete_user_conversation` in agent.py returns False both when the conversation does not exist and when an internal exception is caught (except Exception: return False). 
- The new chat_service.py code checks only if not local_deleted to set not_found=True. 
- A transient database error, storage timeout, or any unexpected backend exception will produce an HTTP 404 instead of a 500. 
- Callers (including the SDK) will silently conclude the conversation never existed, masking real failures.
- `Proposed Solution`: Distinguish "not found" from "error" in delete_user_conversation (e.g., raise, return a tri-state, or use exceptions). 
   - Only map a confirmed not-found result to 404; treat all other false returns as 500.

tests/integration/sdk/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def ensure_onboarding():
+    """Ensure the OpenRAG instance is onboarded before running tests."""
+    global _onboarding_done
+    if _onboarding_done:
+        return
+
+    onboarding_payload = {
```
mpawlow, line 32: ### (1b) [Major] ensure_onboarding fixture hardcodes OpenAI provider and will overwrite live configuration

- `Problem`: The fixture unconditionally POSTs llm_provider: openai / embedding_model: text-embedding-3-small. If the environment is configured for a different provider (Bedrock, Azure, etc.), this will silently break a working deployment. 
- `Proposed Solution`: Check current settings before POSTing onboarding, or read provider/model from environment variables. 
   - Re-raise or log caught exceptions rather than silencing them.
- `Question`: Assuming this hardcoding may be by design or at least temporary for the initial release of the test suite?
   - If so, please feel free to ignore this issue

tests/integration/sdk/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+        "embedding_provider": "openai",
+        "embedding_model": "text-embedding-3-small",
+        "llm_model": "gpt-4o-mini",
+    }
+
+    try:
+        response = httpx.post(
+            f"{_base_url}/api/onboarding",
+            json=onboarding_payload,
+            timeout=30.0,
```
mpawlow: ### (1c) [Major] Blocking httpx.post in a synchronous fixture blocks the async event loop

- `Problem`: ensure_onboarding is a plain def generator fixture, yet it calls httpx.post(...) synchronously. 
- In an async pytest session this blocks the event loop for up to 30 seconds on a slow or unreachable server.
- `Proposed Solution`: Make the fixture async def ensure_onboarding() and use httpx.AsyncClient, or wrap the sync call with anyio.to_thread.run_sync.

tests/integration/sdk/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+            print("[SDK Tests] Onboarding completed successfully")
+        else:
+            print(f"[SDK Tests] Onboarding returned {response.status_code}: {response.text[:200]}")
+    except Exception as e:
+        print(f"[SDK Tests] Onboarding request failed: {e}")
+
+    _onboarding_done = True
+
+
+def get_api_key() -> str:
```
mpawlow: ### (1d) [Normal] Thread-unsafe global API key cache; pytest.skip called outside a fixture

- `Problem`: get_api_key() uses a module-level _cached_api_key global with no lock.
- Under any parallel fixture execution, two callers can both observe _cached_api_key is None simultaneously, each call httpx.post(...), and create two API keys (leaking credentials).
- `Proposed Solution`: Use a threading.Lock around cache population, or convert this to a session-scoped async fixture that yields the key.

tests/integration/sdk/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+    """Get or create an API key for testing (cached for the session)."""
+    global _cached_api_key
+    if _cached_api_key is None:
+        response = httpx.post(
+            f"{_base_url}/api/keys",
+            json={"name": "SDK Integration Test"},
+            timeout=30.0,
+        )
+        if response.status_code == 401:
+            pytest.skip("Cannot create API key - authentication required")
```
mpawlow: ### (1e) [Normal] pytest.skip called outside a fixture

- `Problem`:  pytest.skip() is called from a non-fixture helper function, which can cause obscure DID NOT RAISE errors.

tests/integration/sdk/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+        assert response.status_code == 200, f"Failed to create API key: {response.text}"
+        _cached_api_key = response.json()["api_key"]
+    return _cached_api_key
+
+
+@pytest.fixture
+def client():
```
mpawlow: ### (1f) [Major] client fixture never calls close() and HTTP connections are leaked

- `Problem`: The client fixture returns an OpenRAGClient directly with no yield, so the underlying httpx.AsyncClient is never closed.
- Every test using this fixture leaks an open connection. 
- The test_auth.py tests correctly call await bad_client.close() in finally blocks, demonstrating awareness of this requirement, but the shared fixture does not follow the same pattern.
- `Proposed Solution`:
```python
  @pytest.fixture
  async def client():
      c = OpenRAGClient(base_url=_base_url, api_key=get_api_key())
      yield c
      await c.close()
```

tests/integration/sdk/test_search.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class TestSearch:
+    """Core search query tests."""
+
+    @pytest.mark.asyncio
+    async def test_search_query(self, client, test_file: Path):
+        """A basic search query returns a results list."""
+        await client.documents.ingest(file_path=str(test_file))
+        await asyncio.sleep(2)
```
mpawlow: ### (1g) [Major] Hardcoded asyncio.sleep delays may be brittle versus task polling

- `Multiple locations`: tests/integration/sdk/test_search.py:22,35,45,71; test_chat.py:120; test_e2e.py:33,66
- `Problem`: All search and RAG tests wait a fixed 2–3 seconds after ingestion before querying.
- On slow CI runners these delays will be insufficient, causing flaky failures. 
- On fast runners they waste time.
- The SDK provides wait_for_task() for proper polling.
- `Proposed Solution`: After ingest(wait=False), use await client.documents.wait_for_task(...) to poll until the ingestion task completes before proceeding.

tests/integration/sdk/test_chat.py
```diff
@@ -0,0 +1,184 @@
...
+        """delete() returns True for a conversation that exists."""
+        response = await client.chat.create(message="Test message for delete.")
+        assert response.chat_id is not None
+
+        result = await client.chat.delete(response.chat_id)
+
+        assert result is True
+
+    @pytest.mark.asyncio
+    async def test_chat_with_sources(self, client, test_file: Path):
```
mpawlow, line 113: ### (1h) [Normal] test_chat_with_sources does not clean up the ingested document

- `Problem`: The test ingests a document but never deletes it 
   - e.g. Not applicable for pytest.skip() exit. 
- The indexed chunks remain in OpenSearch, potentially polluting other tests' search results.
- `Proposed Solution`: Wrap ingestion and deletion in a try/finally block.

tests/conftest.py
```diff
@@ -26,15 +26,22 @@ async def onboard_system():
 
     This ensures the OpenRAG config is marked as edited and properly initialized
     so that tests can use the /settings endpoint.
+
+    Skips in-process backend setup when SDK_TESTS_ONLY=true (SDK tests talk to
```
mpawlow, line 30: ### (1i) [Major] Module-level backend imports run even when SDK_TESTS_ONLY=true

- See: tests/conftest.py:14–20
- `Problem`: Even with SDK_TESTS_ONLY=true, lines 14–20 unconditionally import config.settings, session_manager, and main at module load time.
- These will raise ModuleNotFoundError in an environment where backend packages are not installed, defeating the purpose of SDK_TESTS_ONLY.
- `Proposed Solution`
```python
  if os.environ.get("SDK_TESTS_ONLY") != "true":
      from config import settings
      from session_manager import ...
      # etc.
```

tests/integration/sdk/test_auth.py
```diff
@@ -0,0 +1,59 @@
+"""Tests for authentication and API key behaviour."""
+
+import os
+
+import pytest
+
+from .conftest import _base_url
```
mpawlow, line 7: ### (1j) [Minor] Test files import private _base_url directly instead of using the fixture

- Also, see: tests/integration/sdk/test_errors.py:9
- `Problem`: from .conftest import _base_url imports a module-level private variable directly, bypassing pytest's fixture mechanism. 
- If OPENRAG_URL is modified after module import, the test files will use a stale value.
- `Proposed Solution`: Use the base_url fixture already defined in conftest.py:74–77.
- Note: This issue is Minor severity. Please feel free to optionally fix or ignore

tests/integration/sdk/test_chat.py
```diff
@@ -0,0 +1,184 @@
...
+    os.environ.get("SKIP_SDK_INTEGRATION_TESTS") == "true",
+    reason="SDK integration tests skipped",
+)
+
+
+class TestChat:
+    """Core chat operation tests."""
+
+    @pytest.mark.asyncio
+    async def test_chat_non_streaming(self, client):
```
mpawlow, line 18: ### (1k) [Normal] Chat tests create conversations with no cleanup

- `Problem`: Tests creating conversations via client.chat.create(...) never delete them. In a long-lived development environment these accumulate indefinitely.
- `Proposed Solution`: Add await client.chat.delete(response.chat_id) in a finally block, or use a fixture for conversation lifecycle management.

tests/integration/sdk/test_filters.py
```diff
@@ -0,0 +1,95 @@
...
+    os.environ.get("SKIP_SDK_INTEGRATION_TESTS") == "true",
+    reason="SDK integration tests skipped",
+)
+
+
+class TestKnowledgeFilters:
+    """Test knowledge filter create, read, update, delete and usage."""
+
+    @pytest.mark.asyncio
+    async def test_knowledge_filter_crud(self, client):
```
mpawlow, line 17: ### (1l) [Minor] test_knowledge_filter_crud is very monolithic and any step failure leaks the filter resource

- `Problem`: The full CRUD lifecycle is a single test. If create succeeds but a later step fails, the filter is never deleted. 
- i.e. There's no try/finally around the delete call.
- `Proposed Solution`: Wrap the body in try/finally to ensure deletion, or split into separate tests with a session-scoped fixture managing filter lifecycle.
- Note: This issue is Minor severity. Please feel free to optionally fix or ignore

tests/integration/sdk/test_search.py
```diff
@@ -0,0 +1,76 @@
...
+        assert isinstance(results.results, list)
+
+    @pytest.mark.asyncio
+    async def test_search_unicode_query(self, client):
+        """Unicode and emoji characters in the query must not cause an error."""
+        results = await client.search.query("こんにちは 🦩 Ñoño résumé")
+        assert results.results is not None
+        assert isinstance(results.results, list)
+
+    @pytest.mark.asyncio
```
mpawlow, line 63: ### (1m) [Normal] test_search_returns_result_fields silently passes with zero results

- `Problem`: For result in results.results: is a no-op if the list is empty.
- The test passes green even though nothing about result fields was verified.
- `Proposed Solution`: Add assert len(results.results) > 0 before the loop.

tests/integration/sdk/conftest.py
```diff
@@ -0,0 +1,90 @@
+"""
```
mpawlow, line 1: ### (1n) [Minor] asyncio_mode not configured in root pyproject.toml

- Note: Feedback for `pyproject.toml` file
- Applicable to all SDK test files
- `Problem`: Tests use @pytest.mark.asyncio on every async method.
- The root pyproject.toml does not set asyncio_mode = "auto" (only sdks/python/pyproject.toml does). 
- Running pytest tests/integration/sdk/ from the project root may produce collection warnings or failures depending on the installed pytest-asyncio version.
- `Proposed Solution`: Add asyncio_mode = "auto" to the [tool.pytest.ini_options] section in the root pyproject.toml.
- Note: This issue is Minor severity. Please feel free to optionally fix or ignore

tests/integration/sdk/conftest.py
edwinjosechittilappilly: Hi Mike, Currently The Integration tests are using OpenAI. It is the reuse of the existing code.

in future we will improve it for multiple providers.

src/services/chat_service.py
edwinjosechittilappilly: The key fix in agent.py: removed the except Exception: return False wrapper that was conflating "not found" with storage errors. Storage errors now propagate up to delete_session's except Exception block which correctly returns a
  500, not a 404.

tests/integration/sdk/conftest.py
edwinjosechittilappilly: Made Three changes:                 

  1. ensure_onboarding — changed to async def decorated with @pytest_asyncio.fixture, uses httpx.AsyncClient with await. No longer blocks the event loop.                                                                                  
  2. get_api_key — replaced the sync helper with _fetch_api_key() as an async def that also uses httpx.AsyncClient. Cached in the same module-level variable, called with await from the client fixture.
  3. client — changed to @pytest_asyncio.fixture / async def with yield so it properly await c.close() after each test, and can await _fetch_api_key() without blocking.

tests/integration/sdk/conftest.py
edwinjosechittilappilly: Fixed with Above Fix

tests/integration/sdk/test_chat.py
edwinjosechittilappilly: Good Catch, We can remove the async io sleep since internally it used wait for...

edwinjosechittilappilly: All 7 asyncio.sleep calls removed across 3 files:

  - test_search.py — 4 sleeps removed (lines 22, 35, 45, 71); import asyncio dropped                                                                                                                                                       
  - test_chat.py — 1 sleep removed (line 120); import asyncio dropped
  - test_e2e.py — 2 sleeps removed (lines 33, 66); import asyncio dropped                                                                                                                                                                  
                                                                                                                                                                                                                                           
  ingest() defaults to wait=True, which already calls wait_for_task() internally and only returns once the task reaches a terminal state. No external sleep needed.

src/services/chat_service.py
mpawlow: 🚀

tests/integration/sdk/conftest.py
mpawlow: 🚀

tests/integration/sdk/conftest.py
mpawlow: 🚀

tests/integration/sdk/test_search.py
mpawlow: 🚀

tests/integration/sdk/conftest.py
mpawlow: 🚀

tests/integration/sdk/conftest.py
mpawlow: 🚀  Partially fixed.

- The module-level _cached_api_key global still has no asyncio.Lock, so two coroutines racing through _fetch_api_key() before either completes can create two API keys 

### (2a) [Minor] TOCTOU race in _fetch_api_key() - two coroutines can both create API keys

- `Problem`: In asyncio, two coroutines that both enter _fetch_api_key() before either reaches the await ac.post(...) checkpoint will both pass the if _cached_api_key is not None guard and both issue API key
creation requests, producing two keys on the server. The second key is permanently orphaned.
- `Proposed Solution`: Use an asyncio.Lock to guard the check-and-set:
```python
_api_key_lock = asyncio.Lock()
async def _fetch_api_key():
    global _cached_api_key
    async with _api_key_lock:
        if _cached_api_key is not None:
            return _cached_api_key
        ...
```
- Note: This issue is Minor severity. Please feel free to optionally fix or ignore

tests/integration/sdk/conftest.py
mpawlow: 📝

tests/integration/sdk/test_chat.py
mpawlow: 🚀

mpawlow: @edwinjosechittilappilly 
- Further analsysis: The TestChatExtended tests now correctly call client.chat.delete(). 
- `Problem`: However, the original TestChat class including still creates conversations with no cleanup.
   - test_chat_non_streaming,
   - test_chat_streaming_create,
   - test_chat_streaming_context_manager,
   - test_chat_text_stream,
   - test_chat_final_text,
   - test_chat_conversation_continuation,
   - test_list_conversations,
   - test_get_conversation,
   - test_stream_chat_id_available_after_iteration

tests/conftest.py
mpawlow: 📝

tests/integration/sdk/test_auth.py
mpawlow: 📝

tests/integration/sdk/test_filters.py
mpawlow: 📝

tests/integration/sdk/test_search.py
mpawlow: 📝

tests/integration/sdk/conftest.py
mpawlow: 📝

tests/conftest.py
```diff
@@ -26,15 +26,22 @@ async def onboard_system():
...
+    an already-running external stack and must not wipe its state).
     """
+    if os.environ.get("SDK_TESTS_ONLY") == "true":
```
mpawlow, line 33: ### (2b) [Major] tests/conftest.py SDK_TESTS_ONLY guard is inside the fixture body; module-level imports already executed at collection time

- `Related Lines`: ~18–20
- `Problem`: The fix added a yield; return guard inside the onboard_system fixture body, but from config.settings import clients, from session_manager import SessionManager, and from main import generate_jwt_keys are module-level imports that execute during pytest's collection phase, before any fixture runs. 
- In a pure SDK-only environment without backend packages installed, pytest fails during collection with ModuleNotFoundError.
- `Proposed Solution`: Move the imports inside the fixture body, guarded by the SDK_TESTS_ONLY check:
```python
@pytest.fixture(scope="session", autouse=True)
def onboard_system():
    if os.environ.get("SDK_TESTS_ONLY") == "true":
        yield
        return
    from config.settings import clients
    from session_manager import SessionManager
    from main import generate_jwt_keys
    ...
```

src/api/connectors.py
```diff
@@ -200,6 +205,51 @@ async def connector_sync(
...
+                    all_file_ids = []
+                    page_token = None
+                    while True:
+                        result = await connector.list_files(page_token=page_token)
+                        for f in result.get("files", []):
+                            all_file_ids.append(f["id"])
+                        page_token = result.get("next_page_token")
+                        if not page_token:
+                            break
+                finally:
```
mpawlow, line 231: ### (2c) [Normal] connector_sync mutation of connector.bucket_names is not concurrency-safe

- `Problem`: If two concurrent sync requests arrive for the same connector with different bucket_filter values, both requests write to connector.bucket_names on what is likely a shared/cached connector instance.
- The finally: connector.bucket_names = original_buckets is per-request; whichever finally runs last wins, meaning one request may use the other's bucket filter mid-flight.
- `Proposed Solution`: Deep-copy the connector before mutating it, or use a per-request lock keyed on the connector's connection ID.

edwinjosechittilappilly: can we create this as an ticket so we can have it in backlog?
Good Catch!

tests/conftest.py
edwinjosechittilappilly: Good catch I belive I already moved it inside onboard_system ?

PR #1167, opened 2026-03-17
fix: persist Langflow database across container restarts

### LINKED ISSUE #1127: [Bug]: Langflow flow edits are not persisting after restart.
### OpenRAG Version

0.3.0

### Deployment Method

uv add (installed in project)

### Operating System

macOs 15.6 

### Python Version

python 3.13

### Affected Area

Settings (configuration, model providers), Infrastructure (Docker, OpenSearch, Langflow)

### Bug Description

Issue Summary : Langflow flow edits are not persisting after restart.

Edits made to flows within Langflow do not persist after restarting the OpenRAG application. The changes appear correctly after saving the flow, but once the containers are restarted, the flow reverts to the previous state and the saved changes

### Steps to Reproduce

1. Open OpenRAG and launch the Langflow UI.
2. Open any existing flow, such as the chat flow.
3. Make a visible change in the flow. Example: added the chat output

Image

4. Verify the change is visible in Langflow before restart.
5. Restart the container
6. Open Langflow again after restart.
7. Open the same flow.
8. Check whether the saved change is still present.

Image

 

### Expected Behavior

The saved flow changes should remain available after restart.

### Actual Behavior

The saved changes are no longer available after restart, and the flow reverts to the previous/default state.

### Relevant Logs

```shell

```

### Screenshots

Saved changes before restart 
Image

After restart 

Image

### Additional Context

_No response_

### Checklist

mpawlow:
### Issue

- Fixes #1127

### Summary

- Fixes Langflow flow edits not being persisted across container restarts by replacing the LANGFLOW_LOAD_FLOWS_PATH blind-upsert mechanism with a create-only ensure_flows_exist() approach that only seeds missing flows, preserving any user edits made in the Langflow UI
- Corrects the Langflow database and config directory mount path so the SQLite database is actually stored on the host-mounted volume and survives container restarts

### Root Cause Fix: Langflow Data Persistence

- Changed the docker-compose.yml volume mount for Langflow from /root/.langflow (the old default home directory) to /app/langflow-data (the path the langflow user actually owns inside the container), ensuring the SQLite database is written to the host-mounted volume
- Replaced the LANGFLOW_LOAD_FLOWS_PATH env var with LANGFLOW_CONFIG_DIR, which sets Langflow's config directory rather than triggering a blind flow upsert on every startup
- Made LANGFLOW_DATABASE_URL configurable via .env with a sensible default; supports override with a PostgreSQL URL for production deployments
- Pre-creates the /app/langflow-data directory in Dockerfile.langflow so it is owned by the correct UID (1000) before the volume is mounted

### New ensure_flows_exist() Behavior

- Added FlowsService.ensure_flows_exist() — a create-only method called at startup that checks whether each configured flow exists in the Langflow database (via GET /api/v1/flows/{id}) and only creates it (via PUT) if a 404 is returned
- Existing flows are never patched or overwritten, preserving user edits made in the Langflow UI
- Replaces LANGFLOW_LOAD_FLOWS_PATH, which performed a blind upsert on every container start and silently discarded user customizations
- startup_tasks() in src/main.py now calls ensure_flows_exist() before check_flows_reset(), and filters newly-seeded flows out of the reset-detection logic to avoid false positives

### Configuration & TUI

- Added LANGFLOW_DATA_PATH as a configurable field in the TUI setup wizard (src/tui/config_fields.py, src/tui/managers/env_manager.py, src/tui/screens/config.py), with a directory picker matching the existing OpenSearch data path pattern
- Default TUI value: $HOME/.openrag/data/langflow-data
- LANGFLOW_DATA_PATH added to .env.example and frontend/.env.test.example
- LANGFLOW_DATA_PATH documented in docs/docs/reference/configuration.mdx

### CI / Developer Experience

- Updated Makefile: added ensure-langflow-data target that pre-creates ./langflow-data with world-writable permissions; all dev, dev-cpu, dev-local, dev-local-cpu, dev-branch, dev-branch-cpu, test-ci, test-ci-local, and restart-dev targets depend on it
- Updated factory-reset to also remove the langflow-data directory
- Updated .github/workflows/test-e2e.yml and test-integration.yml cleanup steps to also delete langflow-data (root-owned files) via an Alpine container
- Updated scripts/setup-e2e.sh to pre-create and chmod 777 the langflow-data directory before starting infrastructure
- Added langflow-data/ to .gitignore

### Kubernetes

- Updated kubernetes/helm/openrag/templates/langflow/deployment.yaml to replace LANGFLOW_LOAD_FLOWS_PATH with LANGFLOW_CONFIG_DIR, consistent with the Docker Compose change

Builds on #1129 

Co-authored by @JasonOA888

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -682,6 +682,7 @@ test-integration: ## Run integration tests (requires infrastructure)
 	uv run pytest tests/integration/core/ -v
 
 test-ci: ensure-langflow-data ## Start infra, run integration + SDK tests, tear down (uses DockerHub images)
+	@chmod 777 langflow-data
 	@set -e; \
 	echo "$(YELLOW)Installing test dependencies...$(NC)"; \
 	uv sync --group dev; \
@@ -811,6 +812,7 @@ test-ci: ensure-langflow-data ## Start infra, run integration + SDK tests, tear 
 	exit $$TEST_RESULT
 
 test-ci-local: ensure-langflow-data ## Same as test-ci but builds all images locally
+	@chmod 777 langflow-data
 	@set -e; \
 	echo "$(YELLOW)Installing test dependencies...$(NC)"; \
 	uv sync --group dev; \
--- a/scripts/setup-e2e.sh
+++ b/scripts/setup-e2e.sh
@@ -49,6 +49,11 @@ echo "Starting E2E Setup using $E2E_ENV..."
 echo "Cleaning up..."
 make factory-reset FORCE=true ENV_FILE=$E2E_ENV
 
+# Pre-create langflow-data as world-writable so the Langflow container (UID 1000)
+# and the runner (UID 1001) can both access it, regardless of Docker's :U flag behavior.
+mkdir -p langflow-data
+chmod 777 langflow-data
+
 # Start infrastructure using make (this will use the new .env)
 echo "Starting infrastructure..."
 make dev-local-cpu ENV_FILE=$E2E_ENV


.env.example
```diff
@@ -58,6 +58,11 @@ OPENSEARCH_PASSWORD=
 # Default: ./opensearch-data
 OPENSEARCH_DATA_PATH=./opensearch-data
 
+# Path to persist Langflow database and state (flows, credentials, settings)
+# Without this volume, flow edits will be lost on container restart
+# Default: ./langflow-data
+LANGFLOW_DATA_PATH=./langflow-data
```
edwinjosechittilappilly, line 80: I agree this is a good change.
- We need to test/ check the following:
- What is the data thats getting saved in this Folder.
- If we move to use ext DB like Postgre will this be an issue.

Dockerfile.langflow
```diff
@@ -1,7 +1,20 @@
...
+RUN pip install uv \
+    && mkdir -p /app/langflow-data
+
+COPY --chmod=755 docker-entrypoint-langflow.sh /docker-entrypoint-langflow.sh
 
 EXPOSE 7860
 
-CMD ["langflow", "run", "--host", "0.0.0.0", "--port", "7860"]
\ No newline at end of file
+ENTRYPOINT ["/docker-entrypoint-langflow.sh"]
```
edwinjosechittilappilly: @zzzming  Is there any effects of this on SaaS

docker-compose.yml
```diff
@@ -145,7 +146,8 @@ services:
       - WATSONX_URL=${WATSONX_URL:-${WATSONX_ENDPOINT}}
       - WATSONX_PROJECT_ID=${WATSONX_PROJECT_ID}
       - OLLAMA_BASE_URL=${OLLAMA_ENDPOINT}
-      - LANGFLOW_LOAD_FLOWS_PATH=/app/flows
+      - LANGFLOW_CONFIG_DIR=/app/langflow-data
+      - LANGFLOW_DATABASE_URL=sqlite:////app/langflow-data/langflow.db
```
edwinjosechittilappilly: We cannot hard code this!
Potentiall in future we might have deployemnts that use postgre.
Can we load this as an. env so it could be configurable.

.env.example
mpawlow: @edwinjosechittilappilly 

> What is the data thats getting saved in this Folder.

image

- For a full recursive listing, see: [langflow-data_recursive-listing.txt](https://github.com/user-attachments/files/26125226/langflow-data_recursive-listing.txt)

mpawlow: > If we move to use ext DB like Postgre will this be an issue.
- I need to review / consult with @rodrigosnader and / or @zzzming

Dockerfile.langflow
mpawlow: > ENTRYPOINT ["/docker-entrypoint-langflow.sh"]

- I'm really hoping to revert this change
- Motivation: It is an work-around / kludge used to work-around file permissions / access issues by Langflow to it's own database on the new volume / mount (when running in the GitHub actions workflow / CI integration tests
   - Langflow USER running as UID 1000
   - The work-around did not resolve the integration tests failure

docker-compose.yml
mpawlow: Acknowledged. Looking into this  👍

mpawlow: ✅  Fixed `LANGFLOW_DATABASE_URL`

```
LANGFLOW_DATABASE_URL=${LANGFLOW_DATABASE_URL:-sqlite:////app/langflow-data/langflow.db}
```

Dockerfile.langflow
mpawlow: ✅  I reverted the change

.env.example
mpawlow: Reviewed with @rodageve for OpenRAG SaaS implications
- See feedback: https://github.com/langflow-ai/openrag/pull/1167#pullrequestreview-3982595501

Issue #1138, opened 2026-03-13
[Bug]: Uploaded document is not retrievable/searchable using its unique content.


sreedevi-m-2026:
### OpenRAG Version

0.3.0

### Deployment Method

uv add (installed in project)

### Operating System

macOs 15.6

### Python Version

python 3.13

### Affected Area

Retrieval (search, OpenSearch, hybrid search)

### Bug Description

When searching for a unique token in the Project Knowledge search bar, the results include unrelated documents that do not contain the searched text.

This indicates the search is not filtering documents strictly based on content or metadata.

### Steps to Reproduce

1. Upload multiple documents to the knowledge base.
One document contains a unique token:
OPENSEARCH-7419-ZX
2. Navigate to:
Knowledge → Project Knowledge
Enter the unique token in the search bar:

OPENSEARCH-7419-ZX

### Expected Behavior

Only documents containing the searched token should appear.

Example:

OpenRAG Index Test Document.txt

### Actual Behavior

The search results display multiple unrelated documents such as:

rag_intro.pdf
ai_concepts.xlsx
rag_notes.csv
nlp.html
ai_overview.md
docling_info.adoc

These files do not contain the searched token.

### Relevant Logs

```shell

```

### Screenshots

<img width="1639" height="968" alt="Image" src="https://github.com/user-attachments/assets/f7efa8b5-b714-4dd8-83b4-b0648967bfe2" />

[OpenRAG Index Test Document.txt](https://github.com/user-attachments/files/25962873/OpenRAG.Index.Test.Document.txt)

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

PR #1181, opened 2026-03-18
fix: Uploaded document is not retrievable/searchable using its unique content.

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1138

Improved Knowledge search relevance and partial-query handling by tuning frontend threshold defaults and backend hybrid query behavior.
Before:
Default scoreThreshold was effectively 0 when not provided, so low-relevance files were often included.
Short partial queries (like vita) often returned nothing.
Unique terms could still return unrelated files due to semantic spillover.
Returned total could reflect raw backend hits instead of post-filtered displayed results.
https://github.com/user-attachments/assets/47b4ff45-62a4-473d-b440-e1b82720be9e

Now:
Default scoreThreshold is 1.25 (frontend), with dynamic relaxation to 1.0 for short single-token queries (<=4 chars).
Backend keyword branch is stricter (operator: and, fuzziness: AUTO:4,7) and includes bool_prefix support for partials.
Added exact-text file preference: if query appears verbatim in specific files, results are narrowed to those files.
Fixed exact-file filtering robustness (only valid string filenames considered).
total now matches the actual filtered results shown (len(chunks)).
https://github.com/user-attachments/assets/7d073caf-c00d-41ee-aaa7-6a3735d86a91

### DIFF
--- a/flows/openrag_agent.json
+++ b/flows/openrag_agent.json
@@ -5152,7 +5152,7 @@
                 ],
                 "list": false,
                 "list_add_label": "Add More",
-                "load_from_db": false,
+                "load_from_db": true,
                 "multiline": false,
                 "name": "input_value",
                 "override_skip": false,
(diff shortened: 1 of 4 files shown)


Wallgau:
create branch with same name, probably forgot the -b, will reopen a PR with only my changes (3files)

PR #1183, opened 2026-03-18
fix: Uploaded document is not retrievable/searchable using its unique content. 

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1138

Improved Knowledge search relevance and partial-query handling by tuning frontend threshold defaults and backend hybrid query behavior.
Before:
Default scoreThreshold was effectively 0 when not provided, so low-relevance files were often included.
Short partial queries (like vita) often returned nothing.
Unique terms could still return unrelated files due to semantic spillover.
Returned total could reflect raw backend hits instead of post-filtered displayed results.
https://github.com/user-attachments/assets/47b4ff45-62a4-473d-b440-e1b82720be9e

Now:
Default scoreThreshold is 1.25 (frontend), with dynamic relaxation to 1.0 for short single-token queries (<=4 chars).
Backend keyword branch is stricter (operator: and, fuzziness: AUTO:4,7) and includes bool_prefix support for partials.
Added exact-text file preference: if query appears verbatim in specific files, results are narrowed to those files.
Fixed exact-file filtering robustness (only valid string filenames considered).
total now matches the actual filtered results shown (len(chunks)).
https://github.com/user-attachments/assets/7d073caf-c00d-41ee-aaa7-6a3735d86a91

### DIFF
--- a/src/services/search_service.py
+++ b/src/services/search_service.py
@@ -1,5 +1,6 @@
 import copy
 import json
+from collections import Counter
 from typing import Any, Dict
 from agentd.tool_decorator import tool
 from config.settings import EMBED_MODEL, clients, get_embedding_model, get_index_name, WATSONX_EMBEDDING_DIMENSIONS
@@ -317,19 +318,23 @@ class SearchService:
                                 "query": query,
                                 "fields": ["text^2", "filename^1.5"],
                                 "type": "best_fields",
-                                "operator": "and",
+                                "operator": "or",
                                 "fuzziness": "AUTO:4,7",
                                 "boost": 0.3,  # 30% weight for keyword search
                             }
                         },
                         {
-                            # Help prefix/partial inputs (e.g. "vita" -> "vitamin")
-                            # without making the whole query fully fuzzy.
-                            "multi_match": {
-                                "query": query,
-                                "fields": ["text^1.5", "filename^2"],
-                                "type": "bool_prefix",
-                                "boost": 0.25,
+                            # Prefix fallback for partial input (e.g. "vita" -> "vitamin").
+                            # Avoid bool_prefix here because our current mappings are:
+                            # - text: standard "text" (not search_as_you_type / edge-ngram)
+                            # - filename: "keyword"
+                            # match_phrase_prefix with a bounded expansion is safer.
+                            "match_phrase_prefix": {
+                                "text": {
+                                    "query": query,
+                                    "max_expansions": 50,
+                                    "boost": 0.25,
+                                }
                             }
                         },
                     ],
@@ -495,6 +500,7 @@ class SearchService:
         # If query text appears verbatim in one subset of files, prefer those files
         # to avoid broad semantic spillover for unique lookups.
         normalized_query = query.strip().lower()
+        aggregations = results.get("aggregations", {})
         if (
             normalized_query
             and not is_wildcard_match_all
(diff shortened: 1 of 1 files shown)


src/services/search_service.py
```diff
@@ -317,10 +317,21 @@ async def embed_with_model(model_name):
                                 "query": query,
                                 "fields": ["text^2", "filename^1.5"],
                                 "type": "best_fields",
-                                "fuzziness": "AUTO",
```
ricofurtado, line 320: # Current
  "operator": "and",                                                                                                                                                                                                                             
  "fuzziness": "AUTO:4,7",                                                   
                                                                                                                                                                                                                                                 
  operator: "and" means all query tokens must match. Combined with fuzziness, each token can match fuzzily, but every token must appear. For multi-word queries (e.g. "vitamin c supplement"), all three terms must have a fuzzy match — this    
  aggressively reduces recall. The PR description says this is for strictness, but it's at odds with the prefix branch added on line 325-334 which is permissive (OR by default).
                                                                                                                                                                                                                                                 
  Suggested fix — either document the intentional strictness explicitly or use operator: "or" for the keyword branch and rely on score threshold to filter low-quality results:                                                                  
   
- Option A — keep strict but document it                                                                                                                                                                                                       
  "operator": "and",      # all terms must match; prefer precision over recall                                                                                                                                                                   
  "fuzziness": "AUTO:4,7",                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                 
- Option B — align with prefix branch permissiveness                                                                                                                                                                                           
  "operator": "or",                                                          
  "fuzziness": "AUTO:4,7",

src/services/search_service.py
```diff
@@ -317,10 +317,21 @@ async def embed_with_model(model_name):
...
+                                "operator": "and",
+                                "fuzziness": "AUTO:4,7",
                                 "boost": 0.3,  # 30% weight for keyword search
                             }
                         },
+                        {
+                            # Help prefix/partial inputs (e.g. "vita" -> "vitamin")
+                            # without making the whole query fully fuzzy.
+                            "multi_match": {
```
ricofurtado: bool_prefix may not work without index mapping (search_service.py:328-333) — Significant
                                                                    
multi_match with type: bool_prefix works well only when the field is mapped as search_as_you_type or uses an edge-ngram analyzer. On standard text fields, OpenSearch rewrites it as a prefix query on the last token only — which can be very 
  slow (linear scan) and may not produce the intended behaviour on large indexes.
                                                                                                                                                                                                                                                 
Verify the index mapping, or add a comment confirming the field is search_as_you_type:                                                                                                                                         
   
  {                                                                                                                                                                                                                                              
      # Requires 'text' and 'filename' to be mapped as search_as_you_type    
      # or edge-ngram tokenized for efficient prefix matching.                                                                                                                                                                                   
      "multi_match": {                                                                                                                                                                                                                           
          "query": query,
          "fields": ["text^1.5", "filename^2"],                                                                                                                                                                                                  
          "type": "bool_prefix",                                                                                                                                                                                                                 
          "boost": 0.25,
      }                                                                                                                                                                                                                                          
  }

src/services/search_service.py
```diff
@@ -481,15 +492,35 @@ async def embed_with_model(model_name):
...
+                if isinstance(filename, str)
+                and (
+                    normalized_query in filename.lower()
+                    or (
+                        isinstance(chunk.get("text"), str)
+                        and normalized_query in chunk.get("text", "").lower()
+                    )
+                )
+            }
+            if exact_files:
```
ricofurtado, line 522: Aggregations are stale after exact_files filter
                                                                                                                                                                                                                                                 
  # Current (line 516-524)
  if exact_files:                                                                                                                                                                                                                                
      chunks = [chunk for chunk in chunks if chunk.get("filename") in exact_files]
                                                                                                                                                                                                                                                 
  return {                                                                                                                                                                                                                                       
      "results": chunks,
      "aggregations": results.get("aggregations", {}),  # line 522 — stale!                                                                                                                                                                      
      "total": len(chunks),                                                  
  }

aggregations (data_sources, document_types, owners, etc.) are computed by OpenSearch over the full result set. After exact_files filters chunks, the aggregations still reflect removed filenames. For example, the data_sources bucket may  show files that are no longer in results. This is a UI consistency bug.
                                                                                                                                                                                                                                                 
Suggested fix: recompute aggregations from the filtered chunks, or at least filter data_sources:                                                                                                                                              
   
  if exact_files:                                                                                                                                                                                                                                
```
      chunks = [chunk for chunk in chunks if chunk.get("filename") in exact_files]                                                                                                                                                               
      # Rebuild data_sources aggregation to match filtered results                                                                                                                                                                               
      aggs = results.get("aggregations", {})                                                                                                                                                                                                     
      if "data_sources" in aggs:                                                                                                                                                                                                                 
          aggs = {                                                                                                                                                                                                                               
              **aggs,                                                                                                                                                                                                                            
              "data_sources": {                                                                                                                                                                                                                  
                  "buckets": [                                                                                                                                                                                                                   
                      b for b in aggs["data_sources"].get("buckets", [])                                                                                                                                                                         
                      if b.get("key") in exact_files                                                                                                                                                                                             
                  ]                                                                                                                                                                                                                              
              },                                                                                                                                                                                                                                 
          }                                                                                                                                                                                                                                      
  else:                                                                      
      aggs = results.get("aggregations", {})

  return {
      "results": chunks,
      "aggregations": aggs,
      "total": len(chunks),
  }        
```

Issue #1174, opened 2026-03-18
[Bug]: Overwrite functionality unavailable for .txt files


Vaisakh-r558:
### OpenRAG Version

0.3.2

### Deployment Method

uv add (installed in project)

### Operating System

macOS Tahoe 26.3

### Python Version

3.13

### Affected Area

Ingestion (document processing, upload, Docling)

### Bug Description

Usually for other file extensions (such as pdf,csv,png) when we try to upload a file that is already existing in the index a message pops up saying "Overwriting will replace the existing document with another version. This can't be undone." along with choice to user to either overwrite or cancel the operation. In case of .txt files this overwrite warning is not displayed, instead the file content gets stored as an extra chunk in the existing .txt file.

### Steps to Reproduce

1. Go to knowledge section.
2. Upload the .txt file sample.txt
3. Wait for ingestion completion and view chunks.
4. Upload the same document again.

### Expected Behavior

An overwrite warning must come up if a duplicate file is detected.

### Actual Behavior

No overwrite warning at all, the existing doc in the index just gets updated to contain one extra chunk with same content.

### Relevant Logs

```shell

```

### Screenshots

<img width="1728" height="963" alt="Image" src="https://github.com/user-attachments/assets/ee85b8d2-7315-4359-8fc0-8c739792d54f" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

PR #1190, opened 2026-03-18
fix: .txt and .md filenames should treated as duplicates.

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1174
this issue happened because .txt uploads were internally renamed and indexed as .md, the duplicate check looked for filename.txt while existing chunks were stored under filename.md, so it missed the match and appended chunks instead of triggering overwrite.

Before:
https://github.com/user-attachments/assets/4a9168fd-fac5-44b5-b92b-d84bbb1412e8

Now:
https://github.com/user-attachments/assets/5e4e5009-2990-4518-ae48-55efb9766dc4

### DIFF
--- a/src/api/documents.py
+++ b/src/api/documents.py
@@ -123,12 +123,14 @@ async def check_filename_exists(
         from utils.opensearch_queries import build_filename_search_body
         from utils.file_utils import get_filename_aliases
 
-        candidate_filenames = get_filename_aliases(filename) or [filename]
+        candidate_filenames = get_filename_aliases(filename)
+        if not candidate_filenames:
+            return JSONResponse({"exists": False, "filename": filename}, status_code=200)
 
         logger.debug("Checking filename existence", filename=filename, index_name=get_index_name())
+        exists = False
 
         try:
-            exists = False
             for candidate in candidate_filenames:
                 search_body = build_filename_search_body(candidate, size=1, source=["filename"])
                 response = await opensearch_client.search(
--- a/src/models/processors.py
+++ b/src/models/processors.py
@@ -76,11 +76,22 @@ class TaskProcessor:
         max_retries = 3
         retry_delay = 1.0
 
-        candidate_filenames = get_filename_aliases(filename) or [filename]
+        candidate_filenames = get_filename_aliases(filename)
+        if not candidate_filenames:
+            return False
+        # Keep track of aliases that still need checking across retries.
+        # If one alias was already checked successfully with no hits, we avoid
+        # re-querying it when another alias fails transiently.
+        pending_candidates = list(candidate_filenames)
+        # Retry strategy: only retry aliases that have not completed successfully.
+        # This avoids re-querying aliases already checked with no hits when a later
+        # alias fails transiently (e.g., timeout).
 
         for attempt in range(max_retries):
             try:
-                for candidate in candidate_filenames:
+                i = 0
+                while i < len(pending_candidates):
+                    candidate = pending_candidates[i]
                     search_body = build_filename_search_body(
                         candidate, size=1, source=False
                     )
@@ -91,6 +102,10 @@ class TaskProcessor:
                     hits = response.get("hits", {}).get("hits", [])
                     if hits:
                         return True
+                    # Successfully checked this alias with no hits; don't
+                    # re-query it on future retries.
+                    pending_candidates.pop(i)
+                    continue
                 return False
 
             except (asyncio.TimeoutError, Exception) as e:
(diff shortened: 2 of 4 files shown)


src/utils/file_utils.py
```diff
@@ -98,4 +98,26 @@ def clean_connector_filename(filename: str, mimetype: str) -> str:
         return clean_name
     if not clean_name.lower().endswith(suffix.lower()):
         return clean_name + suffix
-    return clean_name
\ No newline at end of file
+    return clean_name
+
+
+def get_filename_aliases(filename: str) -> list[str]:
```
ricofurtado, line 104: I would re-write it to make more explicity:
      """Return equivalent filename variants used by ingestion/indexing.                                                                                                                                                                         
                                                                                                                                                                                                                                                 
      Legacy Langflow ingest indexes `.txt` uploads as `.md` (see LangflowFileProcessor).                                                                                                                                                        
      The alias always uses a lowercase extension to match how the rename is applied:                                                                                                                                                            
        original_filename[:-4] + '.md'                                                                                                                                                                                                           
      So 'FOO.TXT' aliases to 'FOO.md', not 'FOO.MD'.                                                                                                                                                                                            
      """

ricofurtado, line 104: Empty filename fallback inconsistency (**processors.py:79, processors.py:134, documents.py:126**)                                                                                                                                        
                                                                                                                                                                                                                                                 
  All three callers use the same pattern:                                                                                                                                                                                                        
  candidate_filenames = get_filename_aliases(filename) or [filename]                                                                                                                                                                             
                                                                             
  When filename="", get_filename_aliases("") returns [], so the or kicks in and produces [""] — a search/delete for an empty-string filename. The function's [] return for empty input was meant as an early-exit guard, but the callers undo it.
                                                                                                                                                                                                                                                 
  Suggested fix — remove the early-exit in file_utils.py:110-112 and instead handle it at the function level by always including the original:                                                                                                   
                                                                                                                                                                                                                                                 
```
  def get_filename_aliases(filename: str) -> list[str]:                                                                                                                                                                                          
      normalized = (filename or "").strip()                                  
      if not normalized:                                                                                                                                                                                                                         
          return [filename] if filename else []  # caller gets original back, not []
                                                                                                                                                                                                                                                 
      aliases = [normalized]                                                 
      ...                 
```                                                                                                                                                                                                                       
                                                                             
  Or, simpler — document that callers should never pass an empty filename and add an assert in dev mode.

src/api/documents.py
```diff
@@ -121,26 +121,31 @@ async def check_filename_exists(
...
+        candidate_filenames = get_filename_aliases(filename) or [filename]
 
         logger.debug("Checking filename existence", filename=filename, index_name=get_index_name())
 
         try:
-            response = await opensearch_client.search(
-                index=get_index_name(),
-                body=search_body
-            )
+            exists = False
```
ricofurtado: Code clarity: a reader must verify the raise on line 147 ensures exists is always assigned before line 149. Initializing before the try makes the intent explicit.                                                        
                                                                             
  Suggested fix (documents.py:130):                                                                                                                                                                                                              
  exists = False   # move here, before the try                               
  try:                                                                                                                                                                                                                                           
      for candidate in candidate_filenames:                                                                                                                                                                                                      
          ...

src/models/processors.py
```diff
@@ -72,19 +76,22 @@ async def check_filename_exists(
...
-
-                response = await opensearch_client.search(
-                    index=get_index_name(),
-                    body=search_body
-                )
-
-                # Check if any hits were found
-                hits = response.get("hits", {}).get("hits", [])
-                return len(hits) > 0
+                for candidate in candidate_filenames:
```
ricofurtado: Double OpenSearch queries for every .txt/.md file (processors.py:83-93)
                                                                                                                                                                                                                                                 
  The retry loop re-queries all candidates from scratch on each attempt. If candidate 1 succeeds but candidate 2 times out, the next attempt re-queries candidate 1 unnecessarily.
                                                                                                                                                                                                                                                 
I would do:                               
```
  for attempt in range(max_retries):                                                                                                                                                                                                             
      try:                                                                   
          for candidate in candidate_filenames:
              ...                              
              if hits:
                  return True
          return False       
```                                                                                                                                                                                                                    
                      
This is already how it's written, which is fine. The real note: for high-volume deployments, .txt uploads now always cost 2 OpenSearch calls even when the first hit returns True and breaks early. Acceptable, but worth a comment:           
  - Check both .txt and .md variants — LangflowFileProcessor indexes .txt as .md (see processors.py:777)                                                                                                                                         
  for candidate in candidate_filenames:

src/utils/file_utils.py
Wallgau: Good catch. I addressed the empty filename issue, but at the call sites rather than by changing get filename aliases return semantics.

get filename aliases still returns [] for empty or whitespace input (guard remains).
The problematic behavior came from callers doing get filename aliases(filename) or [filename], which could turn empty input into [""].

I removed that fallback in the affected paths and now handle empty alias lists explicitly:

TaskProcessor.check filename exists() returns False when aliases are empty.
TaskProcessor.delete document by filename() no ops (with log) when aliases are empty.
api/documents.check filename exists() returns { "exists": false } early when aliases are empty.
This avoids empty string OpenSearch queries while keeping get filename aliases() focused on normalized alias generation.

PR #1192, opened 2026-03-19
docs: Update supported file types, provider management, connector management

### LINKED ISSUE #1128: [Bug]: RTF file ingestion failed
### OpenRAG Version

0.3.0

### Deployment Method

uv add (installed in project)

### Operating System

macOs 15.6

### Python Version

python 3.13

### Affected Area

### Bug Description

RTF File Ingestion Fails in Knowledge Upload
Uploading an .rtf file to the knowledge base fails during ingestion. The file is marked failed and the pipeline throws an error in the Docling export step.

### Steps to Reproduce

1. Open OpenRAG.
2. Go to Knowledge.
3. Click Add Knowledge.
4. Upload a valid .rtf file.
5. Start ingestion.
6. Check the ingestion result / failure details.

### Expected Behavior

The .rtf file should ingest successfully because .rtf is listed as a supported file type. https://docs.openr.ag/ingestion/#supported-file-types

### Actual Behavior

The .rtf file ingestion fails with an error similar to:

Error running graph: Error building Component Export DoclingDocument:
Error running method "as_dataframe":
Column 'doc' not found in DataFrame. Available columns: []

### Relevant Logs

```shell
Error log :
Error running graph: Error building Component Export DoclingDocument: Error running method "as_dataframe": Column 'doc' not found in DataFrame. Available columns: []. Possible solutions: 1. Use the 'Data' output from Docling component instead of 'DataFrame' output 2. Update the 'Doc Key' parameter to match one of the available columns 3. If using VLM pipeline, try using the standard pipeline
```

### Screenshots

Image

Image

attached the uploaded document also
[openrag_readme.rtf](https://github.com/user-attachments/files/25960678/openrag_readme.rtf)

### Additional Context

_No response_

### Checklist

aimurphy:
* Update supported file types
* Add cloud connector sync
* Update cloud connector management
* Inspect failed tasks
* Expand model provider management and add more warnings about changing and removing models/providers
* Deleted the openrag-documentation.pdf since we now ingest from docs.openr.ag

Closes #1128


docs/docs/core-components/ingestion-configure.mdx
```diff
@@ -59,14 +59,13 @@ Because the OpenRAG UI validates model availability and compatibility, setting m
...
 
-OpenRAG _doesn't_ reprocess existing documents when you change the embedding model.
+OpenRAG _doesn't_ reprocess existing documents when you change, add, or remove embedding models.
+
 If you want to generate new embeddings for existing documents, you must reupload the documents after enabling the new embedding model.
 To remove previously generated embeddings, you must delete the existing document from the knowledge base.
 
 If you use different embedding models for different documents, you can create [filters](/knowledge-filters) to separate documents that were embedded with different models.
 
-If you use multiple embeddings models, be aware that similarity search (in **Chat**) can take longer as the agent searches each model's embeddings separately.
```
aimurphy, line 68: I don't think this is true. Please correct me if otherwise.

docs/docs/core-components/knowledge-browse.mdx
```diff
@@ -66,6 +66,16 @@ If you [delete these documents](#delete-knowledge), then you won't be able to as
 It is recommended that you keep these documents, and use [filters](/knowledge-filters) to separate them from your other knowledge.
 An **OpenRAG Docs** filter is created automatically for these documents.
 
+### Get latest OpenRAG documentation
+
+On the **Knowledge** page, click <Icon name="Library" aria-hidden="true"/> **Fetch latest docs** to ingest the latest OpenRAG documentation from docs.openrag.com (this website).
```
aimurphy:
```suggestion
On the **Knowledge** page, click <Icon name="Library" aria-hidden="true"/> **Fetch latest docs** to ingest the latest OpenRAG documentation from docs.openr.ag (this website).
```

openrag-documents/openrag-documentation.pdf
aimurphy, line 1: Is it ok to delete this?

aimurphy: @Wallgau

docs/docs/core-components/knowledge-browse.mdx
```diff
@@ -66,6 +66,16 @@ If you [delete these documents](#delete-knowledge), then you won't be able to as
...
 An **OpenRAG Docs** filter is created automatically for these documents.
 
+### Get latest OpenRAG documentation
+
+On the **Knowledge** page, click <Icon name="Library" aria-hidden="true"/> **Fetch latest docs** to ingest the latest OpenRAG documentation from docs.openr.ag (this website).
+
+:::tip
+**Fetch latest docs** _only_ gets the latest OpenRAG documentation.
+
+To update docs ingested from cloud storage connectors], see [Configure connectors](/knowledge-connectors).
```
mendonk:
```suggestion
To update docs ingested from cloud storage connectors, see [Configure connectors](/knowledge-connectors).
```

mendonk (approved):
One typo, approved

docs/docs/core-components/ingestion-configure.mdx
Wallgau: I think that statement is true.
src/services/search_service.py detects all indexed embedding_model values, generates a query embedding for each, then builds multiple KNN clauses (one per model field) and runs them together via dis_max. So with more embedding models, retrieval work increases (more embedding calls + larger query), which can make Chat retrieval slower, depending on provider latency and OpenSearch load.

aimurphy: Ok thanks! I will add it back.

Caveat: If you remove a provider, then the Chat cannot generate an embedding with the appropriate embedding model. What happens in that case? Will the Chat ignore documents embedded by the missing providing? Or does it search with a "fallback" embedding model (potentially incorrect/missing results because the embedding dimensions/structure don't match)?

Wallgau: from my understanding but always good to double check with @edwinjosechittilappilly.
If a provider is removed, OpenRAG still discovers embedding models from indexed documents and attempts to generate a query embedding for each, so if one of those models is no longer available, retrieval may fail and Chat may return an error or no useful answer; there is no safe cross-model fallback for those documents because embedding spaces and dimensions differ, and the system does not consistently fallback to searching only documents tied to still-available models.

openrag-documents/openrag-documentation.pdf
aimurphy: confirmed via slack

Issue #1091, opened 2026-03-10
[Bug]: Performance: Adding, removing, updating model providers and changing language or embedding models is slow after upgrade to Langflow 1.8.0


mpawlow:
### OpenRAG Version

0.2.5 + latest from main (commit c006d45ae582c038590e6b049b4d6b16597a353f)

### Deployment Method

uvx (uvx openrag)

### Operating System

Ubuntu 24.04.4 LTS

### Python Version

Python 3.13.12

### Affected Area

Settings (configuration, model providers)

### Bug Description

- `Problem`: Adding, removing, updating model providers and changing language or embedding models is slow after upgrade to Langflow 1.8.0
   - e.g. Encountering severe performance issues
   - See attached video below

### Steps to Reproduce

1. Go to Settings UI
2. Add a new model provider
   - BUG: Slow performance to persist the change
3. Change the language model
   - BUG: Very slow performance to persist the change
   - In my case, ~23 seconds
4. Change the embedding model
   - BUG: Very slow performance to persist the change
5. Update the new model provider (with a new API key, etc.)   
   - BUG: Slow performance to persist the change4. 
6. Remove the new model provider
   - BUG: Slow performance to persist the change

### Expected Behavior

No performance lag to persist the changes

### Actual Behavior

Performance lag to persist the changes

### Relevant Logs

```shell
N/A
```

### Screenshots

https://github.com/user-attachments/assets/696b1fd2-e872-4298-9ba2-c46d3f59352d

### Additional Context

- Note: I rebuilt the **Langflow** image via ` make dev-local-build-lf-cpu` after doing a factory reset

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

PR #1193, opened 2026-03-19
fix: Performance: Adding, removing, updating model providers and changing language or embedding models is slow after upgrade to Langflow 1.8.0 

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1091
Keep settings persistence responsive by saving config and returning immediately, while running heavy Langflow global-variable/model propagation asynchronously with safe error logging.

https://github.com/user-attachments/assets/2ec489cd-fc83-49a5-9fa0-edaabc117577

### DIFF
--- a/src/api/settings.py
+++ b/src/api/settings.py
@@ -37,6 +37,7 @@ from dependencies import (
 from session_manager import User
 
 logger = get_logger(__name__)
+_background_tasks: set[asyncio.Task] = set()
 
 
 class SettingsUpdateBody(BaseModel):
@@ -821,12 +822,13 @@ async def update_settings(
 
         # Run expensive Langflow sync in the background to keep settings updates responsive.
         if should_validate or provider_updated:
-            asyncio.create_task(
+            task = asyncio.create_task(
                 _run_async_post_save_langflow_updates(
                     session_manager=session_manager,
                     update_mcp_servers=(
                         body.embedding_provider is not None
                         or body.embedding_model is not None
+                        or provider_updated
                     ),
                     update_model_values=(
                         body.llm_provider is not None
@@ -837,6 +839,9 @@ async def update_settings(
                     ),
                 )
             )
+            # Keep a strong reference until completion to avoid premature GC cancellation.
+            _background_tasks.add(task)
+            task.add_done_callback(_background_tasks.discard)
 
 
         set_fields = [k for k, v in body.model_dump().items() if v is not None]
--- a/tests/unit/test_settings_async_post_save.py
+++ b/tests/unit/test_settings_async_post_save.py
@@ -1,0 +1,136 @@
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
+import pytest
+
+import api.settings as settings_api
+
+
+class _FakeTask:
+    def __init__(self):
+        self.done_callback = None
+
+    def add_done_callback(self, cb):
+        self.done_callback = cb
+
+
+def _make_config_for_ollama_removal():
+    providers = SimpleNamespace(
+        openai=SimpleNamespace(api_key="openai-key", configured=True),
+        anthropic=SimpleNamespace(api_key="", configured=False),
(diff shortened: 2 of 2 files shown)


src/api/settings.py
```diff
@@ -815,30 +817,31 @@ async def update_settings(
...
-
-                # Update model values if provider or model changed (including removals that trigger fallback)
-                if body.llm_provider is not None or body.llm_model is not None or body.embedding_provider is not None or body.embedding_model is not None or provider_updated:
-                    await _update_langflow_model_values(current_config, flows_service)
-
-            except Exception as e:
-                logger.error(f"Failed to update Langflow settings: {str(e)}")
-                # Don't fail the entire settings update if Langflow update fails
-                # The config was still saved
+            task = asyncio.create_task(
```
mpawlow, line 825: ### (a) [Major] asyncio.create_task() result is discarded — background sync task is vulnerable to silent GC cancellation

- `Problem`: asyncio.create_task() is called to schedule the background Langflow sync, but the returned Task object is never assigned to a variable or stored in a collection. 
   - Python's asyncio documentation explicitly warns that a task not held by a strong reference is eligible for garbage collection before it completes — at which point it is silently cancelled with no error raised and no log emitted. 
   - The background task performs significant I/O (updating global variables, MCP server credentials, and model values in Langflow), which can take several seconds. 
   - Under any GC pressure after the HTTP response is sent, the task is at risk of being cancelled mid-execution, causing Langflow's state to silently diverge from the saved config. The codebase already uses the correct safe pattern in TaskService (store the task in a set + done_callback), but the new code in settings.py does not follow it.
- `Proposed Solution`: Follow the existing pattern already established in TaskService. Store the task in a module-level set and add a done_callback to discard it upon completion:

```python
_background_tasks: set = set()

task = asyncio.create_task(_run_async_post_save_langflow_updates(...))
_background_tasks.add(task)
task.add_done_callback(_background_tasks.discard)
```

src/api/settings.py
```diff
@@ -1380,6 +1383,37 @@ async def _update_langflow_global_variables(config, flows_service=None):
...
+    update_mcp_servers: bool,
+    update_model_values: bool,
+) -> None:
+    """Apply post-save Langflow synchronization asynchronously."""
+    try:
+        current_config = get_openrag_config()
+        flows_service = _get_flows_service()
+
+        # Update global variables
+        await _update_langflow_global_variables(
```
mpawlow, line 1397: ### (b) [Major] update_mcp_servers flag does not account for provider removals that implicitly change the active embedding provider

- `Problem`: The update_mcp_servers boolean passed to _run_async_post_save_langflow_updates is computed solely from body.embedding_provider is not None or body.embedding_model is not None.
   - However, when a provider is removed (e.g. remove_ollama_config when Ollama is the currently active
embedding provider), the system internally invokes _first_configured_embedding_provider() to select a fallback, sets provider_updated = True, and clears current_config.knowledge.embedding_model — but body.embedding_provider and body.embedding_model both remain None since the user
did not explicitly set them. 
   - As a result, update_mcp_servers evaluates to False and _update_mcp_servers_with_provider_credentials is never called, leaving MCP servers pointing at stale credentials for the now-removed provider. 
   - The parallel update_model_values flag already handles this case correctly by including provider_updated in its condition (lines 831–837); update_mcp_servers needs the same treatment.
   - Related: src/api/settings.py, lines 827–830 — update_mcp_servers condition computation
- `Proposed Solution`
   - Mirror the update_model_values condition by including provider_updated:

```python
update_mcp_servers=(
  body.embedding_provider is not None
  or body.embedding_model is not None
  or provider_updated  # covers provider-removal fallback case
),
```

src/api/settings.py
mpawlow: Verified

src/api/settings.py
mpawlow: Verified

Issue #1170, opened 2026-03-18
[Bug]: TransportError(503,'search_phase_execution_exception') while uploading documents


Vaisakh-r558:
### OpenRAG Version

0.3.2

### Deployment Method

uv add (installed in project)

### Operating System

macOS Tahoe 26.3

### Python Version

3.13

### Affected Area

Ingestion (document processing, upload, Docling)

### Bug Description

While uploading file an error popped up and all the existing documents from the index disappeared on reloading the page. The error keeps popping up on refreshing the page or clicking Fetch latest docs.

### Steps to Reproduce

Prerequisite: There must be around 30- 35 docs existing in the index before this error is reproduced. 
Steps:-
1.  Go to knowledge section
2. Ingest the sample.txt file.

### Expected Behavior

File should be ingested successfully with task completion message appearing in bottom right corner. On clicking "Fetch latest docs" or reloading the page it should be present in the list.

### Actual Behavior

Shows error message in bottom right corner as shown in the attached video, on clicking "Fetch latest docs" or reloading the error message pops up in the bottom right corner, also all the existing docs have disappeared from the list.

### Relevant Logs

```shell
Logs for open search service is attached as logs_opensearch.txt.
```

### Screenshots

<img width="1481" height="907" alt="Image" src="https://github.com/user-attachments/assets/b04fd51c-ffdd-4bd2-aa69-1cd8d44bd15d" />

### Additional Context

[sample.txt](https://github.com/user-attachments/files/26079104/sample.txt)

[logs_opensearch.txt](https://github.com/user-attachments/files/26079246/logs_opensearch.txt)

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

Wallgau:
not able to reproduce in release-0.4.0: 
https://github.com/user-attachments/assets/4a240d81-d0e0-463f-bfdc-205b66af3252

Vaisakh-r558:
<img width="1824" height="1101" alt="Image" src="https://github.com/user-attachments/assets/3458cf51-1596-4f18-9b53-9a694f6724f6" />
<img width="1824" height="1101" alt="Image" src="https://github.com/user-attachments/assets/1c057927-f8ee-4ecd-80a5-536da502c614" />
[...]

PR #1207, opened 2026-03-20
fix: self-heal stale url-ingestion provider state

Wallgau:
**This PR fixes URL-ingestion failures caused by stale provider/runtime state in Langflow and MCP wiring.**
- Scopes ingestion and MCP provider credentials to the selected embedding provider instead of sending all provider vars.
- Removes stale provider headers (for example old OLLAMA_BASE_URL) from MCP server args so non-selected providers are not accidentally used.
- Adds conditional self-heal for URL ingestion: on known stale-state signatures (for example “Failed to connect to Ollama” while provider is not Ollama), reconcile once and retry once.
- Improves retry-failure errors with provider-aware messaging so failures are easier to diagnose.

**Issue:** https://github.com/langflow-ai/openrag/issues/1170
- Users could still hit “Failed to connect to Ollama” even when OpenAI was selected, due to stale runtime/MCP state and mixed provider headers. This PR makes provider selection authoritative and adds safe, targeted recovery.

**What Changed**

- src/utils/langflow_headers.py
- Provider-scoped credential/header/global-var generation.
- src/services/langflow_mcp_service.py
- Prunes stale provider headers not in the desired selected-provider set.
- src/services/langflow_file_service.py
- Uses selected provider when building ingestion headers.
- Supports dynamic URL-ingest flow-id resolution when available.
- Adds conditional stale-state reconcile + single retry path.
- Adds provider-specific retry error formatting.

**Tests**
Added TDD coverage with mocked HTTP (CI-safe, no real external calls):
- tests/unit/test_url_ingestion_pipeline_tdd.py
- selected-provider-only ingestion headers
- prefers resolved flow-id over stale configured id
- reconcile-and-retry once on known stale-state errors
- no retry on non-reconcilable errors
- provider-specific error message after retry failure
- integration-style mocked chain for provider selection + self-heal

tests/unit/test_mcp_provider_selection_tdd.py
- provider-scoped MCP global vars
- stale provider header pruning
- server update applies selected-provider vars only

Before user will get error trying to ingest url:
Screenshot 2026-03-20 at 11 03 30 AM

Now urls are ingest properly and reflect in the knowledge table:
Screenshot 2026-03-20 at 11 02 35 AM


src/utils/langflow_headers.py
```diff
@@ -47,29 +56,31 @@ async def build_mcp_global_vars_from_config(config, flows_service=None) -> Dict[
...
-            ollama_endpoint = await flows_service.resolve_ollama_url(config.providers.ollama.endpoint)
-        else:
-            ollama_endpoint = transform_localhost_url(config.providers.ollama.endpoint)
-        global_vars["OLLAMA_BASE_URL"] = ollama_endpoint
+    selected_provider = (
+        getattr(getattr(config, "knowledge", None), "embedding_provider", "") or ""
+    ).lower().strip()
+
+    # Include credentials only for the selected embedding provider.
+    if selected_provider == "openai":
```
edwinjosechittilappilly: we might need to provide all providers.
case 1. When users have ingested different documents with different models.

@Wallgau can you test this sceanrio

Issue #1130, opened 2026-03-13
[Bug]: Knowledge Filter Does Not Restrict Document Retrieval Scope


sreedevi-m-2026:
### OpenRAG Version

0.3.0

### Deployment Method

uv add (installed in project)

### Operating System

macOs 15.6

### Python Version

python 3.13

### Affected Area

Knowledge Filters (partitions, document filtering)

### Bug Description

The Knowledge Filter is expected to restrict retrieval results to the selected documents or search criteria. However, even after applying the filter, the system retrieves information from unrelated documents.

### Steps to Reproduce

1. Open Knowledge section in OpenRAG.
2. Upload multiple documents, for example:
Project code for QA validation.docx
rag_intro.pdf
3. Create a Knowledge Filter.
4. Configure the filter:
Filter Name: testfilter
Search Query: ZX-4819-RAG QA project
Select a specific document source.
5. Apply the filter.
6. Go to the Chat interface.
7. Ask a question related to another document.
Example question:

What does RAG stand for?

### Expected Behavior

The system should retrieve information only from the documents selected in the filter.

Example:
If the filter targets Project code for QA validation.docx, responses should only use that document.

### Actual Behavior

The system retrieves information from other documents such as:

rag_intro.pdf

even though the filter was applied.

### Relevant Logs

```shell

```

### Screenshots

<img width="468" height="64" alt="Image" src="https://github.com/user-attachments/assets/05a6c348-29a3-4bc9-ac7b-91a13d7881b1" />

<img width="452" height="284" alt="Image" src="https://github.com/user-attachments/assets/f047dbdd-25ee-45ff-95e9-3754acc55466" />

<img width="452" height="265" alt="Image" src="https://github.com/user-attachments/assets/cf7f6d26-fae9-41d0-9799-f52a36686a53" />

docs attached

[rag_intro.pdf](https://github.com/user-attachments/files/25960866/rag_intro.pdf)

[Project code for QA validation is ZX.docx](https://github.com/user-attachments/files/25960869/Project.code.for.QA.validation.is.ZX.docx)

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

sreedevi-m-2026:
Retested in the new build 0.4.0 dev5 and issue is reproducible again . hence opened the bug @Wallgau 

[...]

PR #1228, opened 2026-03-23
fix: Filter chat scope

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1130

What changed
src/utils/opensearch_filter_merge.py

Centralizes coercion of chat filter JSON into OpenSearch term / terms clauses (explicit filter array + context keys aligned with chat_service, including connector_types → connector_type).
merge_filter_clauses_into_search_body: merges clauses into query via bool (existing behavior), and scopes top-level knn by injecting filters into each vector field’s filter (OpenSearch does not apply query.bool.filter to sibling top-level knn). Avoids adding a filter-only query when the body is kNN-only in a way that left kNN global.
Applies limit / score_threshold when size / min_score are not already set.
flows/components/opensearch_multimodal.py (+ synced flows/*.json)

raw_search runs the request body through apply_chat_filter_expression_to_search_body; stricter JSON validation with clear ValueErrors.
_coerce_filter_clauses delegates to the shared helper.
Dockerfile.langflow

COPY src and PYTHONPATH so from utils.opensearch_filter_merge import … works in the Langflow container.
docker-compose.yml / .env.example / Makefile / .gitignore / docs

Langflow DB/config dir under ./langflow-data, ensure-langflow-data for dev targets, factory-reset / CI cleanup include langflow-data, configuration docs.
tests/unit/test_opensearch_filter_merge.py

Unit coverage for merge/coerce behavior, including top-level knn.
How to verify
uv run pytest tests/unit/test_opensearch_filter_merge.py -v
Run stack with a locally built Langflow image if needed (docker compose build langflow or make dev-local-build-lf), so the Dockerfile PYTHONPATH + src copy match this branch.
In chat with a knowledge filter, confirm raw_search logs show query (after chat filter merge): … and results stay within selected sources.

### DIFF
--- a/src/utils/opensearch_filter_merge.py
+++ b/src/utils/opensearch_filter_merge.py
@@ -76,32 +76,66 @@ def coerce_filter_clauses_from_filter_obj(filter_obj: dict | str | None) -> list
     return context_clauses
 
 
+def _append_to_bool_filter(bool_q: dict[str, Any], filter_clauses: list[dict]) -> None:
+    existing = bool_q.get("filter")
+    if existing is None:
+        bool_q["filter"] = list(filter_clauses)
+    elif isinstance(existing, list):
+        bool_q["filter"] = [*existing, *filter_clauses]
+    else:
+        bool_q["filter"] = [existing, *filter_clauses]
+
+
+def _inject_filter_into_top_level_knn(
+    knn_payload: dict[str, Any], filter_clauses: list[dict]
+) -> dict[str, Any]:
+    """Add OpenSearch kNN ``filter`` to each vector field in a top-level ``knn`` block."""
+    out = copy.deepcopy(knn_payload)
+    scope = {"bool": {"filter": list(filter_clauses)}}
+    for _field, spec in out.items():
+        if not isinstance(spec, dict):
+            continue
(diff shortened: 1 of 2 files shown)


flows/components/opensearch_multimodal.py
```diff
@@ -10,6 +10,11 @@
 from opensearchpy import OpenSearch, helpers
 from opensearchpy.exceptions import OpenSearchException, RequestError
 
+from utils.opensearch_filter_merge import (
```
edwinjosechittilappilly, line 13: This isa Langflwo component code.
Hence importing packages from the OpenRAG folder would not work

Wallgau: Valid concern. It is addressed this by updating Dockerfile.langflow to copy OpenRAG src into the Langflow image and set PYTHONPATH=/app/openrag-src. So the component can import utils.opensearch_filter_merge when running this branch’s image. Without that Dockerfile change (older image), the import would fail.

edwinjosechittilappilly: True but this would bring inconsistency in the apporach and we should device an alternative fix.

@lucaseduoli

PR #1231, opened 2026-03-24
fix: aws connection is failing in OpenRAG version 0 4 0 dev5

ricofurtado:
- Implement `verify_s3_credentials` to validate AWS S3 credentials.
- Update `S3Connector.authenticate` to use the new verification method.
- Add unit tests for `verify_s3_credentials` covering various scenarios.

### **There is a behaviour question(maybe product can anwer it):**

- If the credentials are valid but don’t have s3:ListAllMyBuckets, “Test connection” should still succeed?
In this case, the UI should not block saving just because bucket listing failed.

- If bucket listing is unavailable, do we need to explain why in plain terms? E.g. “Credentials are valid, but this IAM user can’t list all buckets. Enter/select bucket names manually.”

-After saving, the connection should only be treated as usable if it has either:
ListBuckets permission, so auto-discovery can work, or
explicit bucket_names, so the connector can operate without bucket discovery.

### So the practical fix is to split today’s combined step into two states:

- validate credentials
- try to discover buckets
- Bucket discovery should be best-effort, not a prerequisite for saving. If discovery fails with AccessDenied, we should still save the connection and guide the user into a manual bucket-name path instead of surfacing a generic failure.

### DIFF
--- a/src/connectors/aws_s3/api.py
+++ b/src/connectors/aws_s3/api.py
@@ -10,7 +10,7 @@ from dependencies import get_connector_service, get_session_manager, get_current
 from session_manager import User
 from utils.logging_config import get_logger
 
-from .auth import create_s3_resource
+from .auth import create_s3_resource, verify_s3_credentials
 from .models import S3ConfigureBody
 from .support import build_s3_config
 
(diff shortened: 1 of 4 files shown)


edwinjosechittilappilly (changes requested):
Rico can you check with QA team . that S3 and COS would have been missing since we disabled it for OpenSource and works only in IBM cloud deployments.

ricofurtado:
Fixed by another PR, this one have too many conflicts, have to open a new one.

PR #1245, opened 2026-03-24
fix:  Knowledge Filter Does Not Restrict Document Retrieval Scope

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1130 (check there for before behavior)

Now: 
Screenshot 2026-03-24 at 5 02 48 PM

Summary
Applies chat knowledge filters to the Langflow OpenSearch component’s raw_search path so behavior matches search_documents (filter clauses, optional limit / score_threshold, top-level knn scoping). The merge logic lives entirely inside the component (and in exported flow JSON). No imports from OpenRAG src (e.g. no utils.opensearch_filter_merge), so the Langflow image does not need COPY src or PYTHONPATH for this feature.

Motivation
Custom Langflow components run in the Langflow process with imports resolved like normal Python. OpenRAG’s utils package is not part of the Langflow dependency set. Importing it implies putting OpenRAG src on the image and bloating the Langflow Dockerfile. Review feedback: keep helper logic in the component source until a proper upstream Langflow story exists.

What changed
flows/components/opensearch_multimodal.py: Inlined helpers (coerce_filter_clauses_from_filter_obj, merge_filter_clauses_into_search_body, apply_chat_filter_limits_to_body, apply_chat_filter_expression_to_search_body, etc.), raw_search updated to merge filter_expression into the request body with validation errors for bad JSON, _coerce_filter_clauses delegates to the shared module-level helper (includes connector_types → connector_type and empty terms handling).
flows/components/opensearch_filter_merge_standalone.py: Stdlib-only copy of the same logic for unit tests and as the reference to keep in sync with the inline block.
tests/unit/test_opensearch_filter_merge.py: TDD coverage for coerce / merge / limits / end-to-end behavior (37 tests).
flows/ingestion_flow.json, openrag_agent.json, openrag_url_mcp.json, openrag_nudges.json: Embedded component source synced from opensearch_multimodal.py.


flows/components/opensearch_filter_merge_standalone.py
```diff
@@ -1,8 +1,8 @@
 """
-Merge chat/knowledge filter clauses into OpenSearch search request bodies.
+Chat filter → OpenSearch request body merge (stdlib only).
 
-Used by the Langflow OpenSearch ``raw_search`` path so document scope matches
```
edwinjosechittilappilly, line 4: @Wallgau  May I know why do we need this file if the change are already merged to component file ?

tests/unit/test_mcp_provider_selection_tdd.py
```diff
@@ -0,0 +1,150 @@
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
```
edwinjosechittilappilly, line 2: Lets remove these unit tests.
I am not sure if they are required, if possible lets add a new integration test in another follow up PR.

src/services/langflow_file_service.py
```diff
@@ -33,6 +33,58 @@ def _is_transient_request_error(error: Exception) -> bool:
             ),
         )
 
+    @staticmethod
+    def _should_reconcile_url_ingestion_error(error_text: str, selected_provider: str) -> bool:
```
edwinjosechittilappilly, line 37: Could you explain the use of these functions ?

src/services/langflow_mcp_service.py
```diff
@@ -90,6 +90,38 @@ def _upsert_global_var_headers_in_args(self, args: List[str], global_vars: Dict[
...
+        }
+        desired_provider_header_keys = {
+            f"x-langflow-global-var-{k.lower()}"
+            for k in global_vars.keys()
+            if str(k).upper() != "SELECTED_EMBEDDING_MODEL"
+        }
+
+        # Remove stale provider headers not present in the desired set.
+        # Keep non-provider headers (JWT/OWNER/etc.) untouched.
+        pruned_args: List[str] = []
```
edwinjosechittilappilly, line 108: Cant we just check if the keys are present or not and not pass if the values are none or empty ?

flows/components/opensearch_filter_merge_standalone.py
Wallgau: reverting changes from last PR and used for unit test but we can remove it since you mention having those changes in another PR.

src/services/langflow_file_service.py
Wallgau: _reconcile_url_ingestion_runtime_state(selected_provider)

When _should_reconcile_… is true, this runs once before one retry of the same URL ingest request. It tries to reset and realign Langflow with OpenRAG settings:

reset_langflow_flow("url_ingest") (if flows_service supports it) — reload/reset the URL ingest flow so stale graph state is cleared.
change_langflow_model_value(...) — push the current selected_provider, embedding_model, and force_embedding_update=True into the url_ingest flow so globals match the backend config (e.g. stop calling Ollama when the user chose OpenAI).
Both steps are best-effort (exceptions are logged, not fatal); then the code refreshes the flow id and re-POSTs the run. If the retry still fails, _raise_retry_provider_error turns the body into a clearer, provider-oriented error where applicable.

edwinjosechittilappilly:
@Wallgau  May I know if the changes in langflow mcp service and headers requried for fixing the filter changes?

Issue #1139, opened 2026-03-13
[Bug]: Long-Term Stability Issue: Ingestion Fails After Prolonged Application Use with Invalid Docling Serve URL Error


sreedevi-m-2026:
### OpenRAG Version

0.3.0

### Deployment Method

uv add (installed in project)

### Operating System

macOs 15.6

### Python Version

3.13

### Affected Area

Ingestion (document processing, upload, Docling)

### Bug Description

After the application remains running for a prolonged period of continuous use (approximately 4+ hours), document ingestion starts failing. The failure occurs in the Docling Serve component with a URL protocol error, even though ingestion was working earlier in the same session.

### Steps to Reproduce

1. Launch the OpenRAG application.
2. Keep the application running for an extended period, around 4 or more hours.
3. Use the application normally during this period.
4. After prolonged usage, try to ingest a document.
5. Observe the ingestion result.

### Expected Behavior

The application should remain stable and responsive after prolonged continuous use, and document ingestion should continue to work normally.

### Actual Behavior

Document ingestion fails after prolonged usage with the following error:

Error running graph: Error building Component Docling Serve:
Request URL is missing an 'http://' or 'https://' protocol.

### Relevant Logs

```shell
Error running graph: Error building Component Docling Serve:
Request URL is missing an 'http://' or 'https://' protocol.
```

### Screenshots

<img width="452" height="217" alt="Image" src="https://github.com/user-attachments/assets/97213f57-05d4-413b-9508-c88be8f56e61" />

<img width="452" height="283" alt="Image" src="https://github.com/user-attachments/assets/18176580-5870-484d-bbd2-d528431714c1" />

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

PR #1258, opened 2026-03-25
fix: show unsupported-file ingest warning and overwrite for for folder duplicates

Wallgau:
issue: https://github.com/langflow-ai/openrag/issues/1139
Summary
Fix folder upload UX by detecting unsupported files and showing a clear error toast with ingest counts.
Reuse overwrite handling for folder duplicates so users can choose to overwrite duplicates or skip duplicates and continue.
Improve duplicate dialog copy for single vs multiple duplicates, and keep action buttons readable/aligned.
Test Plan
Upload a folder with mixed supported/unsupported files and confirm toast shows “only X of Y will be ingested”.
Upload a folder with duplicates and choose Overwrite duplicates; verify duplicates are replaced.
Upload a folder with duplicates and choose Skip duplicates & continue; verify non-duplicates still ingest.

Before the user won't receive any info about the unsupported files in the folder, will just be ignore and the user wasn't able to overwrite duplicate in folder:
here in this example 2 duplicates (just skipped by default) and only 7 ingested which makes it 9files acknowledge over the 12 files in the folder, 3 of them are not supported type files:

https://github.com/user-attachments/assets/7ec039a9-7a68-4401-a860-dd7926ce2e9e

Now, the user get info about unsupported files and the ability to overwrite the duplicates files:
https://github.com/user-attachments/assets/a243c601-1666-4922-8533-4fd6f76d09bb

### DIFF
--- a/frontend/components/duplicate-handling-dialog.tsx
+++ b/frontend/components/duplicate-handling-dialog.tsx
@@ -42,7 +42,7 @@ export const DuplicateHandlingDialog: React.FC<
         ? "1 duplicate document already exists. Overwriting will replace the existing document version. This can't be undone."
         : `${duplicateCount} duplicate documents already exist. Overwriting will replace the existing document versions. This can't be undone.`
       : duplicateLabel
-        ? `A document with "${duplicateLabel}" already exists. Overwriting will replace the existing document version. This can't be undone.`
+        ? `A document named "${duplicateLabel}" already exists. Overwriting will replace the existing document version. This can't be undone.`
         : "Overwriting will replace the existing document with another version. This can't be undone.";
   const overwriteLabel =
     typeof duplicateCount === "number" ? "Overwrite duplicates" : "Overwrite";


frontend/components/knowledge-dropdown.tsx
```diff
@@ -296,7 +337,54 @@ export function KnowledgeDropdown() {
...
+
+    console.log(
+      `[Folder Upload] Uploading ${filesToUpload.length} file(s) in ${batches.length} batch(es), replace=${replace}`,
+    );
+
+    for (const batch of batches) {
+      try {
+        const result = await uploadFiles(batch, replace);
+        addTask(result.taskId);
+        console.log(
```
mfortman11: Cleanup the logs

frontend/components/knowledge-dropdown.tsx
```diff
@@ -97,6 +128,16 @@ export function KnowledgeDropdown() {
   const [s3Configured, setS3Configured] = useState(false);
   const [pendingFile, setPendingFile] = useState<File | null>(null);
   const [duplicateFilename, setDuplicateFilename] = useState<string>("");
+  const [duplicateCountForDialog, setDuplicateCountForDialog] = useState<
```
mfortman11: `pendingFolderUpload` already has the duplicate count why do we need this?

frontend/components/knowledge-dropdown.tsx
```diff
@@ -296,7 +337,54 @@ export function KnowledgeDropdown() {
...
+        pendingFolderUpload;
+      await uploadFolderBatches(allFiles, true);
+      const unsupportedMessage =
+        unsupportedCount > 0 ? `, skipped ${unsupportedCount} unsupported` : "";
+      toast.success(
+        `Processed ${allFiles.length} file(s), including ${duplicateCount} overwrite(s)${unsupportedMessage}`,
+      );
+      setPendingFolderUpload(null);
+      setDuplicateFilename("");
+      setDuplicateCountForDialog(undefined);
```
mfortman11: This reset block is used a couple times could make it reusable

frontend/components/knowledge-dropdown.tsx
Wallgau: good catch

PR #1266, opened 2026-03-25
fix: added check for podman compose inside startup checks

lucaseduoli:
This pull request expands and improves support for Podman Compose in the startup checks utility. The main focus is to ensure environments using Podman have the necessary compose tooling available, whether as a built-in subcommand or a standalone binary, and to provide automated installation options for `podman-compose` across platforms.

The most important changes are:

**Compose availability checks:**

* The `compose_available()` function now checks for both Docker Compose and Podman Compose, including the built-in `podman compose` subcommand and the standalone `podman-compose` binary. This ensures broader compatibility with different container runtimes and installation methods. [[1]](diffhunk://#diff-bc4b81a0c70cccd0bfc86382a4c665814c706e5aa02470b47699cd03d38b5b64L194-R194) [[2]](diffhunk://#diff-bc4b81a0c70cccd0bfc86382a4c665814c706e5aa02470b47699cd03d38b5b64L205-R217)

**Podman Compose installation:**

* Added a new `install_podman_compose()` function that installs the standalone `podman-compose` binary if it is not present, supporting macOS (via Homebrew) and Linux/WSL (via pip3 or the system package manager). This function is robust to different environments and provides user prompts and error handling.

**Startup check flow improvements:**

* Updated the `run_startup_checks()` function to always attempt installing `podman-compose` when Podman is the selected runtime, regardless of whether the built-in compose subcommand is available. This helps ensure all required dependencies are present for OpenRAG to function reliably with Podman.

### DIFF
--- a/src/tui/utils/startup_checks.py
+++ b/src/tui/utils/startup_checks.py
@@ -204,6 +204,7 @@ def compose_available() -> bool:
     # Try docker-compose (v1)
     if has_cmd("docker-compose"):
         return True
+    return False
 
 
 # =============================================================================


src/tui/utils/startup_checks.py
```diff
@@ -672,14 +744,16 @@ def run_startup_checks() -> bool:
...
 
-    # 6. Check compose
-    if not compose_available():
+    # 6. Check / install compose
+    if runtime == "podman":
+        # Always ensure the standalone podman-compose binary is present —
+        # it installs required dependencies even when the built-in subcommand exists.
+        if not install_podman_compose():
+            say("Warning: podman-compose is not available. OpenRAG may fail to start.")
+    elif not compose_available():
```
mpawlow, line 746: ### (a) [Critical] Dead Code: Podman Checks in compose_available() Are Never Reached in Main Flow

Problem

- compose_available() was extended to also detect podman compose and podman-compose 
- However, in run_startup_checks(), compose_available() is only called in the elif branch that explicitly excludes the podman runtime
- compose_available() is not called anywhere else in the codebase
- The podman-specific additions to compose_available() are therefore unreachable dead code in any real execution path

Proposed Solution

- Remove the podman-specific checks (lines 207–217) from compose_available() since they are unused — keep the function scoped to Docker only and rename/document it accordingly

src/tui/utils/startup_checks.py
```diff
@@ -283,6 +295,66 @@ def install_podman() -> bool:
...
+        except Exception as e:
+            say(f"Failed: {e}")
+            return False
+
+    elif plat in ("Linux", "WSL"):
+        say("Installing podman-compose (may prompt for sudo password)...")
+        # Prefer pip3 as it always has the latest version
+        if has_cmd("pip3"):
+            try:
+                subprocess.run(["pip3", "install", "--user", "podman-compose"], check=True)
```
mpawlow, line 322: ### (b) [Major] pip3 install --user Does Not Update PATH — Binary May Be Silently Unfindable After Install

Problem

- pip3 install --user podman-compose installs the binary into ~/.local/bin/, which may not be in the current process's PATH
- The function returns True indicating success, but has_cmd("podman-compose") would still return False for the remainder of the session
- Compare with install_homebrew() (lines 241–244), which correctly updates os.environ["PATH"] after installation
- The net effect: OpenRAG proceeds thinking podman-compose is installed, but the subsequent compose commands will fail

Proposed Solution

- After a successful pip3 install --user, update os.environ["PATH"] to include ~/.local/bin, consistent with how install_homebrew() handles this:
```shell
local_bin = str(Path.home() / ".local" / "bin")
if local_bin not in os.environ.get("PATH", ""):
  os.environ["PATH"] = f"{local_bin}:{os.environ['PATH']}"
```

src/tui/utils/startup_checks.py
```diff
@@ -283,6 +295,66 @@ def install_podman() -> bool:
...
+        try:
+            if has_cmd("apt-get"):
+                subprocess.run(["sudo", "apt-get", "update", "-y"], check=True)
+                subprocess.run(["sudo", "apt-get", "install", "-y", "podman-compose"], check=True)
+            elif has_cmd("dnf"):
+                subprocess.run(["sudo", "dnf", "install", "-y", "podman-compose"], check=True)
+            elif has_cmd("yum"):
+                subprocess.run(["sudo", "yum", "install", "-y", "podman-compose"], check=True)
+            elif has_cmd("pacman"):
+                subprocess.run(["sudo", "pacman", "-Sy", "--noconfirm", "podman-compose"], check=True)
```
mpawlow: ### (c) [Major] pacman -Sy Without -u Is a Dangerous Partial Upgrade

Problem

- pacman -Sy (sync database only) followed by package install, without -u (upgrade), is an Arch Linux anti-pattern known as a "partial upgrade"
- It can leave the system in a broken state because newly installed packages may link against library versions not yet installed
- The Arch Wiki explicitly warns against this: the correct pattern is pacman -Syu or pacman -S (if db is already synced)
- This bug exists in the pre-existing install_podman() code and is now copied into install_podman_compose()
- Also see line 286 (pre-existing)

Proposed Solution

- Replace pacman -Sy --noconfirm <pkg> with pacman -Syu --noconfirm <pkg> in both install_podman() and install_podman_compose() to perform a full system sync+upgrade before installing

src/tui/utils/startup_checks.py
```diff
@@ -202,7 +202,19 @@ def compose_available() -> bool:
...
+    if has_cmd("docker-compose"):
+        return True
+    # Try podman compose (built-in subcommand, podman v4.7+)
+    try:
+        result = subprocess.run(
+            ["podman", "compose", "version"], capture_output=True, timeout=5
+        )
+        if result.returncode == 0:
+            return True
+    except:
```
mpawlow: ### (d) [Minor] Bare except: Clauses Swallow KeyboardInterrupt and SystemExit

- Note: This is Minor severity. Please feel free to optionally implement or ignore.

Problem

- Two new bare except: blocks were added in compose_available() (lines 202 and 214)
- Bare except: catches everything including KeyboardInterrupt and SystemExit, meaning Ctrl+C during a subprocess call would be silently swallowed and the function would continue to the next check
- This is a pre-existing pattern in the file, but the new code perpetuates it
- Should use except Exception: to allow signal-based exits to propagate
- See also: Line 214

Proposed Solution

- Replace except: with except Exception: in both new blocks — consistent with the handling used elsewhere in the file (e.g., lines 105, 246, 292)

src/tui/utils/startup_checks.py
```diff
@@ -672,14 +744,16 @@ def run_startup_checks() -> bool:
...
             pass
 
-    # 6. Check compose
-    if not compose_available():
+    # 6. Check / install compose
+    if runtime == "podman":
+        # Always ensure the standalone podman-compose binary is present —
+        # it installs required dependencies even when the built-in subcommand exists.
+        if not install_podman_compose():
+            say("Warning: podman-compose is not available. OpenRAG may fail to start.")
```
mpawlow: ### (e) [Minor] Non-Blocking Failure on Missing podman-compose — May Lead to Silent Runtime Errors

- Note: This is Minor severity. Please feel free to optionally implement or ignore.

Problem

- When install_podman_compose() returns False (user declined or install failed), run_startup_checks() emits a warning but returns True and proceeds:
if not install_podman_compose():
  say("Warning: podman-compose is not available. OpenRAG may fail to start.")
- The "may fail" wording understates the likely outcome — if neither podman compose nor podman-compose exists, OpenRAG will definitely fail at the compose step
- No check is made afterward to see if compose_available() would pass via the built-in podman compose subcommand, which was the previous fallback

Solution

- After install_podman_compose() returns False, fall back to checking compose_available() to determine if the built-in podman compose subcommand would suffice — and only warn if both are absent:
```python
if not install_podman_compose() and not compose_available():
  say("Warning: No compose tool found. OpenRAG will fail to start.")
```

src/tui/utils/startup_checks.py
```diff
@@ -186,23 +186,24 @@ def podman_ready() -> bool:
...
         )
         if result.returncode == 0:
             return True
-    except:
+    except Exception:
         pass
     # Try docker-compose (v1)
-    return has_cmd("docker-compose")
+    if has_cmd("docker-compose"):
+        return True
```
mpawlow, line 206: ### (2a) [Minor] compose_available() Missing Explicit return False — Implicit None Return

- Note: This is Minor severity. Please feel free to optionally implement or ignore

Problem

- The current code ends at line 206 with return True inside an if block — if neither docker compose v2 nor docker-compose is found, the function falls off the end and returns None implicitly

Solutions

- Add return False as the final line of the function:

src/tui/utils/startup_checks.py
mpawlow: 🚀

src/tui/utils/startup_checks.py
mpawlow: 🚀

src/tui/utils/startup_checks.py
```diff
@@ -672,14 +736,16 @@ def run_startup_checks() -> bool:
             # User declined fix, but continue anyway
             pass
 
-    # 6. Check compose
-    if not compose_available():
+    # 6. Check / install compose
+    if runtime == "podman":
+        # Always ensure the standalone podman-compose binary is present —
+        # it installs required dependencies even when the built-in subcommand exists.
+        if not install_podman_compose() and not compose_available():
```
mpawlow, line 744: ### (2b) [Minor] False "No Compose Tool" Warning When podman compose Built-in Is Available

- Note: This is Minor severity. Please feel free to optionally implement or ignore

Problem

- The podman-specific checks were removed from compose_available() as part of addressing CR-001, but this created a regression in the compose fallback logic at line 743
- When runtime == "podman", install_podman_compose() returns False (user declined or install failed), the code falls back to compose_available() to check if any compose tool exists
- However, compose_available() now only checks for docker compose — it no longer checks podman compose (built-in subcommand) or podman-compose
- A user with a modern Podman v4.7+ installation (which ships podman compose built-in) who declines the podman-compose standalone install will see: Warning: No compose tool found. OpenRAG will fail to start.
- ...even though podman compose is fully functional and available
- This is a direct regression introduced by removing the podman checks from compose_available()

Solution

- Add a dedicated podman_compose_available() helper and use it in the fallback check:
```python
  def podman_compose_available() -> bool:
      """Check if podman compose (built-in) or podman-compose (standalone) is available."""
      try:
          result = subprocess.run(
              ["podman", "compose", "version"], capture_output=True, timeout=5
          )
          if result.returncode == 0:
              return True
      except Exception:
          pass
      return has_cmd("podman-compose")
```
- In run_startup_checks():
```python
  if not install_podman_compose() and not podman_compose_available():
      say("Warning: No compose tool found. OpenRAG will fail to start.")
```

src/tui/utils/startup_checks.py
```diff
@@ -186,23 +186,24 @@ def podman_ready() -> bool:
...
         )
         return result.returncode == 0
-    except:
+    except Exception:
         return False
 
 
 def compose_available() -> bool:
-    """Check if docker compose or docker-compose is available."""
+    """Check if docker compose, docker-compose, podman compose, or podman-compose is available."""
```
mpawlow: ### (2c) [Minor] Stale Docstring on compose_available() — Still References Podman

- Note: This is Minor severity. Please feel free to optionally implement or ignore

Problem

- The podman-specific checks (podman compose and podman-compose) were removed from the function body
- The docstring is now factually incorrect and misleads callers about what the function actually checks

Solution

- Update the docstring to match the current implementation:
```python
  """Check if docker compose (v2) or docker-compose (v1) is available."""
```

src/tui/utils/startup_checks.py
mpawlow: 🚀

src/tui/utils/startup_checks.py
mpawlow: 🚀

src/tui/utils/startup_checks.py
mpawlow: 🚀

src/tui/utils/startup_checks.py
lucaseduoli: that is because podman-compose is needed even if the user has a modern podman installation. That is the main purpose of this PR, because podman-compose installs a required dependency that fixes installation for some users

PR #1300, opened 2026-03-30
fix: add IBM user to the session

edwinjosechittilappilly:
This pull request updates the user authentication flow to ensure that users authenticated via IBM AMS cookie are properly tracked in the `session_manager`. Now, after a user is authenticated with IBM AMS, the user is added to the `session_manager.users` dictionary if they are not already present.

**Session management improvements:**

* After authenticating a user with IBM AMS cookie in `get_current_user`, the user is added to `session_manager.users` if not already tracked.
* Similarly, in `get_optional_user`, the user is added to `session_manager.users` if they have a `user_id` and are not already present.

### DIFF
--- a/src/dependencies.py
+++ b/src/dependencies.py
@@ -254,7 +254,7 @@ async def get_current_user(
     if IBM_AUTH_ENABLED:
         logger.debug("[IBM Auth] IBM auth mode enabled, getting current user")
         user = await _get_ibm_user(request, required=True)
-        if user and user.user_id not in session_manager.users:
+        if user and user.user_id and user.user_id not in session_manager.users:
             session_manager.users[user.user_id] = user
         return user
 


src/dependencies.py
```diff
@@ -253,7 +253,10 @@ async def get_current_user(
     # IBM AMS cookie auth takes priority when enabled
     if IBM_AUTH_ENABLED:
         logger.debug("[IBM Auth] IBM auth mode enabled, getting current user")
-        return await _get_ibm_user(request, required=True)
+        user = await _get_ibm_user(request, required=True)
+        if user and user.user_id not in session_manager.users:
```
mpawlow: ### (a) [Normal] Missing user.user_id truthiness check in get_current_user

- Similar to the Copilot feedback above

Problem

- get_current_user adds the IBM user to session_manager.users without first asserting that user.user_id is truthy. 
- In _get_ibm_user, the lh_credentials path (Option 0) creates a User with user_id=user_id where user_id is initialized to None
- This happens when the IBM JWT cookie is absent (so claims are never parsed) but the credentials header is still present. 
- If user.user_id is None, the expression user.user_id not in session_manager.users evaluates to True (because None is not yet a key), and session_manager.users[None] = user is silently executed, corrupting the users dict.
- Note: get_optional_user (the other site added in this PR) correctly guards with the extra user.user_id check:

```python
# get_optional_user — correct
if user and user.user_id and user.user_id not in session_manager.users:
```
```python
# get_current_user — missing check
if user and user.user_id not in session_manager.users:   # <-- no user.user_id truthiness
```

Solution

- Mirror the guard used in get_optional_user:
```python
if user and user.user_id and user.user_id not in session_manager.users:
  session_manager.users[user.user_id] = user
```

mpawlow: ### (b) [Normal] Stale IBM credentials never refreshed in session cache

Problem

- IBM credentials can rotate on every request
- Traefik injects fresh X-IBM-LH-Credentials on each forwarded call. 
- The new code only populates session_manager.users when the user_id is not already present
- Once a user is cached, subsequent requests with rotated credentials do not update the stored User object.
- Any component that calls session_manager.get_user(user_id) will receive the original (potentially stale) User instance 
- For OpenSearch client construction, IBM mode already bypasses the cache (if IBM_AUTH_ENABLED: return clients.create_user_opensearch_client(jwt_token)) so that specific path is safe. 
- But the pattern is fragile and any new call-site that reads credentials from the cached user would silently use stale data.
- Also affects: src/dependencies.py:301–303 (get_optional_user)

Solution

- Always overwrite the entry, regardless of whether it already exists. IBM users have no concept of a "login session" — each request is the authoritative source of truth:
```python
if user and user.user_id:
  session_manager.users[user.user_id] = user
```

mpawlow: ### (c) [Minor] Unbounded growth of session_manager.users in IBM multi-user deployments

- This is Minor severity. Please feel free to ignore or optionally implement

Problem

- session_manager.users is an in-memory plain dict with no eviction policy. 
- For IBM deployments with many unique users, this dict will grow indefinitely for the lifetime of the process.
- Note: This is a pre-existing structural issue; however, this PR extends the pattern to IBM auth, which is the primary auth mode in multi-tenant IBM deployments where user churn may be higher.
- Also affects src/dependencies.py:301–303

Solution

- Introduce a bounded LRU cache or TTL-based eviction in SessionManager 
   - e.g., use cachetools.TTLCache or cachetools.LRUCache as a drop-in replacement for the users dict.

edwinjosechittilappilly: it is done by the FE and the IBM APi key is stored locally in json

mpawlow: 🚀

PR #1304, opened 2026-03-30
fix: added number_of_replicas as 0 to not create unused nodes in OpenSearch

lucaseduoli:
This pull request focuses on optimizing OpenSearch index configurations by reducing the number of replicas to zero, which can improve indexing speed and reduce resource usage in development or single-node environments. The changes affect both the knowledge filters index and dynamically created embedding indices.

**OpenSearch index configuration improvements:**

* Set `"number_of_replicas": 0` in the knowledge filters index settings in `src/main.py` to ensure no replicas are created for this index.
* Set `"number_of_replicas": 0` (and remove the previous default of 1) for dynamically created embedding indices in `src/utils/embeddings.py`, further optimizing resource usage.

### DIFF
--- a/frontend/app/chat/_components/chat-input.tsx
+++ b/frontend/app/chat/_components/chat-input.tsx
@@ -22,6 +22,7 @@ import {
   PopoverAnchor,
   PopoverContent,
 } from "@/components/ui/popover";
+import { useIsCloudBrand } from "@/contexts/brand-context";
 import { useFileDrag } from "@/hooks/use-file-drag";
 import { cn } from "@/lib/utils";
 import { useGetAllFiltersQuery } from "../../api/queries/useGetAllFiltersQuery";
@@ -78,6 +79,7 @@ export const ChatInput = forwardRef<ChatInputHandle, ChatInputProps>(
     const [isWrapped, setIsWrapped] = useState(false);
     const isMultiline = input.includes("\n") || isWrapped;
     const isDragging = useFileDrag();
+    const isCloudBrand = useIsCloudBrand();
 
     // Internal state for filter dropdown
     const [isFilterDropdownOpen, setIsFilterDropdownOpen] = useState(false);
@@ -356,9 +358,15 @@ export const ChatInput = forwardRef<ChatInputHandle, ChatInputProps>(
           <div
             {...getRootProps()}
             className={cn(
-              "flex flex-col w-full p-2 rounded-xl border border-input transition-all",
+              "flex flex-col w-full p-2 rounded-xl border transition-all",
+              isCloudBrand
+                ? "border-[color:var(--chat-input-border)]"
+                : "border-input",
+              isCloudBrand ? "bg-[var(--chat-prompt-bg)]" : "bg-transparent",
               !isDragging &&
-                "hover:[&:not(:focus-within)]:border-muted-foreground focus-within:border-foreground",
+                (isCloudBrand
+                  ? "hover:[&:not(:focus-within)]:border-muted-foreground focus-within:border-[color:var(--chat-input-border)]"
+                  : "hover:[&:not(:focus-within)]:border-muted-foreground focus-within:border-foreground"),
               isDragging && "border-dashed",
             )}
           >
@@ -426,7 +434,12 @@ export const ChatInput = forwardRef<ChatInputHandle, ChatInputProps>(
                       type="button"
                       variant="ghost"
                       size="iconSm"
-                      className="h-8 w-8 p-0 rounded-md hover:bg-muted/50"
+                      className={cn(
+                        "h-8 w-8 p-0 rounded-md",
+                        isCloudBrand
+                          ? "hover:bg-[var(--layered-select-bg)]"
+                          : "hover:bg-muted/50",
+                      )}
                       onMouseDown={(e) => {
                         e.preventDefault();
                       }}
@@ -472,7 +485,12 @@ export const ChatInput = forwardRef<ChatInputHandle, ChatInputProps>(
                       type="button"
                       variant="ghost"
                       size="iconSm"
-                      className="h-8 w-8 p-0 rounded-md hover:bg-muted/50"
+                      className={cn(
+                        "h-8 w-8 p-0 rounded-md",
+                        isCloudBrand
+                          ? "hover:bg-[var(--layered-select-bg)]"
+                          : "hover:bg-muted/50",
+                      )}
                       onMouseDown={(e) => {
                         e.preventDefault();
                       }}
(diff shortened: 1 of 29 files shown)


src/utils/embeddings.py
```diff
@@ -156,9 +156,8 @@ async def create_dynamic_index_body(
 
     return {
         "settings": {
-            "index": {"knn": True},
+            "index": {"knn": True, "number_of_replicas": 0},
```
mpawlow: ### (a) [Normal] Inconsistent Settings Nesting in embeddings.py

Problem

- In embeddings.py, number_of_replicas was moved inside the "index" sub-key, while number_of_shards remains at the top level of "settings". 
- This creates an inconsistent structure within the same dict, and diverges from the canonical INDEX_BODY definition in settings.py where both shards and replicas are top-level under "settings".
- Before (pre-PR, only knn was nested):
```json
"settings": {
  "index": {"knn": True},
  "number_of_shards": 1,
  "number_of_replicas": 1,
}
```
- After (this PR):

```json
"settings": {
  "index": {"knn": True, "number_of_replicas": 0},  # replicas inside "index"
  "number_of_shards": 1,                             # shards still outside
}
```
- Canonical form in src/config/settings.py:143-148 (INDEX_BODY):
```json
"settings": {
  "index": {"knn": True},
  "number_of_shards": 1,
  "number_of_replicas": 0,   # top-level
}
```
- While OpenSearch accepts both forms, the mixed placement (number_of_replicas nested, number_of_shards flat) is harder to reason about and inconsistent with the rest of the codebase.

Solution

- Keep both settings at the same level, matching INDEX_BODY:
```json
"settings": {
  "index": {"knn": True},
  "number_of_shards": 1,
  "number_of_replicas": 0,
}
```

mpawlow: ### (b) [Normal] Hardcoded zero replicas may not be Production-Safe. May need to verify with SaaS

Problem

- Setting number_of_replicas: 0 unconditionally across all indices removes data redundancy with no override mechanism. 
- In multi-node production deployments (e.g., Kubernetes via the Helm charts in helm/openrag/), zero replicas means:
   - No failover if a data node goes down
   - Data loss risk during rolling upgrades
- This value was previously 1 (the OpenSearch default) for dynamic embedding indices. 
- There is no environment variable, config option, or comment noting the production trade-off.

Solution

- Introduce an env var (e.g., OPENSEARCH_NUMBER_OF_REPLICAS, defaulting to 0) and read it in settings.py, so production operators can override it without code changes.

src/main.py
```diff
@@ -249,17 +249,25 @@ async def init_index(opensearch_client=None):
...
+                body={"index": {"number_of_replicas": 0}}
+            )
             await TelemetryClient.send_event(
                 Category.OPENSEARCH_INDEX, MessageId.ORB_OS_INDEX_EXISTS
             )
 
         # Create knowledge filters index
         knowledge_filter_index_name = "knowledge_filters"
         knowledge_filter_index_body = {
+            "settings": {
```
mpawlow, line 276: ### (c) [Minor] knowledge_filters Index Missing Explicit number_of_shards

- Note: This is Minor severity. Please feel free to ignore or optionally implement

Problem

- The knowledge_filters index body in main.py now sets number_of_replicas but does not explicitly set number_of_shards. 
- All other index bodies in the codebase (INDEX_BODY, API_KEYS_INDEX_BODY, and the dynamic embedding index) explicitly set "number_of_shards": 1. 
- The implicit default in OpenSearch is 1 for newer versions (changed from 5 in OpenSearch 2.x) but not guaranteed across all supported versions, and the inconsistency makes the code harder to audit.

Solutions

- Add "number_of_shards": 1 explicitly to knowledge_filter_index_body["settings"]["index"] (or at top level) to match conventions established by INDEX_BODY and API_KEYS_INDEX_BODY.

src/utils/embeddings.py
mpawlow: ### (d) [Normal] put_settings Silently Overrides Production Replica Config on Every Startup

Problem

- The new put_settings call in the else branch runs unconditionally on every application startup when an index already exists. 
- This means:
   - A production operator who has deliberately set number_of_replicas > 0 for data redundancy will have it silently reset to 0 every time the app restarts.
   - Unlike index creation (which is a one-time operation), settings migration should be a best-effort, idempotent upgrade step — not a forced override on every boot.
- Runs on every restart, overrides any operator-configured replica count
```python
await os_client.indices.put_settings(
  index=index_name,
  body={"index": {"number_of_replicas": 0}}
)
```

- Also applicable to src/main.py:304-307 (knowledge_filters index)

Solutions

- Gate the put_settings call on reading the current replica count first, only updating if it differs from 0:
```python
current = await os_client.indices.get_settings(index=index_name)
current_replicas = int(
  current[index_name]["settings"]["index"].get("number_of_replicas", 1)
)
if current_replicas != 0:
  await os_client.indices.put_settings(...)
```
- Recommendation for (b): Introduce an env var (e.g., OPENSEARCH_NUMBER_OF_REPLICAS, default 0) and only enforce it if the var is explicitly set, allowing operators to retain control in production

src/main.py
```diff
@@ -249,17 +249,33 @@ async def init_index(opensearch_client=None):
...
             )
+            # Set number of replicas to 0 to not create unused nodes in OpenSearch, in case it was created with more replicas
+            current = await os_client.indices.get_settings(index=index_name)
+            current_replicas = int(
+                current[index_name]["settings"]["index"].get("number_of_replicas", 1)
+            )
+            if current_replicas != 0:
+                await os_client.indices.put_settings(
+                    index=index_name,
+                    body={"index": {"number_of_replicas": 0, "number_of_shards": 1}},
```
mpawlow: ### (2a) [Major] number_of_shards Included in put_settings - Immutable Setting Will Cause OpenSearch to Reject the Request

Problem

- number_of_shards is a static index setting in OpenSearch. 
- It cannot be changed after an index is created. Including it in a put_settings call against an existing, open index will cause OpenSearch to reject the request with:

▎ IllegalArgumentException: Can't update non dynamic settings [[index.number_of_shards]] for open indices

- This exception is unhandled — it propagates up through the outer try/except in init_index, which only intercepts disk-watermark errors and re-raises everything else (line 348). 
- The result: startup will fail for every existing deployment that triggers the replica migration path (i.e., any deployment with pre-existing indices that have number_of_replicas != 0).

```python
# documents index migration — INVALID: number_of_shards is static
await os_client.indices.put_settings(
  index=index_name,
  body={"index": {"number_of_replicas": 0, "number_of_shards": 1}},
)
```
```python
# knowledge_filters migration — same problem
await os_client.indices.put_settings(
  index=knowledge_filter_index_name,
  body={"index": {"number_of_replicas": 0, "number_of_shards": 1}},
)
```
Solution

- Remove number_of_shards from both put_settings bodies — only number_of_replicas is a dynamic setting that can be changed live:
```
body={"index": {"number_of_replicas": 0}}
```

PR #1319, opened 2026-04-01
fix: added graceful shutdown and entrypoint wrapper

### LINKED ISSUE #1170: [Bug]: TransportError(503,'search_phase_execution_exception') while uploading documents

lucaseduoli:
This pull request introduces improvements to the OpenSearch container orchestration and application shutdown process, focusing on more reliable startup, graceful shutdown, and better operational health management. The most significant changes include the addition of a custom OpenSearch entrypoint wrapper for graceful shutdown, enhancements to the Docker Compose setup for health checks and persistent data, and the implementation of a graceful OpenSearch client shutdown in the application.

**OpenSearch Container Entrypoint and Lifecycle Management:**
* Added a new `opensearch-entrypoint-wrapper.sh` script that starts OpenSearch, applies security setup after a delay, and handles graceful shutdown signals to ensure OpenSearch stops cleanly. (`opensearch-entrypoint-wrapper.sh`, `Dockerfile`) [[1]](diffhunk://#diff-b4e2abd3c200400ab79dcd92fb5e1291ea4f07a951ed56c0d0d908bff1730ef9R1-R30) [[2]](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R125-R129) [[3]](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L136-R142)
* Updated Docker Compose to use the new entrypoint, added a healthcheck, persistent named volume (`opensearch-data`), and improved stop behavior with a grace period. (`docker-compose.yml`) [[1]](diffhunk://#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3L13-R26) [[2]](diffhunk://#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R183-R185)

**Application Shutdown Improvements:**
* Implemented `graceful_opensearch_shutdown` in `utils/opensearch_utils.py` to ensure all operations complete and connections close properly on app shutdown. (`src/utils/opensearch_utils.py`)
* Updated the main application shutdown event to call this new shutdown function and improved logging around shutdown. (`src/main.py`)

**Code Quality and Minor Fixes:**
* Minor code cleanup and parameter fix in OpenSearch client usage. (`src/api/settings.py`) [[1]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1731-R1729) [[2]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1667-L1668)
* Commented out the direct image usage for `openrag-backend` in favor of building from source. (`docker-compose.yml`)

Closes #1170

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,7 +38,7 @@ services:
       - "5601:5601"
 
   openrag-backend:
-    # image: langflowai/openrag-backend:${OPENRAG_VERSION:-latest}
+    image: langflowai/openrag-backend:${OPENRAG_VERSION:-latest}
     build:
       context: .
       dockerfile: Dockerfile.backend


src/main.py
```diff
@@ -1861,15 +1861,25 @@ async def shutdown_event():
         await TelemetryClient.send_event(
             Category.APPLICATION_SHUTDOWN, MessageId.ORB_APP_SHUTDOWN
         )
+        logger.info("Application shutdown initiated")
+        
+        # Gracefully shutdown OpenSearch connection first
+        try:
+            from utils.opensearch_utils import graceful_opensearch_shutdown
+            await graceful_opensearch_shutdown(clients.opensearch)
```
mpawlow, line 1869: ## (a) [Major] Double-Close of OpenSearch Client on Shutdown

### Problem

- In `shutdown_event()`, `graceful_opensearch_shutdown(clients.opensearch)` is called first
  - It explicitly calls `await opensearch_client.close()`
- Immediately after, `await clients.cleanup()` is called
  - `clients.cleanup()` (in `settings.py`) checks `if self.opensearch is not None:` before calling `await self.opensearch.close()` again
- `graceful_opensearch_shutdown` **never sets the reference to `None`**, so `self.opensearch` remains non-`None`
  - This causes a second `close()` call on an already-closed client
- The inline comment acknowledges the risk ("this will also close OpenSearch client if not already closed") but provides no mechanism to prevent the double-close

```python
# main.py — shutdown_event()
await graceful_opensearch_shutdown(clients.opensearch)   # closes client
...
await clients.cleanup()   # also calls opensearch.close() again (settings.py:611)
```

```python
# settings.py:608-616
if self.opensearch is not None:
    try:
        await self.opensearch.close()   # second close
```

**Affected files:**
- `src/main.py:1869`
- `src/config/settings.py:608-616`

### Solutions

1. **(Recommended)** Null out the reference in `shutdown_event` after the graceful shutdown call:
   ```python
   await graceful_opensearch_shutdown(clients.opensearch)
   clients.opensearch = None   # prevent double-close in clients.cleanup()
   ```
2. Remove the separate `graceful_opensearch_shutdown` step entirely and rely on the existing `clients.cleanup()`, which already handles the OpenSearch close with error handling.

docker-compose.yml
```diff
@@ -10,26 +10,20 @@ services:
...
+      - OPENSEARCH_JAVA_OPTS=-Xms1g -Xmx1g
     ports:
       - "9200:9200"
       - "9600:9600"
     volumes:
-      - ${OPENSEARCH_DATA_PATH:-./opensearch-data}:/usr/share/opensearch/data:U,z
+      - opensearch-data:/usr/share/opensearch/data
+    stop_grace_period: 2m
+    healthcheck:
+      test: ["CMD-SHELL", "curl -ks https://localhost:9200 >/dev/null 2>&1 || exit 1"]
```
mpawlow: ## (b) [Normal] Healthcheck Does Not Verify OpenSearch Authentication

### Problem

- The Docker Compose healthcheck uses an unauthenticated `curl` request:
  ```yaml
  test: ["CMD-SHELL", "curl -ks https://localhost:9200 >/dev/null 2>&1 || exit 1"]
  ```
- `curl` exits `0` for **any** HTTP response, including `401 Unauthorized`
- After the security plugin is configured, the root endpoint returns `401`
  - The healthcheck still reports the container as healthy in this case
- Services using `depends_on` with `condition: service_healthy` would start even if security is not yet correctly configured
  - This can cause authentication failures at runtime

### Solutions

1. **(Recommended)** Include credentials and assert a successful response:
   ```yaml
   test: ["CMD-SHELL", "curl -ku admin:$$OPENSEARCH_PASSWORD https://localhost:9200/_cluster/health?wait_for_status=yellow&timeout=5s -s | grep -q '\"status\"'"]
   ```
2. Use `-w "%{http_code}"` to assert the HTTP status code explicitly rather than relying on curl's exit code.
3. Fall back to TCP-only checking (`nc -z localhost 9200`) but document the limitation that it does not validate auth.

opensearch-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+set -e
+
+# Start OpenSearch in the background
+/usr/share/opensearch/opensearch-docker-entrypoint.sh opensearch &
+OPENSEARCH_PID=$!
+
+# Function to handle shutdown signals
+shutdown() {
```
mpawlow, line 9: ## (b) [Normal] Healthcheck Does Not Verify OpenSearch Authentication

### Problem

- The Docker Compose healthcheck uses an unauthenticated `curl` request:
  ```yaml
  test: ["CMD-SHELL", "curl -ks https://localhost:9200 >/dev/null 2>&1 || exit 1"]
  ```
- `curl` exits `0` for **any** HTTP response, including `401 Unauthorized`
- After the security plugin is configured, the root endpoint returns `401`
  - The healthcheck still reports the container as healthy in this case
- Services using `depends_on` with `condition: service_healthy` would start even if security is not yet correctly configured
  - This can cause authentication failures at runtime

### Solutions

1. **(Recommended)** Include credentials and assert a successful response:
   ```yaml
   test: ["CMD-SHELL", "curl -ku admin:$$OPENSEARCH_PASSWORD https://localhost:9200/_cluster/health?wait_for_status=yellow&timeout=5s -s | grep -q '\"status\"'"]
   ```
2. Use `-w "%{http_code}"` to assert the HTTP status code explicitly rather than relying on curl's exit code.
3. Fall back to TCP-only checking (`nc -z localhost 9200`) but document the limitation that it does not validate auth.

mpawlow, line 9: ## (c) [Normal] No SIGKILL Timeout Fallback in Entrypoint Wrapper

### Problem

- The `shutdown()` trap sends `SIGTERM` to OpenSearch but then calls `wait "$OPENSEARCH_PID"` with no timeout:
  ```bash
  shutdown() {
      kill -TERM "$OPENSEARCH_PID" 2>/dev/null || true
      wait "$OPENSEARCH_PID"   # no timeout; waits forever if OpenSearch hangs
      exit 0
  }
  ```
- If OpenSearch hangs during shutdown (e.g., flushing segments), the wrapper waits indefinitely
- Docker's `stop_grace_period: 2m` is the only backstop
  - When that expires, Docker sends `SIGKILL` to the **wrapper** PID
  - The OpenSearch child process may not receive it if it is in its own process group

### Solutions

1. **(Recommended)** Add a timeout loop with a SIGKILL fallback inside the trap:
   ```bash
   shutdown() {
       echo "Received shutdown signal, stopping OpenSearch gracefully..."
       kill -TERM "$OPENSEARCH_PID" 2>/dev/null || true
       for i in $(seq 1 90); do
           kill -0 "$OPENSEARCH_PID" 2>/dev/null || break
           sleep 1
       done
       kill -KILL "$OPENSEARCH_PID" 2>/dev/null || true
       wait "$OPENSEARCH_PID"
       exit 0
   }
   ```
2. Use the `timeout` command: `timeout 90s bash -c "wait $OPENSEARCH_PID" || kill -KILL "$OPENSEARCH_PID"`.

src/utils/opensearch_utils.py
```diff
@@ -121,3 +121,42 @@ async def wait_for_opensearch(
...
+    """
+    if opensearch_client is None:
+        logger.debug("OpenSearch client is None, skipping graceful shutdown")
+        return
+    
+    try:
+        logger.info("Initiating graceful OpenSearch shutdown...")
+        
+        # Flush any pending operations by checking cluster health one last time
+        try:
```
mpawlow, line 143: ## (d) [Normal] Misleading Comment — "Flush Pending Operations"

### Problem

- The `graceful_opensearch_shutdown` function claims its cluster health check "flushes pending operations":
  ```python
  # Flush any pending operations by checking cluster health one last time
  try:
      await asyncio.wait_for(
          opensearch_client.cluster.health(),
          timeout=10.0
      )
  ```
- A `cluster.health()` call is a **read-only diagnostic query**
  - It does not flush write operations, drain bulk buffers, or close active shards
- The comment is factually incorrect and misleading to future maintainers
- If draining in-flight writes is the intent, a `/_flush` or `/_refresh` call would be needed

### Solutions

1. **(Recommended)** Fix the comment to accurately describe what the call does:
   ```python
   # Verify cluster is still reachable before closing the connection
   ```
2. Replace with an actual index flush if draining in-flight writes is a requirement:
   ```python
   await opensearch_client.indices.flush(index="_all", wait_if_ongoing=True)
   ```

docker-compose.yml
```diff
@@ -44,7 +38,7 @@ services:
       - "5601:5601"
 
   openrag-backend:
-    image: langflowai/openrag-backend:${OPENRAG_VERSION:-latest}
+    # image: langflowai/openrag-backend:${OPENRAG_VERSION:-latest}
```
mpawlow: ## (e) [Question] Production Image Commented Out in `docker-compose.yml` - Was this intentional?

### Problem

- The PR comments out the pre-built image and forces a local build for `openrag-backend`:
  ```yaml
  # image: langflowai/openrag-backend:${OPENRAG_VERSION:-latest}
  build:
    context: .
    dockerfile: Dockerfile.backend
  ```
- If left in production `docker-compose.yml`:
  - Every deployment builds from source, increasing startup time
  - Requires the full build context to be available at deploy time
- The PR description does not mention this as an intentional change

### Solutions

1. **(Recommended)** Restore the published image as the default and move build-from-source to a `docker-compose.override.yml` for local development.

docker-compose.yml
```diff
@@ -10,26 +10,20 @@ services:
...
+      - OPENSEARCH_JAVA_OPTS=-Xms1g -Xmx1g
     ports:
       - "9200:9200"
       - "9600:9600"
     volumes:
-      - ${OPENSEARCH_DATA_PATH:-./opensearch-data}:/usr/share/opensearch/data:U,z
+      - opensearch-data:/usr/share/opensearch/data
+    stop_grace_period: 2m
+    healthcheck:
+      test: ["CMD-SHELL", "curl -ku admin:$$OPENSEARCH_PASSWORD 'https://localhost:9200/_cluster/health' -s | grep -qE '\"status\":\"(green|yellow)\"'"]
```
mpawlow, line 21: Misplaced PR comment (b)
- See: https://github.com/langflow-ai/openrag/pull/1319/changes#r3024930796

docker-compose.yml
```diff
@@ -10,26 +10,20 @@ services:
...
+      interval: 15s
```
mpawlow, line 22: - One concern I have about the more complicated health check test (curl command + grep) is it might slightly more of a performance hit
- 15 seconds is a good health check interval
- However, the default 30 second interval might be more appropriate

lucaseduoli: 15s is fine I think

src/utils/opensearch_utils.py
mpawlow: 🚀

src/main.py
mpawlow: 🚀

opensearch-entrypoint-wrapper.sh
mpawlow: 🚀

mpawlow: 🚀

PR #1296, opened 2026-03-30
fix: added OpenSearch security calls under feature flag

lucaseduoli:
This pull request introduces a new set of OpenSearch security configurations specifically for cloud deployments, improves the application of Document-Level Security (DLS), and adds a comprehensive test script for validating ACL enforcement. It also updates Dockerfiles to include the new security configuration files and makes minor improvements to the OpenSearch index initialization process.

**Cloud Security Configuration:**

* Adds a new `cloud_securityconfig/` directory with OpenSearch security configuration files for cloud environments, including `roles.yml`, `roles_mapping.yml`, `config.yml`, `internal_users.yml`, and others. These files define custom roles (such as `openrag_user_role`), role mappings, authentication domains (OIDC and internal), and DLS rules for fine-grained access control. [[1]](diffhunk://#diff-1757422d792834bc54182e413cd1f99f26aafa9f3f8739a299201d17b933ce38R1-R5) [[2]](diffhunk://#diff-d0fb6fb4d0b6ad6d603486d3d4fe35939bb1822ca98b0946d57f21c3480b35efR1-R29) [[3]](diffhunk://#diff-4aa24b978c74e09d59131332239bb9e121ae18a5bcc6bf2a212d495107817730R1-R7) [[4]](diffhunk://#diff-e89b9a48c68b509da17a6224449c9bf69bf596f2c26cdc2204bcb0ce2485b7c8R1-R5) [[5]](diffhunk://#diff-4f8069dc790acc3690c0031d196c99c783753780999a82f936363c71d043b2bbR1-R57) [[6]](diffhunk://#diff-ea4d94d358a0a297561b96a92408811cad14e8bb64e9ba08b3c66ba7c7427e2dR1-R16) [[7]](diffhunk://#diff-c8ecb79e4cb1e38e26e1985776ed261cd5876165ae0c3aeea457e801f76e362dR1-R5)

* Updates the main `securityconfig/roles.yml` to include additional permissions and index patterns, ensuring parity with cloud settings and better DLS support. [[1]](diffhunk://#diff-fd4ec5a6e3af101ed9a0d495875b81d98fa241617ca60003688920227e7799f8R12) [[2]](diffhunk://#diff-fd4ec5a6e3af101ed9a0d495875b81d98fa241617ca60003688920227e7799f8R29-R57)

**Dockerfile and Build Updates:**

[...]

PR #1337, opened 2026-04-06
bug: Ingested documents are being shared will all users

ricofurtado:
Files ingested by any user (including via IBM COS + IBM SSO) are currently visible to ALL authenticated users. The owner, allowed_users, and allowed_groups fields are stored in OpenSearch during ingestion but are never filtered

### DIFF
--- a/src/api/acl.py
+++ b/src/api/acl.py
@@ -1,0 +1,179 @@
+from typing import List
+
+from fastapi import Depends
+from fastapi.responses import JSONResponse
+from pydantic import BaseModel
+
+from config.settings import get_index_name
+from dependencies import get_current_user, get_session_manager
+from session_manager import User
+from utils.acl_utils import update_document_acl
+from connectors.base import DocumentACL
+from utils.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+
+class ShareDocumentBody(BaseModel):
+    filename: str
+    user_ids: List[str]
+
+
+class UnshareDocumentBody(BaseModel):
+    filename: str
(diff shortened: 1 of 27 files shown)


edwinjosechittilappilly (changes requested):
The DLS releated changes are in https://github.com/langflow-ai/openrag/pull/1296

Lets merge this 1296 once tested and then revisit this functionality. cause this might chnage the ACL behaviour

src/api/acl.py
```diff
@@ -0,0 +1,179 @@
...
+from utils.acl_utils import update_document_acl
+from connectors.base import DocumentACL
+from utils.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+
+class ShareDocumentBody(BaseModel):
+    filename: str
+    user_ids: List[str]
```
edwinjosechittilappilly, line 19: It would be potentialy the owner name and email.

ACL to be be deduced or inherited from the Source for most connectors which priovide ACL.
Suggestion toRefer: dependecies.py for more context!

src/main.py
```diff
@@ -1751,6 +1752,26 @@ async def create_app():
         tags=["internal"],
     )
 
+    # Document ACL endpoints
+    app.add_api_route(
```
edwinjosechittilappilly, line 1756: Good one!
Lets plan and implement it with respect to the product and Design team.

I like this idea of unshare.

ricofurtado:
Closing as the issue was transferred to another PR with a broader scope

PR #1342, opened 2026-04-06
fix: centralize backend paths and accept env var to define them, change deployments to centralize read only and writable volumes

lucaseduoli:
This pull request introduces a major storage refactoring and breaking changes to the OpenRAG Helm chart, consolidating several PersistentVolumeClaims (PVCs) and updating the configuration and deployment logic accordingly. It also adds a migration job to help users upgrade safely and updates the Helm chart version to `0.2.0`. If you are upgrading from a previous version, please follow the migration instructions to avoid data loss.

**Major storage refactor and migration:**

* Consolidated legacy PVCs (`config`, `keys`, `documents`) into new `data` and `shared` PVCs, simplifying volume management for the backend and Langflow components. Old PVCs are deprecated and should be deleted after migration. (`kubernetes/helm/openrag/templates/backend/deployment.yaml` [[1]](diffhunk://#diff-e06f98ddead95ac084fceff4ce888a157b14038eb802a10ec2ccca37a47f63b1L139-R216) [[2]](diffhunk://#diff-e06f98ddead95ac084fceff4ce888a157b14038eb802a10ec2ccca37a47f63b1L203-R273); `kubernetes/helm/openrag/templates/langflow/deployment.yaml` [[3]](diffhunk://#diff-58bdbd991bb4229ab3b25a1bb54170139336db1e0dc7c10cc983b81f76936095R61-R63) [[4]](diffhunk://#diff-58bdbd991bb4229ab3b25a1bb54170139336db1e0dc7c10cc983b81f76936095R105-R109); `kubernetes/helm/openrag/templates/storage/backend-sc.yaml` [[5]](diffhunk://#diff-86792058366f606db851f8ba09e940bfe0217b1490fc9797048272d1101e387dL1-R1) [[6]](diffhunk://#diff-86792058366f606db851f8ba09e940bfe0217b1490fc9797048272d1101e387dL21-R21); `kubernetes/helm/openrag/templates/storage/config-pvc.yaml` [[7]](diffhunk://#diff-e27cf53823079b11e0e6c5c1a261edcafb27da7dbe21c25bf7eae139692dfd91L1-L18); `kubernetes/helm/openrag/templates/storage/keys-pvc.yaml` [[8]](diffhunk://#diff-d22f3b27ddb2ade2fc7a0ff35d6330b6e6ac165400de8fd28937819190aacb0fL1-L18); `kubernetes/helm/openrag/templates/storage/data-pvc.yaml` [[9]](diffhunk://#diff-7bf606961745681e47ef7714ba7418ee9745144c8372b7e90af8c231c16d38f5R1-R18)
* Added a migration job (`storage-migration`) to copy existing config and keys to the new data volume during upgrade. This job must be enabled in `values.yaml` when upgrading from older versions. (`kubernetes/helm/openrag/templates/storage/migration-job.yaml` [[1]](diffhunk://#diff-aff8d3dd86ea50d892dd7b75d03d7d3a97522cc97a3e7a759298f018c955c7f8R1-R65); `kubernetes/helm/openrag/templates/NOTES.txt` [[2]](diffhunk://#diff-6ab686a0277e14721f8647ea54adff3c286eba94548aa074ecdab8e909fcdbaeR80-R97)

**Deployment and configuration updates:**

* Updated backend and Langflow deployments to mount the new `data` and `shared` volumes, and removed legacy volume mounts and related init containers. (`kubernetes/helm/openrag/templates/backend/deployment.yaml` [[1]](diffhunk://#diff-e06f98ddead95ac084fceff4ce888a157b14038eb802a10ec2ccca37a47f63b1L139-R216) [[2]](diffhunk://#diff-e06f98ddead95ac084fceff4ce888a157b14038eb802a10ec2ccca37a47f63b1L203-R273); `kubernetes/helm/openrag/templates/langflow/deployment.yaml` [[3]](diffhunk://#diff-58bdbd991bb4229ab3b25a1bb54170139336db1e0dc7c10cc983b81f76936095L36-L107) [[4]](diffhunk://#diff-58bdbd991bb4229ab3b25a1bb54170139336db1e0dc7c10cc983b81f76936095R61-R63)
* Centralized path configuration using new environment variables and values in `values.yaml`, supporting custom directory layouts. (`kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml` [[1]](diffhunk://#diff-1849432f77af6e31f80cd40437aba26a12ba2e6f550dd04ca18366105cf2972eR169-R188); `.env.example` [[2]](diffhunk://#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cL203-R209); `docker-compose.yml` [[3]](diffhunk://#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R112)

**Default data loading and flows:**

* Improved logic for loading default documents and flows, now using the `shared` volume and supporting robust download with retry mechanisms for flows. (`kubernetes/helm/openrag/templates/backend/deployment.yaml` [kubernetes/helm/openrag/templates/backend/deployment.yamlL105-R177](diffhunk://#diff-e06f98ddead95ac084fceff4ce888a157b14038eb802a10ec2ccca37a47f63b1L105-R177))
* Changed the flag for disabling default document loading: `backend.persistence.documents.enabled` is removed; use `backend.persistence.documents.loadDefaults: false` instead. (`kubernetes/helm/openrag/templates/NOTES.txt` [kubernetes/helm/openrag/templates/NOTES.txtR80-R97](diffhunk://#diff-6ab686a0277e14721f8647ea54adff3c286eba94548aa074ecdab8e909fcdbaeR80-R97))

**Versioning and documentation:**

* Bumped Helm chart version to `0.2.0` and documented the breaking changes and migration steps in the release notes. (`kubernetes/helm/openrag/Chart.yaml` [[1]](diffhunk://#diff-d9caba0b96b1db620e666b8d7c810f602ae67fa9e2c6d873b8590c8b113dcedcL5-R5); `kubernetes/helm/openrag/templates/NOTES.txt` [[2]](diffhunk://#diff-6ab686a0277e14721f8647ea54adff3c286eba94548aa074ecdab8e909fcdbaeR80-R97)

Please review the release notes and migration instructions carefully before upgrading.


kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
```diff
@@ -166,9 +166,14 @@ stringData:
     {{- if .Values.global.imageTag }}
     OPENRAG_VERSION={{ .Values.global.imageTag | quote }}
     {{- end }}
-    {{- if .Values.backend.persistence.documents.mountPath }}
-    OPENRAG_DOCUMENTS_PATH={{ .Values.backend.persistence.documents.mountPath | quote }}
-    {{- end }}
+    # Centralized Path Configuration
```
rodageve, line 169: These paths should not be hard-coded, they should come from the helm values.yaml.

kubernetes/helm/openrag/templates/backend/deployment.yaml
```diff
@@ -34,25 +34,25 @@ spec:
...
+        {{- if and .Values.backend.persistence.shared.enabled .Values.backend.persistence.shared.documents.loadDefaults .Values.backend.persistence.shared.documents.git.enabled }}
         - name: load-default-documents
           image: curlimages/curl:8.5.0
           command: ["/bin/sh", "-c"]
           args:
             - |
               set -eu
 
-              DOCS_DIR="{{ .Values.backend.persistence.documents.mountPath }}"
+              DOCS_DIR="/shared/openrag-documents"
```
rodageve: Same thing here.  We want configurable paths with specific paths coming from the helm values file.
Keep `{{ .Values.backend.persistence.documents.mountPath }}` and put the value "/shared/openrag-documents" in values.yaml under `backend.persistence.documents.mountPath`

kubernetes/helm/openrag/templates/backend/deployment.yaml
```diff
@@ -102,8 +102,8 @@ spec:
                 echo "Documents already exist in $DOCS_DIR, skipping download."
               fi
           volumeMounts:
-            - name: documents
```
rodageve, line 105: This will break the initContainers (Requires updating the respective path in the initContainer code).

You will likely see exactly where this fails if you run `helm template test . --values values.yaml` from the `kubernetes/helm/openrag` folder

kubernetes/helm/openrag/templates/backend/deployment.yaml
```diff
@@ -136,22 +136,13 @@ spec:
...
-            - name: keys
-              mountPath: {{ .Values.backend.persistence.keys.mountPath }}
-            {{- end }}
-            {{- if .Values.backend.persistence.config.enabled }}
-            - name: config
-              mountPath: {{ .Values.backend.persistence.config.mountPath }}
-            {{- end }}
-            {{- if .Values.langflow.persistence.enabled }}
-            - name: flows
-              mountPath: /app/flows
```
rodageve, line 153: This also breaks the langflow initContainer

kubernetes/helm/openrag/templates/backend/deployment.yaml
```diff
@@ -102,8 +102,8 @@ spec:
                 echo "Documents already exist in $DOCS_DIR, skipping download."
               fi
           volumeMounts:
```
mpawlow, line 104: ## (a) [Critical] Backend Flows Volume Disconnected from Langflow — Flows No Longer Shared

### Problem

The old backend `deployment.yaml` explicitly mounted the `openrag-langflow` PVC so both services could access the same flow files. The new backend mounts `openrag-shared`. The `langflow/deployment.yaml` template is unchanged and still mounts `openrag-langflow`. No mechanism exists in the PR to synchronize these two volumes.

The backend previously mounted the **same** PVC as Langflow (`openrag-langflow`) so it could read flow JSON files directly from disk. The PR changes the backend's volume to a new `openrag-shared` PVC while leaving `langflow/deployment.yaml` completely unmodified. After this change:

- **Langflow** writes/reads flows from PVC `openrag-langflow` (mounted at `langflow.persistence.mountPath/flows`)
- **Backend** reads flows from PVC `openrag-shared` (mounted at `/shared`, with `OPENRAG_FLOWS_PATH=/shared/flows`)
- These are **different volumes** — the backend will see an empty (or absent) flows directory

Affected services include:
- `FlowsService._get_flows_directory()` — returns `get_flows_path()` = `/shared/flows`
- `LangflowFileService._ensure_url_ingest_flow_id()` — tries to load `openrag_url_mcp.json` from the flows path
- Component template paths (`WATSONX_LLM_COMPONENT_PATH`, etc.) in `settings.py` — computed as `<flows_path>/components/<filename>`

**Code References:**
- `kubernetes/helm/openrag/templates/backend/deployment.yaml` lines 79–83, 108–113 (new `shared` volume replacing `flows`/`langflow` volume)
- `kubernetes/helm/openrag/templates/langflow/deployment.yaml` lines 159–173 (still using `openrag-langflow` PVC — **not changed in this PR**)
- `src/services/flows_service.py` lines 83–84
- `src/services/langflow_file_service.py` line 356
- `src/config/settings.py` lines 436–471

### Solutions

1. **(Most recommended)** Update `langflow/deployment.yaml` to also mount the `openrag-shared` PVC (with the flows subpath) in addition to or instead of its current `langflow-data` volume. Set `LANGFLOW_LOAD_FLOWS_PATH` and `LANGFLOW_DATABASE_URL` accordingly, and update the `langflow-pvc.yaml` template if Langflow's database is also moving to `shared`.

2. Keep a separate `langflow` PVC for Langflow-internal data (database), and only move the flows JSON files to the `shared` PVC. Update the Langflow init container (`FLOWS_DIR`) to point to the `shared` volume's flows subdirectory.

kubernetes/helm/openrag/templates/storage/data-pvc.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if and .Values.backend.enabled .Values.backend.persistence.data.enabled }}
```
mpawlow, line 1: ## (b) [Major] Breaking Helm Upgrade — Data Loss Risk for JWT Keys and Config

### Problem

Helm `helm upgrade` does **not** delete PVCs on chart upgrade unless `helm uninstall` is run first. The old `keys` and `config` PVCs will be orphaned. The new `data` PVC will start empty, meaning JWT keys and `config.yaml` are gone on first pod restart after upgrade.

The PR deletes:
- `kubernetes/helm/openrag/templates/storage/config-pvc.yaml` (held `config.yaml` at `/app/config`)
- `kubernetes/helm/openrag/templates/storage/keys-pvc.yaml` (held `private_key.pem`/`public_key.pem` at `/app/keys`)

And adds:
- `kubernetes/helm/openrag/templates/storage/data-pvc.yaml` (new `data` PVC, starts empty)

On `helm upgrade` of an existing deployment:
- The old `*-config` and `*-keys` PVCs remain but are **no longer mounted** — their data is inaccessible
- The new `*-data` PVC is empty — `config.yaml` and JWT key files do not exist
- The backend will regenerate JWT keys on startup (breaking all existing sessions) and start with a blank `config.yaml` (losing onboarding state, LLM provider configuration, etc.)

**Code References:**
- `kubernetes/helm/openrag/templates/storage/config-pvc.yaml` (deleted)
- `kubernetes/helm/openrag/templates/storage/keys-pvc.yaml` (deleted)
- `kubernetes/helm/openrag/templates/storage/data-pvc.yaml` (added, line 1–18)

### Solutions

1. **(Most recommended)** Add a Helm upgrade notes section (`NOTES.txt` or PR description) providing explicit migration steps: before upgrading, copy data from the old PVCs to the new `data` PVC. Consider providing a one-time migration `Job` manifest that copies `/app/keys/*` and `/app/config/*` into `/data/keys/` and `/data/config/` before the backend starts.

2. Implement a startup check in the backend that detects if keys are missing and looks in legacy paths (e.g., `keys/` relative to working dir) before generating new ones — providing a grace period.

3. Version the chart with a major bump (`Chart.yaml` `version:`) to make the breaking change explicit and surface it in upgrade workflows.

kubernetes/helm/openrag/values.yaml
```diff
@@ -236,40 +236,41 @@ backend:
...
+      size: 10Gi
+      mountPath: /data
+      dataPath: "/data/data"
+      backupPath: "/data/backup"
+      configPath: "/data/config"
+      keysPath: "/data/keys"
+    shared:
+      enabled: true
+      storageClass: ""
+      accessMode: ReadWriteMany  # Often beneficial for shared content between backend/langflow
```
mpawlow, line 256: ## (c) [Major] `shared` PVC Requires `ReadWriteMany` — Not Universally Supported

### Problem

`ReadWriteMany` (RWX) is not supported by AWS EBS, standard GKE persistent disks, or Azure Disk Storage. Deployments on these common providers will fail with a `ProvisioningFailed` error unless an NFS/EFS-backed storage class is explicitly configured.

In `values.yaml`, the new `shared` PVC defaults to:

```yaml
shared:
  accessMode: ReadWriteMany  # Often beneficial for shared content between backend/langflow
```

The comment "Often beneficial" understates the requirement. If both the backend pod and a Langflow pod need to **write** to this volume simultaneously, RWX is a hard requirement. Clusters using cloud block storage (EBS, GCE PD, Azure Disk) do not support RWX. This will cause deployment failures silently or surface only during PVC provisioning.

**Code References:**
- `kubernetes/helm/openrag/values.yaml` line 256

### Solutions

1. **(Most recommended)** Update the comment to explicitly state this is a **requirement** when multiple pods write to the volume, and document a fallback for single-pod deployments:
   ```yaml
   # Required if both backend and langflow write to shared. Use ReadWriteOnce for single-node deployments.
   accessMode: ReadWriteMany
   ```
   Add a note in the Helm chart `README.md` or `NOTES.txt` linking to storage class requirements.

2. Consider using a `subPath` approach on a single `ReadWriteOnce` PVC if only one pod writes at a time (e.g., backend reads flows; Langflow is the only writer).

3. Provide a `values-single-node.yaml` override example showing `ReadWriteOnce` for environments without RWX support.

src/config/paths.py
```diff
@@ -0,0 +1,94 @@
+"""Centralized path helpers for OpenRAG.
```
mpawlow, line 1: ## (d) [Normal] Misleading Module Docstring — Env Vars Claimed to Be in `settings.py` But Are Not

### Problem

The assertion in the docstring is directly falsifiable by searching `settings.py`.

`src/config/paths.py` opens with:

```python
"""Centralized path helpers for OpenRAG.

The *environment variables* themselves are declared in ``config.settings``
(``OPENRAG_DOCUMENTS_PATH``, ``OPENRAG_KEYS_PATH``, etc.) for visibility.
This module reads them directly via ``os.getenv`` to avoid circular imports
(settings → config_manager → paths → settings).
"""
```

The claim is false. `settings.py` does not declare `OPENRAG_KEYS_PATH`, `OPENRAG_DATA_PATH`, `OPENRAG_CONFIG_PATH`, `OPENRAG_FLOWS_PATH`, or `OPENRAG_FLOWS_BACKUP_PATH` as module-level constants. Only `OPENRAG_DOCUMENTS_PATH` was previously present (and only in `.env.example`). The new environment variables are invisible to anyone reading `settings.py` for a complete picture of configurable env vars.

**Code References:**
- `src/config/paths.py` lines 1–7 (docstring)
- `src/config/settings.py` — no declarations of the new `OPENRAG_*_PATH` variables

### Solutions

1. **(Most recommended)** Either update the docstring to remove the false claim, or actually add the new env vars as module-level constants in `settings.py`:
   ```python
   # Path configuration (see config/paths.py for helpers)
   OPENRAG_DATA_PATH = os.getenv("OPENRAG_DATA_PATH", "data")
   OPENRAG_KEYS_PATH = os.getenv("OPENRAG_KEYS_PATH", "keys")
   OPENRAG_CONFIG_PATH = os.getenv("OPENRAG_CONFIG_PATH", "config")
   OPENRAG_FLOWS_PATH = os.getenv("OPENRAG_FLOWS_PATH", "flows")
   OPENRAG_FLOWS_BACKUP_PATH = os.getenv("OPENRAG_FLOWS_BACKUP_PATH", "")
   ```

2. Correct the docstring to reflect reality: "This module is the single source of truth for path env vars. Env vars are read directly via `os.getenv` to avoid circular imports."

src/config/paths.py
```diff
@@ -0,0 +1,94 @@
...
+# ---------------------------------------------------------------------------
+# Documents directory
+# ---------------------------------------------------------------------------
+def get_documents_path() -> str:
+    """Return the path to the documents directory.
+
+    Environment variable: OPENRAG_DOCUMENTS_PATH
+    Default: ``openrag-documents``  (relative to the working directory)
+    """
+    return os.getenv("OPENRAG_DOCUMENTS_PATH") or "openrag-documents"
```
mpawlow, line 20: ## (e) [Normal] New Path Env Vars Absent from `.env.example`

### Problem

 `.env.example` is the canonical reference for operators configuring the application. Missing entries leave the new configuration knobs undiscovered.

Five new environment variables control path behavior but are not documented in `.env.example`:
- `OPENRAG_KEYS_PATH`
- `OPENRAG_FLOWS_PATH`
- `OPENRAG_CONFIG_PATH`
- `OPENRAG_DATA_PATH`
- `OPENRAG_FLOWS_BACKUP_PATH`

Only `OPENRAG_DOCUMENTS_PATH` was already present (line 204). Without `.env.example` entries, operators deploying with Docker Compose or running locally have no way to discover these options unless they read `paths.py` source code.

**Code References:**
- `.env.example` line 204 (only existing `OPENRAG_DOCUMENTS_PATH`)
- `src/config/paths.py` lines 21, 33, 45, 66, 83 (env var names)

### Solutions

1. **(Most recommended)** Add commented entries to `.env.example` with defaults and descriptions, grouped with `OPENRAG_DOCUMENTS_PATH`:
   ```bash
   # Path configuration (override defaults for custom directory layouts)
   # OPENRAG_DATA_PATH=./data
   # OPENRAG_KEYS_PATH=./keys
   # OPENRAG_CONFIG_PATH=./config
   # OPENRAG_FLOWS_PATH=./flows
   # OPENRAG_FLOWS_BACKUP_PATH=  # defaults to <OPENRAG_FLOWS_PATH>/backup
   ```

kubernetes/helm/openrag/values.yaml
```diff
@@ -236,40 +236,41 @@ backend:
...
   persistence:
-    documents:
+    data:
       enabled: true
       storageClass: ""
       accessMode: ReadWriteOnce
+      size: 10Gi
+      mountPath: /data
+      dataPath: "/data/data"
+      backupPath: "/data/backup"
```
mpawlow, line 250: ## (f) [Minor] `OPENRAG_FLOWS_BACKUP_PATH` Semantics Inconsistent with Helm Deployment Location

- Note: This is Minor severity. Feel free to ignore or optionally implement

### Problem

The naming implies a path inside or alongside the flows directory, but the Helm chart places it under the data (writable) volume.

The env var `OPENRAG_FLOWS_BACKUP_PATH` implies backups are co-located with flows. The `get_flows_backup_path()` default also reinforces this:

```python
return os.getenv("OPENRAG_FLOWS_BACKUP_PATH") or os.path.join(get_flows_path(), "backup")
# → "flows/backup" by default (relative)
```

However, `values.yaml` sets:
```yaml
data:
  backupPath: "/data/backup"   # → OPENRAG_FLOWS_BACKUP_PATH = /data/backup
```

The backup directory is placed under the **data** (writable) PVC, not the **shared** (flows) PVC. For local dev the default is `flows/backup` (inside the flows directory). This creates an inconsistency between local dev behavior and Kubernetes deployment behavior for backup location, and the env var name doesn't hint that backups go to the data volume.

**Code References:**
- `src/config/paths.py` lines 48–54 (`get_flows_backup_path()`)
- `kubernetes/helm/openrag/values.yaml` line 250 (`backupPath: "/data/backup"`)

### Solutions

1. **(Most recommended)** Rename the env var to `OPENRAG_BACKUP_PATH` (dropping the `FLOWS_` prefix) and update the docstring default fallback description to clarify it defaults to `<flows_path>/backup` locally but should be set to a writable path in production.

2. Alternatively, update the Helm chart to set `backupPath: "/shared/flows/backup"` to make the deployment consistent with the default.

kubernetes/helm/openrag/values.yaml
```diff
@@ -236,40 +236,41 @@ backend:
...
+      storageClass: ""
+      accessMode: ReadWriteMany  # Often beneficial for shared content between backend/langflow
       size: 50Gi
-      mountPath: /app/openrag-documents
-      loadDefaults: true                        # Load default documents on first deployment
+      mountPath: /shared
+      flowsPath: "/shared/flows"
+    documents:
+      mountPath: "/shared/openrag-documents"
+      loadDefaults: true
```
mpawlow, line 262: ## (g) [Minor] `documents.enabled` Field Removed — Behavior Change Undocumented

- Note: This is Minor severity. Feel free to ignore or optionally implement

### Problem

Operators who previously set `backend.persistence.documents.enabled: false` to disable default document loading will find the flag silently ignored after upgrade.

The old `values.yaml` had a `backend.persistence.documents.enabled` field. The PR removes it, and the init container condition was updated from:

```yaml
{{- if and .Values.backend.persistence.documents.enabled .Values.backend.persistence.documents.loadDefaults ... }}
```

to:

```yaml
{{- if and .Values.backend.persistence.shared.enabled .Values.backend.persistence.documents.loadDefaults ... }}
```

Now default document loading is controlled by `shared.enabled` (which defaults to `true`). If an operator previously disabled documents via `documents.enabled: false`, this setting is now ignored and documents will be loaded (since `shared.enabled: true`). The `loadDefaults` flag provides equivalent control, but operators are not notified of the change.

**Code References:**
- `kubernetes/helm/openrag/templates/backend/deployment.yaml` line 37
- `kubernetes/helm/openrag/values.yaml` line 262 (`loadDefaults: true`)

### Solutions

1. **(Most recommended)** Add a Helm `NOTES.txt` or upgrade notes section explaining that `documents.enabled` is removed. Operators should use `backend.persistence.documents.loadDefaults: false` to disable default document loading.

2. If backward compatibility is desired, keep the `documents.loadDefaults` as the sole gate and add a Helm lint warning for the removed key.

mpawlow (changes requested):
@lucaseduoli 

## Code Review 1

- See PR comments: (a) to (g)

kubernetes/helm/openrag/values.yaml
lucaseduoli: We added this backups folder separately to make it be on the writable folder, since it needs write access and the flows folder doesn't need write access

lucaseduoli: And I left it the way it was on the docker compose so it wouldn't affect OSS users

kubernetes/helm/openrag/templates/backend/deployment.yaml
lucaseduoli: Flows don't need to be shared since your change that loads them with API calls

kubernetes/helm/openrag/templates/storage/migration-job.yaml
```diff
@@ -0,0 +1,65 @@
...
+        - name: new-data
+          mountPath: /new-data
+      volumes:
+      - name: legacy-config
+        persistentVolumeClaim:
+          claimName: {{ include "openrag.fullname" . }}-config
+      - name: legacy-keys
+        persistentVolumeClaim:
+          claimName: {{ include "openrag.fullname" . }}-keys
+      - name: new-data
```
mpawlow: ## (2a) [Normal] Migration Job `pre-upgrade` Hook References the New `data` PVC Before It Exists

### Problem

Helm's hook execution model is deterministic: `pre-upgrade` hooks run **before** any regular chart resources (including PVCs) are applied during `helm upgrade`.

The migration job is annotated as a `pre-upgrade` hook:

```yaml
annotations:
  "helm.sh/hook": pre-upgrade
  "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```

Its pod spec mounts the `new-data` volume backed by PVC `<release>-data`:

```yaml
- name: new-data
  persistentVolumeClaim:
    claimName: {{ include "openrag.fullname" . }}-data
```

However, `data-pvc.yaml` is a **regular chart template** (no hook annotation). During `helm upgrade`, Helm's execution order is:
1. `pre-upgrade` hooks run first → migration job pod is scheduled → mounts `<release>-data` PVC → **PVC does not exist yet** → pod stuck in `ContainerCreating` / fails
2. Regular resources applied → `data-pvc.yaml` creates the PVC

A user upgrading from chart `0.1.0` to `0.2.0` with `storageMigration.enabled: true` will get a failed or stuck Helm upgrade, which is the only safe upgrade path.

**Code References:**
- `kubernetes/helm/openrag/templates/storage/migration-job.yaml` lines 10, 62–64
- `kubernetes/helm/openrag/templates/storage/data-pvc.yaml` (no hook annotation)

### Solutions

1. **(Most recommended)** Add a `pre-upgrade` hook annotation with a lower weight to `data-pvc.yaml` so the PVC is created before the migration job runs:
   ```yaml
   annotations:
     "helm.sh/hook": pre-upgrade
     "helm.sh/hook-weight": "-5"
     "helm.sh/hook-delete-policy": before-hook-creation
   ```
   Then give the migration job `helm.sh/hook-weight: "0"` (default) to ensure ordering. Since `data-pvc.yaml` is still needed as a regular resource on fresh installs, also keep it as a regular template (Helm allows a resource to appear both as a hook and a regular resource by using separate files, or by using `pre-install,pre-upgrade` in the hook annotation).

2. Have the migration job's init container create the `data` PVC via `kubectl apply` before starting the copy. This requires a `ServiceAccount` with PVC create privileges, adding complexity.

3. Change the migration job to a `post-upgrade` hook (runs after resources are applied, so the PVC exists), but add an `initContainer` that waits for both the backend and data PVCs to be ready before proceeding. Document that the first backend pod start after upgrade will fail until migration completes and the pod restarts.

kubernetes/helm/openrag/templates/langflow/deployment.yaml
```diff
@@ -171,6 +102,11 @@ spec:
           {{- else }}
           emptyDir: {}
           {{- end }}
+        {{- if .Values.backend.persistence.shared.enabled }}
```
mpawlow, line 105: ## (2b) [Normal] `shared-pvc.yaml` Gated on `backend.enabled` Only — Langflow Mount Fails When Backend Is Disabled

### Problem

Edge case, but a genuine misconfiguration surface that produces a non-obvious scheduling failure.

`shared-pvc.yaml` is only rendered when `backend.enabled: true`:

```yaml
{{- if and .Values.backend.enabled .Values.backend.persistence.shared.enabled }}
```

However, `langflow/deployment.yaml` unconditionally mounts the `shared` PVC whenever `backend.persistence.shared.enabled` is true:

```yaml
{{- if .Values.backend.persistence.shared.enabled }}
- name: shared
  persistentVolumeClaim:
    claimName: {{ include "openrag.fullname" . }}-shared
{{- end }}
```

If `backend.enabled: false` and `backend.persistence.shared.enabled: true`, the PVC is never created but Langflow still tries to mount it. Langflow pods will fail to schedule with:
```
Unable to mount volumes for pod: PersistentVolumeClaim "xxx-shared" not found
```

**Code References:**
- `kubernetes/helm/openrag/templates/storage/shared-pvc.yaml` line 1
- `kubernetes/helm/openrag/templates/langflow/deployment.yaml` lines 105–109

### Solutions

1. **(Most recommended)** Update `shared-pvc.yaml` to also render when Langflow needs it:
   ```yaml
   {{- if and .Values.backend.persistence.shared.enabled (or .Values.backend.enabled .Values.langflow.enabled) }}
   ```

2. Update `langflow/deployment.yaml` to gate the volume reference on both conditions:
   ```yaml
   {{- if and .Values.backend.enabled .Values.backend.persistence.shared.enabled }}
   ```

PR #1347, opened 2026-04-06
fix: re-add litellm, fix ingestion without langflow for all providers, fix search functionality for all providers

### LINKED ISSUE #1173: [Bug]: Issue in knowledge search with Ollama embedding model
### OpenRAG Version

0.3.2

### Deployment Method

uv add (installed in project)

### Operating System

macOS Tahoe 26.3

### Python Version

3.13

### Affected Area

Retrieval (search, OpenSearch, hybrid search), Settings (configuration, model providers)

### Bug Description

When a document is ingested using an Ollama embedding provider(observed currently in case of qwen3-embedding:latest and nomic-embed-text:latest). Document completes ingestion successfully but when a query related to this uploaded document is given, an error message pops up with no results (failed to embed with model [name of the embedding model]). After this error the search feature stops working and shows the same message for each search. Even when there are no files in the list that are embedded using the Ollama models, this error pops up on running a search on the index. 

### Steps to Reproduce

Prerequisite:
1. Have Ollama installed in your system with models qwen3-embedding:latest and nomic-embed-text:latest installed.
2. Have Ollama setup and configured on your OpenRAG.
 
Steps:
1. Select the embedding model as nomic-embed-text:latest or qwen3-embedding:latest) under Settings tab.
2. Go to knowledge section and ingest the document.
3. Search for OPENSEARCH-7419-ZX in the searchbar.
4. Search using any other random query on other docs.
5. Go to settings and change the embedding provider to that of OpenAI(eg:- text-embedding-3-small).
6. Go to knowledge section and delete the doc uploaded earlier and re-upload the doc. 
7. Run a search for OPENSEARCH-7419-ZX.
8. Run search for other docs.

### Expected Behavior

The search should return results with the uploaded document as topmost result.

### Actual Behavior

Shows no results with error message as failed to embed with model [name of the embedding model].

### Relevant Logs

```shell

```

### Screenshots

Image

Image

### Additional Context

[OpenRAG.Index.Test.Document.txt](https://github.com/user-attachments/files/26082001/OpenRAG.Index.Test.Document.txt)

### Checklist

lucaseduoli:
This pull request introduces a centralized and robust approach to embedding model configuration and provider detection, improving maintainability and clarity across the codebase. The main changes include extracting embedding model constants into a dedicated module, refactoring model name formatting into a utility function, and updating dependencies and usages throughout the project to use these new utilities.

**Embedding Model Configuration and Utilities:**

* Introduced a new module `embedding_constants.py` to define embedding model names and their vector dimensions for both OpenAI and WatsonX providers, removing duplication from `settings.py`. [[1]](diffhunk://#diff-485bba07065528682b216399df4ca947d0e73849037da608addcbc3f239b599aR1-R22) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L109-L128)
* Added a utility function `get_formatted_model_name` in `model_utils.py` to standardize model naming conventions for LiteLLM/MCP compatibility, replacing scattered logic with a single source of truth. [[1]](diffhunk://#diff-b464c3f1bf931e79d1b303752f0947ab0accabdca02bf388ee1728ae35076162R1-R50) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L483-R481) [[3]](diffhunk://#diff-c9ae00ae2d057e89b4c38fa968ad73ae675ea77b0af90442708d1f612126fcc5L150-R155)

**Refactoring and Codebase Cleanup:**

* Refactored imports and usages throughout the codebase (`settings.py`, `search_service.py`, `embeddings.py`) to use the new `embedding_constants.py` and `get_formatted_model_name` utility, eliminating redundant or duplicated code. [[1]](diffhunk://#diff-c9ae00ae2d057e89b4c38fa968ad73ae675ea77b0af90442708d1f612126fcc5L5-R8) [[2]](diffhunk://#diff-b066a253c6e52d0624d045f4a76db0b26e77ee7eec74fd9dd8329531a1a69f9fL2-R3) F13f5c3cL13R1)
* Improved thread safety and clarity by using the imported `threading` module directly in client initialization.

**Dependency and Client Initialization Improvements:**

* Added the `agentd` dependency and removed unused dependencies (`openai`, `pyyaml`, `tiktoken`) from `pyproject.toml` to streamline the environment. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R24) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L44-R45)
* Enhanced OpenAI-compatible client initialization to use a dummy API key if none is provided, and ensured all client instantiations are routed through the MCP patch for provider-agnostic access. [[1]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690R442-R446) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L497-R505) [[3]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L539-R544)

**Service Enhancements:**

* Decorated the `search_tool` method with the `@tool` decorator to enable agentd tool integration.

These changes collectively improve the maintainability, clarity, and extensibility of the embedding model infrastructure and provider integration.

- Embedding Model Configuration:
  - Added `embedding_constants.py` to centralize embedding models and vector dimensions, removing hardcoded values from `settings.py`. [[1]](diffhunk://#diff-485bba07065528682b216399df4ca947d0e73849037da608addcbc3f239b599aR1-R22) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L109-L128)
  - Updated imports in `settings.py`, `search_service.py`, and `embeddings.py` to use the new constants. [[1]](diffhunk://#diff-c9ae00ae2d057e89b4c38fa968ad73ae675ea77b0af90442708d1f612126fcc5L5-R8) [[2]](diffhunk://#diff-b066a253c6e52d0624d045f4a76db0b26e77ee7eec74fd9dd8329531a1a69f9fL2-R3) F13f5c3cL13R1)

- Model Name Formatting Utility:
  - Added `get_formatted_model_name` in `model_utils.py` and refactored all model formatting logic to use this utility for consistent provider prefixing. [[1]](diffhunk://#diff-b464c3f1bf931e79d1b303752f0947ab0accabdca02bf388ee1728ae35076162R1-R50) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L483-R481) [[3]](diffhunk://#diff-c9ae00ae2d057e89b4c38fa968ad73ae675ea77b0af90442708d1f612126fcc5L150-R155)

- Dependency and Client Initialization:
  - Added `agentd` as a dependency and removed unused dependencies from `pyproject.toml`. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R24) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L44-R45)
  - Improved OpenAI client initialization to support dummy API keys and ensure all clients are patched for multi-provider compatibility. [[1]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690R442-R446) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L497-R505) [[3]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L539-R544)

- Service and Decorators:
  - Applied the `@tool` decorator to `search_tool` in `SearchService` for agentd integration.

- Thread Safety:
  - Switched to using the imported `threading` module for lock initialization in `settings.py`.
  
  
  Closes #1173


src/config/settings.py
```diff
@@ -480,11 +470,15 @@ def patched_async_client(self):
             except Exception as e:
                 logger.debug("Could not load provider credentials from config", error=str(e))
 
-            # Try to initialize the client - AsyncOpenAI() will read from environment
-            # We'll try HTTP/2 first with a probe, then fall back to HTTP/1.1 if it times out
-            import asyncio
-            import concurrent.futures
-            import threading
+            # Determine model and provider for both probe and production client
+            model_name = config.knowledge.embedding_model or EMBED_MODEL
```
mpawlow: ## (a) [Critical] `config` Variable Referenced Outside Try Block — `UnboundLocalError` on Startup Failure

### Summary

`config` is assigned inside a `try` block but referenced after the `except` clause; if `get_openrag_config()` raises, `config` is unbound and the code crashes with `UnboundLocalError` rather than falling back gracefully.

### Problem

- `config = get_openrag_config()` is on line 436 inside a `try` block
- The `except Exception` on line 470 swallows the error with only a `logger.debug` call
- Lines 474–475 then unconditionally dereference `config`:
  ```python
  model_name = config.knowledge.embedding_model or EMBED_MODEL
  provider = config.knowledge.embedding_provider or "openai"
  ```
- If the config file doesn't exist, is malformed, or the config manager throws for any reason, this results in an `UnboundLocalError: local variable 'config' referenced before assignment`, which propagates to callers instead of using safe defaults

### Potential Solution

Move the model/provider extraction into the `try` block, or provide a fallback after the `except`:

```python
try:
    config = get_openrag_config()
    # ... credential loading ...
    model_name = config.knowledge.embedding_model or EMBED_MODEL
    provider = config.knowledge.embedding_provider or "openai"
except Exception as e:
    logger.debug("Could not load provider credentials from config", error=str(e))
    model_name = EMBED_MODEL
    provider = "openai"
```

pyproject.toml
```diff
@@ -42,9 +43,7 @@ dependencies = [
     "textual-fspicker>=0.6.0",
     "structlog>=25.4.0",
     "zxcvbn>=4.5.0",
-    "openai>=1.0.0",
```
mpawlow, line 45: ## (b) [Critical] `openai` and `tiktoken` Removed from `pyproject.toml` But Still Directly Imported in `document_service.py`

### Summary

`openai` and `tiktoken` were removed as explicit dependencies from `pyproject.toml`, but `src/services/document_service.py` still directly imports both. They remain in the resolved lock file only as transitive dependencies of `litellm` and `agentd` — if either upstream package drops those deps in a future release, `document_service.py` will fail at import time with `ModuleNotFoundError`.

### Problem

- `pyproject.toml` lines 16–18 (deleted): `"openai>=1.0.0"`, `"tiktoken>=0.7.0"` were removed
- `src/services/document_service.py` lines 8–9 still contain:
  ```python
  import openai
  import tiktoken
  ```
- `tiktoken` is used extensively in `get_token_count()` (lines 19–28)
- `import openai` at line 8 appears to be an unused import (the `openai` symbol is never referenced in the file body) — a leftover that should also be removed
- Both packages are currently available only as transitive deps of `litellm` and `agentd`, making this implicitly fragile

### Code References

- `src/services/document_service.py`, lines 8–9, 19–28

### Potential Solution

- Add `tiktoken` back to `pyproject.toml` as a direct dependency (it is actively used):
  ```toml
  "tiktoken>=0.7.0",
  ```
- Remove the unused `import openai` from `document_service.py`

### Alternative Solutions

- **Alt 1**: Replace `tiktoken` usage in `document_service.py` with `litellm.token_counter()` which is now a direct dependency, eliminating the need for a separate `tiktoken` dep
- **Alt 2**: Keep both as explicit deps in `pyproject.toml` with a comment explaining they are also pulled in transitively, for clarity and resilience

pyproject.toml
```diff
@@ -42,9 +43,7 @@ dependencies = [
...
-    "pyyaml>=6.0",
```
mpawlow, line 46: ## (c) [High] `pyyaml` Removed as Explicit Dependency But Still Needed Directly

### Summary

`pyyaml` was removed from `pyproject.toml` but is still imported directly by three files. It currently resolves only as a transitive dependency of `agentd`, making it fragile against upstream changes.

### Problem

- `pyproject.toml` line 17 (deleted): `"pyyaml>=6.0"` was removed
- The following files still import `yaml` directly:
  - `src/config/config_manager.py` line 4: `import yaml` — used for reading/writing the YAML config file (core functionality)
  - `src/api/settings.py` line 1776: `import yaml` — used for config export
  - `src/tui/managers/container_manager.py` line 926: `import yaml`
- `pyyaml` currently arrives via `agentd`'s transitive dependency chain — if `agentd` removes it, all YAML config reads/writes would break at runtime

### Code References

- `src/config/config_manager.py`, line 4
- `src/api/settings.py`, line 1776
- `src/tui/managers/container_manager.py`, line 926

### Potential Solution

Re-add `pyyaml` as an explicit direct dependency in `pyproject.toml`:

```toml
"pyyaml>=6.0",
```

pyproject.toml
```diff
@@ -42,9 +43,7 @@ dependencies = [
...
-    "tiktoken>=0.7.0",
+    "litellm==1.83.3"
```
mpawlow: ## (d) [High] `litellm` Pinned to Exact Version — Breaks Downstream Dependency Resolution

### Summary

`litellm` is pinned to an exact version (`==1.83.3`) in `pyproject.toml`. This is an unusually fast-moving library and exact pinning causes dependency resolution failures for anyone who installs OpenRAG alongside other packages that require a different litellm version.

### Problem

- `pyproject.toml` line 46: `"litellm==1.83.3"` uses `==` (exact pin)
- `litellm` releases multiple versions per week and `==` prevents any version flexibility
- Projects that use `openrag` alongside another package requiring `litellm>=1.84` will get an irresolvable conflict at install time
- The lock file already canonically pins the version; the `pyproject.toml` specifier is for solver constraints, not for pinning
- Missing trailing comma after `"litellm==1.83.3"` also violates the formatting convention of the rest of the file (minor)

### Potential Solution

Use a minimum-version constraint and test periodically:

```toml
"litellm>=1.83.3",
```

src/services/search_service.py
```diff
@@ -147,26 +151,8 @@ async def embed_with_model(model_name):
...
-                        # Ollama models use tags with colons
-                        formatted_model = f"ollama/{model_name}"
-                        logger.debug(f"Formatted Ollama model: {model_name} -> {formatted_model}")
-                    elif model_name in WATSONX_EMBEDDING_DIMENSIONS:
-                        # WatsonX embedding models - use hardcoded list from settings
-                        formatted_model = f"watsonx/{model_name}"
-                        logger.debug(f"Formatted WatsonX model: {model_name} -> {formatted_model}")
-                    # else: OpenAI models don't need a prefix
+                # Use centralized utility for LiteLLM model formatting
+                formatted_model = get_formatted_model_name(model_name)
```
mpawlow: ## (e) [Medium] `get_formatted_model_name` in `search_service.py` Called Without `provider` — Falls Back to Heuristic-Only Detection

### Summary

In `search_service.py`, `get_formatted_model_name(model_name)` is called without providing the `provider` argument. This means provider detection falls back entirely to heuristics (colon-in-name = Ollama, known list = WatsonX), which silently misroutes any non-standard Ollama or WatsonX model names.

### Problem

- `src/services/search_service.py` line 155: `formatted_model = get_formatted_model_name(model_name)` — no `provider` arg
- In `settings.py` (line 478), the same utility is called with the provider: `get_formatted_model_name(model_name, provider)`, which is the correct usage
- The `search_service.py` call site has access to the configured provider at the point of use (it can call `get_openrag_config()` or read from `config.knowledge.embedding_provider`)
- Edge cases that break silently:
  - A custom WatsonX model not in `WATSONX_EMBEDDING_DIMENSIONS` would not get the `watsonx/` prefix
  - A non-standard Ollama model name without a colon (e.g., a local model tagged without a version) would be treated as an OpenAI model
  - A future provider added to `model_constants.py` would not be handled

### Code References

- `src/utils/model_utils.py`, lines 10–50

### Potential Solution

Pass the configured provider when formatting the model name inside `embed_with_model`:

```python
from config.settings import get_openrag_config

provider = get_openrag_config().knowledge.embedding_provider or "openai"
formatted_model = get_formatted_model_name(model_name, provider)
```

src/services/search_service.py
```diff
@@ -17,6 +20,7 @@ class SearchService:
     def __init__(self, session_manager=None):
         self.session_manager = session_manager
 
+    @tool
```
mpawlow: ## (f) [Medium] `@tool` Decorator Applied to Instance Method — Potential `agentd` Compatibility Issue

### Summary

The `@tool` decorator from `agentd.tool_decorator` is applied to `search_tool`, which is an instance method on `SearchService`. Tool decorators commonly expect standalone functions; wrapping instance methods can cause issues with how `agentd` introspects the signature (the `self` parameter may appear as a required tool argument).

### Problem

- `src/services/search_service.py` lines 23–24:
  ```python
  @tool
  async def search_tool(self, query: str, embedding_model: str = None) -> Dict[str, Any]:
  ```
- `agentd`'s `@tool` decorator likely inspects the function signature to register tool parameters; the `self` parameter would appear in that signature unless `agentd` explicitly handles bound methods
- There is no test coverage for the decorated method registration in the PR
- If `agentd` generates a tool schema including `self` as a parameter, any tool call from the agent framework would fail or behave unexpectedly

### Potential Solution

Confirm that `agentd.tool_decorator.tool` correctly handles instance methods (bound vs. unbound). If not, wrap the instance method call in a module-level function:

```python
# Module-level tool function
@tool
async def search_tool(query: str, embedding_model: str = None) -> Dict[str, Any]:
    """..."""
    service = SearchService()
    return await service._search_tool_impl(query, embedding_model)
```

src/config/embedding_constants.py
```diff
@@ -0,0 +1,22 @@
+"""Embedding model constants and dimensions."""
+
+EMBED_MODEL = "text-embedding-3-small"
```
mpawlow: ## (g) [Medium] Architectural Duplication: `EMBED_MODEL` Default Defined in Both `embedding_constants.py` and `model_constants.py`

### Summary

The default OpenAI embedding model (`text-embedding-3-small`) is now defined in two separate files, creating a split-brain risk where the two values can diverge.

### Problem

- `src/config/embedding_constants.py` line 3: `EMBED_MODEL = "text-embedding-3-small"`
- `src/config/model_constants.py` line 34: `OPENAI_DEFAULT_EMBEDDING_MODEL = "text-embedding-3-small"`
- The `CLAUDE.md` architecture guidance explicitly states: _"Model constants (validation lists, defaults, prefixes) centralized in `src/config/model_constants.py` — update this file when adding/removing supported models"_
- By introducing `embedding_constants.py` with its own `EMBED_MODEL`, the PR creates a second constants module that partially duplicates `model_constants.py` rather than extending it

### Code References

- `src/config/model_constants.py`, line 34

### Potential Solution

Remove `EMBED_MODEL` from `embedding_constants.py` and import it from `model_constants.py`:

```python
# embedding_constants.py
from config.model_constants import OPENAI_DEFAULT_EMBEDDING_MODEL as EMBED_MODEL
```

Or move `OPENAI_EMBEDDING_DIMENSIONS` and `WATSONX_EMBEDDING_DIMENSIONS` into `model_constants.py` and remove `embedding_constants.py` entirely, keeping all model constants in one place per the existing architectural pattern.

src/config/settings.py
```diff
@@ -494,20 +488,21 @@ async def probe_http2():
...
-                client = AsyncOpenAI()
-                logger.info("Probing OpenAI client with HTTP/2...")
+                # Use a patched client for the probe so it routes correctly based on the model
+                client = patch_openai_with_mcp(AsyncOpenAI(api_key=api_key))
+                logger.info(f"Probing client with HTTP/2 using model {formatted_model}...")
                 try:
                     await asyncio.wait_for(
                         client.embeddings.create(
-                            model='text-embedding-3-small',
+                            model=formatted_model,
```
mpawlow: ## (h) [Medium] HTTP/2 Probe Always Uses Configured Embedding Model — Real API Call on Every Cold Start

### Summary

The HTTP/2 probe makes a real embedding API call using the configured model on every cold start (first client access). For non-OpenAI providers (Ollama, WatsonX), this probe can fail legitimately and cause unnecessary fallback to HTTP/1.1, misleading logs, and — for metered APIs — unnecessary token usage.

### Problem

- `src/config/settings.py` lines 483–513: `probe_http2()` calls `client.embeddings.create(model=formatted_model, input=['test'])`
- For Ollama models, the probe connects to the local Ollama service. If Ollama isn't running at probe time (e.g., during CI or startup before services are ready), the probe fails and HTTP/1.1 is always used — even though HTTP/2 would have worked
- The original code probed with `text-embedding-3-small` unconditionally; the new code probes with the configured model, which is a correctness improvement but amplifies this timing sensitivity
- A 5-second timeout per probe, run on every cold start, adds latency during initialization

### Potential Solution

Skip the HTTP/2 probe for non-OpenAI providers and default to HTTP/1.1 (most local providers don't support HTTP/2 meaningfully):

```python
if provider_lower == "openai":
    use_http2 = _run_http2_probe(formatted_model, api_key)
else:
    use_http2 = False  # Local providers (Ollama) and WatsonX use HTTP/1.1
```

### Alternative Solutions

- **Alt 1**: Cache the HTTP/2 probe result to disk (e.g., in a `.http2_capable` marker file) so subsequent restarts don't re-probe
- **Alt 2**: Make the HTTP/2 probe opt-in via an env var (`OPENRAG_PREFER_HTTP2=true`) and default to HTTP/1.1

pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
 dependencies = [
+    "agentd>=0.2.2",
```
mpawlow: ## (i) [Low] `agentd` Version Constraint is Very Loose (`>=0.2.2`) But Lock File Resolves to `0.8.1`

### Summary

The `pyproject.toml` specifies `agentd>=0.2.2` but the lock file resolves to `0.8.1` — a significant version gap. If `agentd`'s API changed between `0.2.2` and `0.8.1`, the constraint would allow installing a version that doesn't have `patch_openai_with_mcp` or `tool_decorator`, causing `ImportError` at runtime on fresh installs without the lock file.

### Problem

- `pyproject.toml` line 24: `"agentd>=0.2.2"` — very loose lower bound
- `uv.lock` resolves to `agentd==0.8.1`
- `agentd` is used for `patch_openai_with_mcp` (`settings.py`) and `tool_decorator` (`search_service.py`) — if those APIs were introduced after `0.2.2`, the constraint allows installation of a broken version
- Users who install without the lock file (e.g., Docker builds that run `pip install`) would get whatever version `>=0.2.2` resolves to, which may not have these APIs

### Code References

- `uv.lock`, line 13 (version `0.8.1`)

### Potential Solution

Tighten the lower bound to match the resolved lock version or at minimum the version where the used APIs were introduced:

```toml
"agentd>=0.8.1",
```

pyproject.toml
mpawlow: ## (j) [Low] Missing Trailing Comma on `litellm` Dependency in `pyproject.toml`

### Summary

The new `litellm` dependency entry is missing a trailing comma, inconsistent with every other entry in the `dependencies` array.

### Problem

- `pyproject.toml` line 46: `"litellm==1.83.3"` — no trailing comma
- Every other dependency in the array has a trailing comma (e.g., `"zxcvbn>=4.5.0",`)
- While not a functional issue, it is a style inconsistency that may cause noise in future diffs and is inconsistent with `pyproject.toml` conventions

### Potential Solution

```toml
"litellm==1.83.3",
```

mpawlow (changes requested):
@lucaseduoli 

## Code Review 1

- See PR comments: (a) to (j)
- Note: Issues designated as **Low** severity can be safely ignored or optionally addressed (based on your judgement).

pyproject.toml
lucaseduoli: Locked to prevent further security breaches caused by the future releases. Asked by Prasa.

src/services/search_service.py
lucaseduoli: Can't pass provider: this part of the code gathers all of the embedding models of the files searched. If one of them don't match the current provider, it will fail.

src/config/settings.py
```diff
@@ -477,14 +467,20 @@ def patched_async_client(self):
...
+            # Format model name for LiteLLM compatibility (centralized logic)
+            formatted_model = get_formatted_model_name(model_name, provider)
 
-            # Try to initialize the client - AsyncOpenAI() will read from environment
-            # We'll try HTTP/2 first with a probe, then fall back to HTTP/1.1 if it times out
-            import asyncio
-            import concurrent.futures
-            import threading
+            # API key for AsyncOpenAI constructor
+            api_key = os.environ.get("OPENAI_API_KEY")
```
mpawlow, line 485: ## (2a) [Low] Dummy API Key Not Injected When `get_openrag_config()` Fails — Silent Auth Failure on Config Error

### Summary

The dummy `"no-key-required"` API key injection only happens inside the `try` block. If `get_openrag_config()` throws (e.g., missing or malformed config file), the exception handler falls back to `model_name`/`provider` defaults but never sets the dummy key. The subsequent HTTP/2 probe and client creation then receive `api_key=None`, which causes `AuthenticationError` on the first embedding request.

### Problem

- The dummy key logic is at lines 442–446, inside the `try` block:
  ```python
  elif not os.environ.get("OPENAI_API_KEY"):
      os.environ["OPENAI_API_KEY"] = "no-key-required"
  ```
- The `except` block (line 473–477) correctly sets `model_name` and `provider` fallbacks — addressing the previous `UnboundLocalError` — but does **not** set the dummy API key
- `api_key = os.environ.get("OPENAI_API_KEY")` at line 483 would therefore resolve to `None` if the env var was never set and config loading failed
- The HTTP/2 probe then runs (because `provider == "openai"` from the fallback) via `AsyncOpenAI(api_key=None)` — the OpenAI client will raise `AuthenticationError` when the probe makes its embedding request
- The outer `except Exception` at line 554 would catch it, but the raised `ValueError` provides a misleading error message about completing onboarding or setting `OPENAI_API_KEY`

### Code References

- `src/config/settings.py`, lines 442–446 (dummy key injection, inside `try` block)
- `src/config/settings.py`, lines 473–477 (except fallback — no dummy key set)
- `src/config/settings.py`, line 483 (`api_key = os.environ.get("OPENAI_API_KEY")`)
- `src/config/settings.py`, line 529 (HTTP/2 probe fires when `provider == "openai"`)

### Potential Solution

Add the dummy key fallback to the `except` clause alongside the model/provider fallbacks:

```python
except Exception as e:
    logger.debug("Could not load provider credentials from config", error=str(e))
    model_name = OPENAI_DEFAULT_EMBEDDING_MODEL
    provider = "openai"
    # Ensure a dummy key is available to satisfy the OpenAI constructor
    if not os.environ.get("OPENAI_API_KEY"):
        os.environ["OPENAI_API_KEY"] = "no-key-required"
```

src/services/search_service.py
```diff
@@ -13,9 +18,35 @@
...
+    if not _global_search_service:
+        logger.error("SearchService tool called before initialization")
+        return {"results": [], "error": "Search service not available"}
+    return await _global_search_service.search_tool(query, embedding_model=embedding_model)
+
+
 class SearchService:
     def __init__(self, session_manager=None):
         self.session_manager = session_manager
+        global _global_search_service
```
mpawlow: ## (2b) [Low] `_global_search_service` Overwritten on Every `SearchService` Instantiation — Test Interference and Stale-Instance Risk

### Summary

`SearchService.__init__` unconditionally overwrites the module-level `_global_search_service` singleton. Every new instantiation silently replaces the active service the global `search_tool` function points to. This causes test interference (each test that creates a `SearchService` changes the global state for all subsequent tests) and makes it impossible to distinguish between an intentional replacement and an accidental double-instantiation.

### Problem

- `src/services/search_service.py` lines 44–47:
  ```python
  def __init__(self, session_manager=None):
      self.session_manager = session_manager
      global _global_search_service
      _global_search_service = self
  ```
- `main.py` line 1320 creates one `SearchService(session_manager)` at startup — this is the intended production path
- However, the `@tool`-decorated wrapper at lines 23–40 has no visibility into whether the `_global_search_service` it holds is the intended live instance or a stale one from a prior call
- In tests, any test that imports `SearchService` and instantiates it (directly or indirectly) will silently redirect the tool to its own instance — making tests non-isolated with respect to the global
- No lock protects the write to `_global_search_service` — in a multi-threaded context (e.g., during startup), a race between two instantiations could leave the global pointing to the wrong instance

### Code References

- `src/services/search_service.py`, lines 19–20 (global declaration)
- `src/services/search_service.py`, lines 44–47 (`__init__` sets global)
- `src/services/search_service.py`, lines 37–40 (global `search_tool` wrapper reads the global)
- `src/main.py`, line 1320 (single production instantiation)

### Potential Solution

Register the service explicitly rather than implicitly from `__init__`. Add a class method or module-level function:

```python
_global_search_service = None

def register_search_service(service: "SearchService") -> None:
    """Explicitly register the active search service for the @tool wrapper."""
    global _global_search_service
    _global_search_service = service

class SearchService:
    def __init__(self, session_manager=None):
        self.session_manager = session_manager
        # Caller is responsible for registering via register_search_service()
```

Then in `main.py`:
```python
search_service = SearchService(session_manager)
register_search_service(search_service)
```

### Alternative Solutions

- **Alt 1**: Use a `threading.Lock` around the `_global_search_service` write and add a guard that only allows the first instantiation to register: `if _global_search_service is None: _global_search_service = self`
- **Alt 2**: Use a proper dependency-injection pattern: pass the service instance directly to the agent setup rather than relying on a global, removing the need for `_global_search_service` entirely

src/services/search_service.py
```diff
@@ -142,31 +173,22 @@ async def search_tool(self, query: str, embedding_model: str = None) -> Dict[str
             # Parallelize embedding generation for all models
             import asyncio
 
+            # Configure Ollama endpoint dynamically for LiteLLM routing
+            # (matches backend/docling expectations: host.docker.internal vs localhost)
+            config = get_openrag_config()
```
mpawlow: ## (2c) [Normal] Ollama Env Vars Set on Every Search Request — Concurrent Mutation and Per-Request Config Read

### Summary

The new Ollama endpoint injection code calls `get_openrag_config()` and mutates `os.environ` on every individual search request, inside the body of `search_tool`. This creates two distinct problems: (1) env var mutation from a potentially concurrent async context with no synchronization, and (2) config is re-read from disk on every search invocation rather than once at startup.

### Problem

- `src/services/search_service.py` lines 176–183 (commit `6e325d75`):
  ```python
  config = get_openrag_config()
  ollama_endpoint = config.providers.ollama.endpoint
  if ollama_endpoint:
      fixed_endpoint = transform_localhost_url(ollama_endpoint)
      os.environ["OLLAMA_API_BASE"] = fixed_endpoint
      os.environ["OLLAMA_BASE_URL"] = fixed_endpoint
  ```
- `os.environ` is a process-wide global; concurrent search requests running in the async event loop can interleave, so one request's endpoint write could clobber another's in-flight embedding call — particularly relevant if Ollama endpoint configuration changes mid-flight (e.g., onboarding update)
- `get_openrag_config()` reads the YAML config from disk on every invocation (depending on ConfigManager internals); in high-throughput scenarios this is unnecessary overhead per search request
- The code has no `try/except`: if `get_openrag_config()` raises (e.g., malformed config file), the entire `search_tool` call propagates the exception rather than falling back gracefully — unlike the analogous code in `settings.py` which has a try/except
- The `settings.py` `patched_async_client` property already sets `OLLAMA_BASE_URL` (line 466–467) at client-init time; this new code at search time duplicates that responsibility in a different place, making it unclear which site is authoritative

### Code References

- `src/services/search_service.py`, lines 176–183 (new Ollama env var injection)
- `src/config/settings.py`, lines 464–468 (existing Ollama endpoint injection at client-init time)

### Potential Solution

Move the Ollama endpoint injection out of the per-request path and into `SearchService.__init__` (one-time setup), matching the same pattern used in `settings.py`:

```python
class SearchService:
    def __init__(self, session_manager=None):
        self.session_manager = session_manager
        global _global_search_service
        _global_search_service = self
        self._configure_provider_env()

    def _configure_provider_env(self):
        """Set provider env vars once at init time."""
        try:
            config = get_openrag_config()
            if config.providers.ollama.endpoint:
                fixed = transform_localhost_url(config.providers.ollama.endpoint)
                os.environ.setdefault("OLLAMA_API_BASE", fixed)
                os.environ.setdefault("OLLAMA_BASE_URL", fixed)
        except Exception as e:
            logger.debug("Could not configure Ollama endpoint", error=str(e))
```

### Alternative Solutions

- **Alt 1**: Read the Ollama endpoint once and store it as an instance attribute on `SearchService`; use it directly in the LiteLLM call instead of via env vars, avoiding global mutation entirely
- **Alt 2**: Wrap the existing per-request injection in a `try/except` and cache the result on the instance so the config is only re-read when it may have changed (e.g., expose a `refresh()` method called by the settings API after config updates)

pyproject.toml
mpawlow: @lucaseduoli 
- `Question`: Sanity Check: Is the unused `import openai` at `document_service.py:8` dead code ?

mpawlow (commented):
## Code Review 2

- ✅  Approved / LGTM
- See: PR comments: (2a) to (2c) - for your consideration
- Note: Issues designated as **Low** severity can be safely ignored or optionally addressed (based on your judgement).
- **Functional Review 1** still pending

src/utils/model_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    if not model_name:
+        return ""
+        
+    # Skip formatting if already has a known provider prefix
+    if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
+        return model_name
+        
+    # Detect provider from model name characteristics if prefix is missing
+    if ":" in model_name:
```
edwinjosechittilappilly: is this safe? even openai can use that same right?

src/utils/model_utils.py
```diff
@@ -0,0 +1,50 @@
...
+        # Ollama models use tags with colons
+        formatted_model = f"ollama/{model_name}"
+        logger.debug(f"Detected Ollama model: {model_name} -> {formatted_model}")
+        return formatted_model
+        
+    if model_name in WATSONX_EMBEDDING_DIMENSIONS:
```
edwinjosechittilappilly: is this also safe since there could be models that are not in list?

lucaseduoli: this was an already existing logic, so I assumed it worked

src/utils/model_utils.py
lucaseduoli: this too is an already existing logic, but openai doesnt use the ":"

src/utils/model_utils.py
lucaseduoli: I will take a look and see if there is something else to do to detect if the model is from watsonx

src/config/settings.py
```diff
@@ -454,6 +439,11 @@ def patched_async_client(self):
                 if config.providers.openai.api_key:
                     os.environ["OPENAI_API_KEY"] = config.providers.openai.api_key
                     logger.debug("Loaded OpenAI API key from config")
+                elif not os.environ.get("OPENAI_API_KEY"):
+                    # Provide dummy key to satisfy AsyncOpenAI constructor; 
+                    # LiteLLM/MCP will handle routing to other providers if needed.
+                    os.environ["OPENAI_API_KEY"] = "no-key-required"
+                    logger.debug("Using dummy OpenAI API key to satisfy client constructor")
```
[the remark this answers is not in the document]
lucaseduoli: active embedding cant use openai if the key is missing

src/main.py
```diff
@@ -692,6 +686,7 @@ async def _ingest_default_documents_url(
 
         processor = DocumentFileProcessor(
             document_service,
+            models_service=models_service,
```
mpawlow, line 693: ## (3a) [Critical] Missing `models_service` in `ingest_openrag_docs_when_ready` URL path

### Summary

`_ingest_default_documents_url` now requires `models_service` as a positional argument, but the only path that calls it — `ingest_openrag_docs_when_ready` — never receives or passes `models_service`. This causes a `TypeError` at runtime for URL-based default document ingestion.

### Problem

- `_ingest_default_documents_url` signature (lines 664–669) is:
  ```python
  async def _ingest_default_documents_url(
      document_service,
      models_service,
      docs_url: str,
      crawl_depth: int,
  ):
  ```
- `ingest_openrag_docs_when_ready` (line 430) has no `models_service` parameter and calls `_ingest_default_documents_url` without it:
  ```python
  async def ingest_openrag_docs_when_ready(
      document_service, task_service, langflow_file_service, session_manager
  ):
      ...
      if DISABLE_INGEST_WITH_LANGFLOW:
          task_id = await _ingest_default_documents_url(
              document_service=document_service,
              docs_url=DEFAULT_DOCS_URL,
              crawl_depth=DEFAULT_DOCS_CRAWL_DEPTH,
          )  # TypeError: missing required argument 'models_service'
  ```
- This failure path is activated when `DISABLE_INGEST_WITH_LANGFLOW=True` and URL-based ingestion is configured.

### Code References

- `src/main.py` line 430 — function definition missing `models_service` parameter
- `src/main.py` lines 441–446 — call site missing `models_service` argument
- `src/main.py` lines 664–669 — function requires `models_service`

### Potential Solution

Thread `models_service` through `ingest_openrag_docs_when_ready`:
```python
async def ingest_openrag_docs_when_ready(
    document_service, models_service, task_service, langflow_file_service, session_manager
):
    ...
    if DISABLE_INGEST_WITH_LANGFLOW:
        task_id = await _ingest_default_documents_url(
            document_service=document_service,
            models_service=models_service,
            docs_url=DEFAULT_DOCS_URL,
            crawl_depth=DEFAULT_DOCS_CRAWL_DEPTH,
        )
```
Update all call sites accordingly.

### Alternative Solutions

- Make `models_service` an optional keyword argument with a `None` default in `_ingest_default_documents_url` and handle the `None` case inside `DocumentFileProcessor`. This avoids threading the parameter but loses the benefit of the models service for default doc ingestion.

src/services/models_service.py
```diff
@@ -1,25 +1,134 @@
...
+                    except Exception as e:
+                        logger.debug(f"Could not fetch WatsonX models for registry: {str(e)}")
+
+                ModelsService._model_provider_registry = new_registry
+                logger.info(f"Model registry updated: {len(ModelsService._model_provider_registry)} models registered")
+                
+            except Exception as e:
+                logger.error(f"Error updating model registry: {str(e)}")
+
+    async def get_litellm_model_name(self, model_name: str, provider: Optional[str] = None) -> str:
```
mpawlow, line 105: ## (3b) [Medium] `None` provider yields malformed `"None/model_name"` LiteLLM route

### Summary

When `get_litellm_model_name` cannot determine the provider from the registry (e.g., the model was never registered, or `update_model_registry` failed), `provider_lower` remains `None`. The return statement then produces `"None/model_name"` — an invalid LiteLLM model identifier that silently breaks all embedding calls.

### Problem

`get_litellm_model_name` (lines 105–129):
```python
async def get_litellm_model_name(self, model_name: str, provider: Optional[str] = None) -> str:
    """Synchronous version of model formatting for when provider is already known.
    
    This is useful for synchronous contexts (like properties) where the provider
    information is already available and doesn't need to be looked up in the registry.
    """

    if not model_name:
        return ""
        
    # Skip formatting if already has a known provider prefix
    if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
        return model_name
        
    # Check if provider is explicitly given and not "openai"
    provider_lower = provider.lower() if provider else None
    
    if provider_lower is None:
        # Try looking in registry
        provider_lower = ModelsService._model_provider_registry.get(model_name)
        if provider_lower is None:
            await self.update_model_registry()
            provider_lower = ModelsService._model_provider_registry.get(model_name)
        
    return f"{provider_lower}/{model_name}" if provider_lower != "openai" else model_name
```
- If `update_model_registry()` also fails to populate the model (e.g., API key not configured, network error, model not returned by provider), `provider_lower` is still `None`.
- `None != "openai"` is `True`, so the return value becomes `"None/text-embedding-3-small"` (or similar), which LiteLLM will not recognise.
- The embedding call will fail, breaking search for all users simultaneously. The error message ("Failed to embed with model …") gives no indication that the provider prefix is malformed.


### Potential Solution

Add an explicit guard and fall back to the raw model name when the provider is unknown:
```python
if provider_lower is None:
    logger.warning(
        "Could not determine provider for model; using model name as-is",
        model_name=model_name,
    )
    return model_name  # OpenAI-compatible models work without a prefix

return f"{provider_lower}/{model_name}" if provider_lower != "openai" else model_name
```

### Alternative Solutions

- Raise an explicit exception so callers can handle the unknown-provider case explicitly rather than silently sending a malformed name. This surfaces the misconfiguration earlier and with a clearer message.

src/models/processors.py
```diff
@@ -268,11 +264,23 @@ async def process_document_standard(
...
+
         for batch in text_batches:
             resp = await clients.patched_embedding_client.embeddings.create(
-                model=embedding_model, input=batch
+                model=formatted_model, input=batch
             )
-            embeddings.extend([d.embedding for d in resp.data])
+            embeddings.extend([d["embedding"] if isinstance(d, dict) else d.embedding for d in resp.data])
+
+        dimensions = len(embeddings[0])
```
mpawlow, line 286: ## (3c) [Medium] Unguarded `embeddings[0]` access can raise `IndexError` on empty documents

### Summary

`processors.py` accesses `embeddings[0]` to derive vector dimensions after the embedding loop. If a document produces no text chunks (empty file, corrupt OCR output, etc.), `embeddings` is empty and the access raises an unhandled `IndexError` that will surface as an opaque 500 error.

### Problem

```python
for batch in text_batches:
    resp = await clients.patched_embedding_client.embeddings.create(
        model=formatted_model, input=batch
    )
    embeddings.extend([d["embedding"] if isinstance(d, dict) else d.embedding for d in resp.data])

dimensions = len(embeddings[0])  # IndexError if embeddings is []
```
- If `text_batches` is empty (no text extracted from the document), the loop body never runs, `embeddings` is `[]`, and `embeddings[0]` raises `IndexError`.
- The error message gives no hint to the user that their document was empty or unreadable.

### Potential Solution

Guard the access and return a meaningful error:
```python
if not embeddings:
    logger.error(
        "No embeddings generated — document may be empty or unreadable",
        file_hash=file_hash,
        embedding_model=embedding_model,
    )
    return {"status": "error", "error": "No text content could be extracted from document"}

dimensions = len(embeddings[0])
```

### Alternative Solution

- Validate that `texts` is non-empty earlier in the function, before calling the embedding API, and return early with a descriptive status.

src/services/models_service.py
```diff
@@ -1,25 +1,134 @@
...
-    OPENAI_EMBEDDING_MODEL_PREFIX,
     OPENAI_VALIDATION_MODELS,
 )
+from config.embedding_constants import OPENAI_DEFAULT_EMBEDDING_MODEL, OPENAI_EMBEDDING_MODEL_PREFIX
 from utils.container_utils import transform_localhost_url
 from utils.logging_config import get_logger
 
 logger = get_logger(__name__)
+
+KNOWN_PREFIXES = ["openai", "ollama", "watsonx"]
```
mpawlow: ## (3d) [Medium] `KNOWN_PREFIXES` omits `"anthropic"`, creating inconsistency

- `Question`: Is this by design? 
   - I'm not sure about this one

### Summary

`KNOWN_PREFIXES = ["openai", "ollama", "watsonx"]` is used to detect already-prefixed model names in `get_litellm_model_name`. Anthropic is registered in the model registry but absent from `KNOWN_PREFIXES`. If a model name already carrying an `"anthropic/"` prefix is looked up (e.g., a future Anthropic embedding model), it would be double-prefixed as `"anthropic/anthropic/..."`.

### Problem

- The registry stores Anthropic language models under provider `"anthropic"`.
- The prefix check at line 116 only guards `["openai", "ollama", "watsonx"]`.
- If `provider_lower` is `"anthropic"` and the model name already includes an `"anthropic/"` prefix, the function returns `"anthropic/anthropic/..."`.
- While Anthropic currently provides no embedding models (so this code path is rarely exercised), the inconsistency creates a trap for future contributors.

### Code References

- `src/services/models_service.py` line 116: `if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):`

### Potential Solution

Add `"anthropic"` to `KNOWN_PREFIXES`:
```python
KNOWN_PREFIXES = ["openai", "ollama", "watsonx", "anthropic"]
```

### Alternative Solution

- Derive the prefix list dynamically from the set of known providers to avoid future drift.

mpawlow (changes requested):
@lucaseduoli 

### Code Review 3

- See PR comments: (3a) to (3d) - for your consideration
- 🚧  **Functional Review** still pending

mpawlow:
@lucaseduoli 

### Functional Review 1

- ✅  Approved / LGTM
- DISABLE_INGEST_WITH_LANGFLOW=false
   - ✅  Onboarding worked with Ollama
   - ✅  Ingestion worked with Ollama
   - ✅  Chat worked with Ollama
   - ✅  Ingestion worked with OpenAI
   - ✅  Chat worked with OpenAI
- DISABLE_INGEST_WITH_LANGFLOW=true  
   - ✅  Ingestion worked with Ollama
   - ✅  Ingestion worked with OpenAPI
   - ✅  Knowledge search worked with Ollama
   - ✅  Knowledge search worked with OpenAI

src/services/models_service.py
lucaseduoli: This shouldn't be needed since Anthropic doesn't have an embedding model, but I'm adding it anyway for consistency

src/models/processors.py
```diff
@@ -264,15 +260,40 @@ async def process_document_standard(
 
         texts = [c["text"] for c in slim_doc["chunks"]]
 
-        # Split into batches to avoid token limits (8191 limit, use 8000 with buffer)
-        text_batches = chunk_texts_for_embeddings(texts, max_tokens=8000)
+        litellm_embedding_model = await self.models_service.get_litellm_model_name(embedding_model) if self.models_service is not None else embedding_model
```
mpawlow, line 263: ### (4a) [Low] Redundant double call to get_litellm_model_name in TaskProcessor

- This is Low severity. Please feel free to ignore or optionally implemented

### Summary

get_litellm_model_name(embedding_model) is called twice in the same method on the same input, with each call capable of triggering a full registry refresh via update_model_registry().

### Problem

- Commit b99fba96 added a first call at line 263 to determine litellm_embedding_model for batch-size selection:
- A second, pre-existing call at lines 272–275 computes formatted_model for the embedding API request:
- Both calls receive the same embedding_model argument and return the same value
- get_litellm_model_name may internally call update_model_registry(), which makes HTTP requests to every configured provider — calling it twice doubles that cost per document ingestion
- The two None-guards are also inconsistent: is not None (line 263) vs. truthy check (line 272), though functionally equivalent here

### Code References

- src/models/processors.py lines 272–275 — second call (formatted_model)

### Potential Solution

Compute once; reuse the result:

```python
litellm_embedding_model = (
  await self.models_service.get_litellm_model_name(embedding_model)
  if self.models_service is not None
  else embedding_model
)

# Use for batch-size selection
if "ollama" in litellm_embedding_model:
  text_batches = chunk_texts_for_embeddings(texts, max_tokens=2000)
else:
  text_batches = chunk_texts_for_embeddings(texts, max_tokens=8000)

# Reuse as formatted_model
formatted_model = litellm_embedding_model
```

mpawlow (approved):
@lucaseduoli 

## Code Review 4

- ✅  Approved / LGTM
- See PR comment: (4a) 
   - (Low severity - optional)

PR #1352, opened 2026-04-07
feat: add SSL support to the backend

lucaseduoli:
This pull request adds support for enabling native HTTPS (SSL/TLS) on the backend server by allowing users to specify SSL certificate and key files via environment variables. The changes update documentation, configuration files, and the application code to make HTTPS setup straightforward in both Docker and Kubernetes deployments.

**HTTPS/SSL/TLS Support:**

* Added `SSL_KEYFILE` and `SSL_CERTFILE` environment variables to `.env.example`, allowing users to specify paths to SSL key and certificate files for enabling HTTPS on the backend.
* Updated `docker-compose.yml` to pass `SSL_KEYFILE` and `SSL_CERTFILE` environment variables to the backend container.
* Modified backend startup in `src/main.py` to read the SSL environment variables and start the server with HTTPS if they are set. [[1]](diffhunk://#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R2115-R2118) [[2]](diffhunk://#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312R2127-R2128)
* Updated Kubernetes Helm chart (`values.yaml`, `backend-dotenv.yaml`, and `deployment.yaml`) to support SSL configuration and allow setting the HTTP/HTTPS scheme for liveness and readiness probes. [[1]](diffhunk://#diff-6e1aeb87f3fd1e1bd54e01fa54852890611b685e10ee796db401b432c8773258R279-R283) [[2]](diffhunk://#diff-1849432f77af6e31f80cd40437aba26a12ba2e6f550dd04ca18366105cf2972eR197-R204) [[3]](diffhunk://#diff-e06f98ddead95ac084fceff4ce888a157b14038eb802a10ec2ccca37a47f63b1R164) [[4]](diffhunk://#diff-6e1aeb87f3fd1e1bd54e01fa54852890611b685e10ee796db401b432c8773258R313-R320)

**Documentation Updates:**

* Enhanced documentation to explain how to enable HTTPS in Docker deployments and described the new SSL configuration options in the configuration reference. [[1]](diffhunk://#diff-1b3fc896568878f93ac69bce7f649040274e5d4aa8e6c468f235443b5a9da124L123-R132) [[2]](diffhunk://#diff-72c3c3901a4d6baea27175427663092cc1878488406243cde36442d0d3265ed3R174-R175)


src/main.py
```diff
@@ -2112,6 +2112,10 @@ async def cleanup_subscriptions_proper(services):
     # Enable or disable HTTP access logging events
     access_log = os.getenv("ACCESS_LOG", "true").lower() == "true"
 
+    # SSL/TLS Configuration
+    ssl_keyfile = os.getenv("SSL_KEYFILE")
```
mpawlow, line 2116: ## (a) [Normal] Partial SSL configuration causes cryptic runtime failure

### Summary

If only one of `SSL_KEYFILE` or `SSL_CERTFILE` is set (not both), uvicorn will fail at startup with an opaque SSL-related error rather than a clear, actionable validation message.

### Problem

- `ssl_keyfile` and `ssl_certfile` are read independently from environment variables and passed directly to `uvicorn.run()` without any cross-validation
- A user who sets `SSL_KEYFILE` but forgets `SSL_CERTFILE` will get a low-level SSL context error from Python's `ssl` module or uvicorn, not a clear message about the misconfiguration
- The documentation states both must be set together, but there is no enforcement in code

### Solution

- Add explicit validation before calling `uvicorn.run()`:

```python
ssl_keyfile = os.getenv("SSL_KEYFILE") or None
ssl_certfile = os.getenv("SSL_CERTFILE") or None

if bool(ssl_keyfile) != bool(ssl_certfile):
    raise ValueError(
        "Both SSL_KEYFILE and SSL_CERTFILE must be set together to enable HTTPS. "
        f"Got SSL_KEYFILE={'set' if ssl_keyfile else 'unset'}, "
        f"SSL_CERTFILE={'set' if ssl_certfile else 'unset'}."
    )
```

docker-compose.yml
```diff
@@ -100,6 +100,8 @@ services:
       - LOG_LEVEL=${LOG_LEVEL:-INFO}
       - NO_COLOR=${NO_COLOR:-}
       - ACCESS_LOG=${ACCESS_LOG:-true}
+      - SSL_KEYFILE=${SSL_KEYFILE}
```
mpawlow, line 103: ## (b) [Normal] Empty-string env vars bypass `None` check when SSL vars are unset in Docker Compose

### Summary

`${SSL_KEYFILE}` and `${SSL_CERTFILE}` in `docker-compose.yml` lack the `:-` default syntax used for every other optional variable in the same file. When these variables are not defined in `.env`, Docker Compose passes them to the container as empty strings (`SSL_KEYFILE=`). `os.getenv("SSL_KEYFILE")` then returns `""` instead of `None`, potentially passing an empty path to uvicorn.

### Problem

- Every other optional env var in `docker-compose.yml` uses the `:-` pattern (e.g., `${NO_COLOR:-}`, `${LOG_LEVEL:-INFO}`)
- `${SSL_KEYFILE}` and `${SSL_CERTFILE}` do not, causing Docker Compose to emit a warning and pass empty strings to the container when the user has not set them
- `os.getenv("SSL_KEYFILE")` returns `""` (empty string, not `None`) when the container receives `SSL_KEYFILE=`
- Depending on whether uvicorn's internal SSL check uses `if ssl_keyfile is not None:` rather than `if ssl_keyfile:`, an empty string could trigger an attempt to open a file at path `""`, resulting in a `FileNotFoundError`
- The combined fix requires changes in both `docker-compose.yml` and `src/main.py`

### Solution

- Fix both locations:
   - `docker-compose.yml`: use `:-` to default to empty string:
     ```yaml
     - SSL_KEYFILE=${SSL_KEYFILE:-}
     - SSL_CERTFILE=${SSL_CERTFILE:-}
     ```
   - `src/main.py`: normalize empty strings to `None` (also addresses the partial-config validation in CR-001):
     ```python
     ssl_keyfile = os.getenv("SSL_KEYFILE") or None
     ssl_certfile = os.getenv("SSL_CERTFILE") or None
     ```

kubernetes/helm/openrag/values.yaml
```diff
@@ -305,12 +310,14 @@ backend:
     periodSeconds: 30
     timeoutSeconds: 10
     failureThreshold: 3
+    scheme: HTTP                          # HTTP or HTTPS
```
mpawlow, line 313: ## (c) [Minor] `readinessProbe.scheme` in `values.yaml` is unused and misleading

### Summary

`values.yaml` adds a `scheme: HTTP` field under `readinessProbe`, but the readiness probe in `deployment.yaml` uses `exec` (shell/curl), not `httpGet`. The `scheme` field is therefore never rendered into any Kubernetes manifest and is a dead, misleading value.

### Problem

- The liveness probe is correctly `httpGet`-based and the new `scheme` field is properly used via:
  ```yaml
  scheme: {{ .Values.backend.livenessProbe.scheme | default "HTTP" }}
  ```
- The readiness probe is `exec`-based (runs a shell script with curl) and has no `httpGet` block, so there is no place in the template that reads `.Values.backend.readinessProbe.scheme`
- An operator enabling SSL may set `readinessProbe.scheme: HTTPS` expecting the readiness probe to respect it, but nothing will change — silent misconfiguration

### Code References

- `kubernetes/helm/openrag/templates/backend/deployment.yaml:170-194` — the readiness probe block has no reference to `.Values.backend.readinessProbe.scheme`

### Solution

- Remove `scheme` from the `readinessProbe` section in `values.yaml`

src/main.py
mpawlow, line 2116: ## (d) [Minor] No startup log message indicating SSL is enabled or disabled

- Note: This is a Minor severity issue. Please feel free to ignore or optionally implement.

### Summary

When the backend starts, there is no log message indicating whether HTTPS is active. This makes it harder to diagnose deployment issues where SSL was intended but not enabled (e.g., due to a misconfigured path).

### Problem

- Every significant startup configuration (log level, access logging, etc.) is typically reflected in startup logs
- A silent SSL configuration means a misdeployment (HTTP when HTTPS was intended) could go unnoticed until a client connection is rejected
- The fix is a small, low-risk addition

### Solution

- Add a log line before `uvicorn.run()`:

```python
import structlog
logger = structlog.get_logger()

ssl_keyfile = os.getenv("SSL_KEYFILE") or None
ssl_certfile = os.getenv("SSL_CERTFILE") or None

if ssl_keyfile and ssl_certfile:
    logger.info("SSL enabled", keyfile=ssl_keyfile, certfile=ssl_certfile)
else:
    logger.info("SSL disabled, running in HTTP mode")
```

mpawlow, line 2116: ## (e) [Minor] No file existence check before passing SSL paths to uvicorn

### Summary

If `SSL_KEYFILE` or `SSL_CERTFILE` point to paths that do not exist in the container, uvicorn will fail at startup with a low-level Python `FileNotFoundError` or SSL context error, rather than a clear, actionable message.

### Problem

- File paths are passed directly to `uvicorn.run()` without verifying they are readable
- This is especially likely in Docker and Kubernetes deployments where the user must ensure the correct volume mount or secret is in place
- The error message from uvicorn/OpenSSL is generic and may not clearly identify which file is missing or why

### Solution

- Add a pre-flight check before starting:

```python
import os
ssl_keyfile = os.getenv("SSL_KEYFILE") or None
ssl_certfile = os.getenv("SSL_CERTFILE") or None

if ssl_keyfile and not os.path.isfile(ssl_keyfile):
    raise FileNotFoundError(f"SSL_KEYFILE not found: {ssl_keyfile!r}")
if ssl_certfile and not os.path.isfile(ssl_certfile):
    raise FileNotFoundError(f"SSL_CERTFILE not found: {ssl_certfile!r}")
```

mpawlow (changes requested):
@lucaseduoli 

## Code Review 1

- See PR comments: (a) to (e)

PR #1358, opened 2026-04-08
feat: SaaS handling of MCP/SDK

mfortman11:
This pull request introduces support for IBM authentication as an alternative to API key authentication throughout the OpenRAG codebase. It updates configuration handling, client instantiation, and authentication logic to enable seamless use of IBM credentials when the `IBM_AUTH_ENABLED` environment variable is set. The changes ensure that API key requirements are relaxed when IBM authentication is available, and headers are constructed dynamically based on the authentication method in use.

**Authentication logic updates:**

* The authentication dependency in `src/dependencies.py` now accepts IBM authentication via the `X-IBM-LH-Credentials` and `Authorization: Bearer <jwt>` headers when `IBM_AUTH_ENABLED` is set, constructing the `User` object accordingly.

**Configuration and client improvements:**

* The `Config` class in `sdks/mcp/src/openrag_mcp/config.py` no longer requires an API key if IBM authentication is enabled, and dynamically builds headers to include the API key only when present. [[1]](diffhunk://#diff-4ab415900c943ee5b416ddbb3f1410e29390a15e18b4dcc3fd969c404d3e6b22R46-R48) [[2]](diffhunk://#diff-4ab415900c943ee5b416ddbb3f1410e29390a15e18b4dcc3fd969c404d3e6b22L65-R69)
* The `get_openrag_client` function now instantiates the `OpenRAGClient` with the API key from the config, supporting cases where the API key may be absent due to IBM authentication.
* The `OpenRAGClient` in `sdks/python/openrag_sdk/client.py` no longer raises an error if the API key is missing, and its headers property only adds the `X-API-Key` header when an API key is available, supporting IBM authentication scenarios. [[1]](diffhunk://#diff-5f4d67278e841b52d68d6ee16092dee4df7117969b675b5dfb163c527b87654aR128-L137) [[2]](diffhunk://#diff-5f4d67278e841b52d68d6ee16092dee4df7117969b675b5dfb163c527b87654aL166-R166)

### DIFF
--- a/sdks/mcp/src/openrag_mcp/config.py
+++ b/sdks/mcp/src/openrag_mcp/config.py
@@ -43,9 +43,18 @@ class Config:
     def __init__(self):
         self.openrag_url = os.environ.get("OPENRAG_URL", "http://localhost:3000")
         self.api_key = os.environ.get("OPENRAG_API_KEY")
-        ibm_auth_enabled = _parse_bool("IBM_AUTH_ENABLED", False)
 
-        if not self.api_key and not ibm_auth_enabled:
+        # IBM auth mode: user provides IBM_USERNAME + IBM_API_KEY instead of OPENRAG_API_KEY.
+        # The MCP forwards these as headers on every SDK request to OpenRAG.
+        ibm_username = os.environ.get("IBM_USERNAME")
+        ibm_api_key = os.environ.get("IBM_API_KEY")
+        self.ibm_extra_headers: dict[str, str] = {}
+        if ibm_username:
+            self.ibm_extra_headers["X-Username"] = ibm_username
+        if ibm_api_key:
+            self.ibm_extra_headers["X-Api-Key"] = ibm_api_key
+
+        if not self.api_key and not self.ibm_extra_headers:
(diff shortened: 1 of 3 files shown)


sdks/mcp/src/openrag_mcp/config.py
```diff
@@ -43,8 +43,9 @@ class Config:
     def __init__(self):
         self.openrag_url = os.environ.get("OPENRAG_URL", "http://localhost:3000")
         self.api_key = os.environ.get("OPENRAG_API_KEY")
+        ibm_auth_enabled = _parse_bool("IBM_AUTH_ENABLED", False)
```
edwinjosechittilappilly: This is the tricky part since MCP app / pip install would not know about IBM Auth.
how about we handle it wrt to keys/ variables provided.
lets connect async on it

edwinjosechittilappilly: Like the X-USERNAME etc?

edwinjosechittilappilly (approved):
LGTM awaiting tests

PR #1363, opened 2026-04-09
fix: change factory-reset functions to delete named volume for opensearch-data

lucaseduoli:
This pull request removes support for configuring the OpenSearch data path via environment variables and local directories, fully transitioning OpenSearch data persistence to Docker-managed named volumes. This simplifies configuration and data management, making the system more robust and less error-prone. All related documentation, configuration fields, and code paths for the old directory-based approach have been removed or updated.

**Configuration and Environment Variable Cleanup:**

* Removed `OPENSEARCH_DATA_PATH` from `.env.example`, the `EnvManager`, and all related config fields, so OpenSearch data is no longer configurable via environment variables or local paths. [[1]](diffhunk://#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cL73-L76) [[2]](diffhunk://#diff-5408077f55d63f67ee40e20ce9b6b1a19f84af6dfe1efaab8a8b4a731739bbc7L100) [[3]](diffhunk://#diff-5408077f55d63f67ee40e20ce9b6b1a19f84af6dfe1efaab8a8b4a731739bbc7L226) [[4]](diffhunk://#diff-5408077f55d63f67ee40e20ce9b6b1a19f84af6dfe1efaab8a8b4a731739bbc7L509-L511) [[5]](diffhunk://#diff-6e41c9c8008a4d431a159062b3a21caf92777fca9736a36db3e84f5bbe894f95L90-L95)

**Makefile and Data Reset Updates:**

* Updated the `factory-reset` and `clear-os-data` Makefile targets to remove references to the `opensearch-data` directory and clarify that OpenSearch data is now managed as a Docker volume. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L509) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L527-L533) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L1017-R1012)

**Script and Container Manager Refactor:**

* Refactored `scripts/clear_opensearch_data.py` and `ContainerManager.clear_opensearch_data_volume()` to operate on Docker volumes instead of directories, including logic to handle different possible volume names. [[1]](diffhunk://#diff-66b43037b92e4970806ee19b5321c13e55e0089a307a9c0cd8617c517e092fabL15-L37) [[2]](diffhunk://#diff-26f4c29437cf0f797b720758fa7be3ce43e5018327378cf27893afd79f038c9eL1281-R1315)

**Miscellaneous:**

* Adjusted logic for inserting the `OPENRAG_VERSION` line in `.env` files to account for the removal of `OPENSEARCH_DATA_PATH`.

### DIFF
--- a/kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
+++ b/kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
@@ -70,7 +70,6 @@ stringData:
     {{- if .Values.global.opensearch.password }}
     OPENSEARCH_PASSWORD={{ .Values.global.opensearch.password | quote }}
     {{- end }}
-    OPENSEARCH_DATA_PATH=./opensearch-data
     OPENSEARCH_HOST={{ include "openrag.opensearch.host" . | quote }}
     OPENSEARCH_PORT={{ .Values.global.opensearch.port | quote }}
     {{- if .Values.global.opensearch.username }}
--- a/src/tui/main.py
+++ b/src/tui/main.py
@@ -562,7 +562,7 @@ def migrate_legacy_data_directories():
     """Migrate data from CWD-based directories to ~/.openrag/.
 
     This is a one-time migration for users upgrading from the old layout.
-    Migrates: documents, flows, keys, config, opensearch-data, langflow-data
+    Migrates: documents, flows, keys, config, langflow-data
 
     Prompts user for confirmation before migrating. If user declines,
     exits with a message to downgrade to v1.52 or earlier.
@@ -584,7 +584,8 @@ def migrate_legacy_data_directories():
         (cwd / "flows", target_base / "flows", "flows"),
         (cwd / "keys", target_base / "keys", "keys"),
         (cwd / "config", target_base / "config", "config"),
-        (cwd / "opensearch-data", target_base / "data" / "opensearch-data", "OpenSearch data"),
+        # opensearch-data intentionally omitted: OpenSearch now uses a Docker named
+        # volume; old index data is not portable and must be re-ingested.
         (cwd / "langflow-data", target_base / "data" / "langflow-data", "Langflow data"),
     ]
 
@@ -753,7 +754,6 @@ def setup_host_directories():
     - ~/.openrag/keys/ (for JWT keys)
     - ~/.openrag/config/ (for configuration)
     - ~/.openrag/data/ (for backend data: conversations, OAuth tokens, etc.)
-    - ~/.openrag/data/opensearch-data/ (for OpenSearch index)
     - LANGFLOW_DATA_PATH (for Langflow database and state)
     """
     base_dir = Path.home() / ".openrag"
@@ -763,7 +763,6 @@ def setup_host_directories():
         base_dir / "keys",
         base_dir / "config",
         base_dir / "data",
-        base_dir / "data" / "opensearch-data",
     ]
 
     for directory in directories:
--- a/src/tui/main.py
+++ b/src/tui/main.py
@@ -608,7 +608,6 @@ def migrate_legacy_data_directories():
             env_manager.config.openrag_flows_path = f"{home}/.openrag/flows"
             env_manager.config.openrag_config_path = f"{home}/.openrag/config"
             env_manager.config.openrag_data_path = f"{home}/.openrag/data"
-            env_manager.config.opensearch_data_path = f"{home}/.openrag/data/opensearch-data"
             env_manager.config.langflow_data_path = f"{home}/.openrag/data/langflow-data"
             env_manager.save_env_file()
             logger.info("Updated .env file with centralized paths")
@@ -688,7 +687,6 @@ def migrate_legacy_data_directories():
         env_manager.config.openrag_flows_path = f"{home}/.openrag/flows"
         env_manager.config.openrag_config_path = f"{home}/.openrag/config"
         env_manager.config.openrag_data_path = f"{home}/.openrag/data"
-        env_manager.config.opensearch_data_path = f"{home}/.openrag/data/opensearch-data"
         env_manager.config.langflow_data_path = f"{home}/.openrag/data/langflow-data"
         env_manager.save_env_file()
         print("  Updated .env with centralized paths")
--- a/src/tui/managers/container_manager.py
+++ b/src/tui/managers/container_manager.py
@@ -1278,23 +1278,41 @@ class ContainerManager:
 
         yield False, "Clearing OpenSearch data volume..."
 
-        # Get opensearch data path from env config
-        from .env_manager import EnvManager
-        env_manager = EnvManager()
-        env_manager.load_existing_env()
-        opensearch_data_path = Path(env_manager.config.opensearch_data_path.replace("$HOME", str(Path.home()))).expanduser().absolute()
+        # The volume name is 'opensearch-data' as defined in docker-compose.yml.
+        # Docker Compose usually prefixes it with the project name (defaulting to the directory name).
+        volume_name = "opensearch-data"
+        
+        # Try both the plain name and common prefixed versions
+        possible_names = [volume_name]
+        
+        # Current folder name is a common prefix
+        project_name = Path.cwd().name.lower().replace("-", "").replace("_", "")
+        possible_names.append(f"{project_name}_{volume_name}")
+        
+        # Another common prefix pattern
+        project_name_simple = Path.cwd().name.lower()
+        possible_names.append(f"{project_name_simple}_{volume_name}")
 
-        if not opensearch_data_path.exists():
-            yield True, "OpenSearch data directory does not exist, skipping"
-            return
+        last_error = "Unknown error"
+        for name in possible_names:
+            # Use alpine with root to clear container-owned files
+            cmd = [
+                "run", "--rm",
+                "-v", f"{name}:/work:Z",
+                "alpine",
+                "sh", "-c",
+                "rm -rf /work/* /work/.[!.]* 2>/dev/null; echo done"
+            ]
+            
+            success, stdout, stderr = await self._run_runtime_command(cmd)
+            if success and "done" in stdout:
+                yield True, f"OpenSearch data volume '{name}' cleared successfully"
+                return
 
-        # Use alpine with root to clear container-owned files
-        success, msg = await self.clear_directory_with_container(opensearch_data_path)
+            if stderr:
+                last_error = stderr.strip()
 
-        if success:
-            yield True, "OpenSearch data cleared successfully"
-        else:
-            yield False, f"Failed to clear OpenSearch data: {msg}"
+        yield False, f"Failed to clear OpenSearch data volume: {last_error}"
 
     async def reset_services(self) -> AsyncIterator[tuple[bool, str]]:
         """Reset all services (stop, remove containers/volumes, clear data) and yield progress updates."""
--- a/src/tui/managers/container_manager.py
+++ b/src/tui/managers/container_manager.py
@@ -1282,37 +1282,33 @@ class ContainerManager:
         # Docker Compose usually prefixes it with the project name (defaulting to the directory name).
         volume_name = "opensearch-data"
         
-        # Try both the plain name and common prefixed versions
-        possible_names = [volume_name]
-        
-        # Current folder name is a common prefix
-        project_name = Path.cwd().name.lower().replace("-", "").replace("_", "")
-        possible_names.append(f"{project_name}_{volume_name}")
-        
-        # Another common prefix pattern
-        project_name_simple = Path.cwd().name.lower()
-        possible_names.append(f"{project_name_simple}_{volume_name}")
+        possible_names = list(dict.fromkeys([
+            volume_name,
+            f"{Path.cwd().name.lower()}_{volume_name}",              # e.g. openrag_opensearch-data
+            f"{Path.cwd().name.lower().replace('-', '')}_{volume_name}",  # e.g. openrag_opensearch-data (dashes stripped)
+        ]))
 
-        last_error = "Unknown error"
-        for name in possible_names:
-            # Use alpine with root to clear container-owned files
-            cmd = [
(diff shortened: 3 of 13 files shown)


kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
```diff
@@ -70,7 +70,6 @@ stringData:
     {{- if .Values.global.opensearch.password }}
     OPENSEARCH_PASSWORD={{ .Values.global.opensearch.password | quote }}
     {{- end }}
-    OPENSEARCH_DATA_PATH=./opensearch-data
```
rodageve, line 73: We want this to be configurable as an environment variable setting for cloud deployments

zzzming (changes requested):
Please do not remove existing env variable. We have other provisioning components depending on it.

src/tui/managers/env_manager.py
```diff
@@ -97,7 +97,6 @@ class EnvConfig:
     openrag_flows_path: str = "$HOME/.openrag/flows"
     openrag_config_path: str = "$HOME/.openrag/config"
     openrag_data_path: str = "$HOME/.openrag/data"  # Backend data (conversations, tokens, etc.)
-    opensearch_data_path: str = "$HOME/.openrag/data/opensearch-data"
```
mpawlow, line 100: ## (a) [Critical] TUI Factory Reset Crashes with `AttributeError` on Removed Attribute

### Summary

`src/tui/screens/monitor.py` directly accesses `env_manager.config.opensearch_data_path`, which was deleted from `EnvConfig` in this PR. Any TUI-triggered factory reset will immediately crash with an `AttributeError`.

### Problem

- The PR deletes `opensearch_data_path` from `EnvConfig` (in `src/tui/managers/env_manager.py`) and removes it from the `_env_attr_map`, `save_env_file`, and the `ConfigField` definitions
- However, `src/tui/screens/monitor.py` in the `_do_factory_reset` generator was **not updated** and still reads:
  ```python
  opensearch_data_path = Path(env_manager.config.opensearch_data_path.replace("$HOME", str(Path.home()))).expanduser()
  ```
- At runtime, `env_manager.config.opensearch_data_path` raises `AttributeError: 'EnvConfig' object has no attribute 'opensearch_data_path'`
- This happens before `clear_opensearch_data_volume()` is even called, so the new Docker-volume-based clearing logic is completely unreachable
- The `reset_services()` call on line 544 already runs `docker compose down --volumes --remove-orphans`, which removes the named volume — the subsequent block in `monitor.py` was meant to clear additional directory-based data. With the `AttributeError`, factory reset in the TUI is fully broken

### Code References

- `src/tui/screens/monitor.py` — lines 550–568 (entire block referencing `opensearch_data_path`)

### Potential Solution

Remove the entire `opensearch-data` block from `_do_factory_reset` in `monitor.py` (lines 554–569). Since `reset_services()` already calls `docker compose down --volumes`, the named volume is already removed before this code is reached. The block is now both wrong and redundant:

```python
# DELETE these lines (554–569):
# Clear opensearch-data using container
yield False, "Clearing OpenSearch data..."
opensearch_data_path = Path(env_manager.config.opensearch_data_path.replace("$HOME", str(Path.home()))).expanduser()
if opensearch_data_path.exists():
    async for success, message in self.container_manager.clear_opensearch_data_volume():
        yield success, message
        if not success and "failed" in message.lower():
            return

    # Recreate empty opensearch-data directory
    try:
        opensearch_data_path.mkdir(parents=True, exist_ok=True)
        yield True, "OpenSearch data directory recreated"
    except Exception as e:
        yield False, f"Error recreating opensearch-data directory: {e}"
        return
```

src/tui/managers/container_manager.py
```diff
@@ -1278,23 +1278,41 @@ async def clear_opensearch_data_volume(self) -> AsyncIterator[tuple[bool, str]]:
...
+        possible_names.append(f"{project_name}_{volume_name}")
+        
+        # Another common prefix pattern
+        project_name_simple = Path.cwd().name.lower()
+        possible_names.append(f"{project_name_simple}_{volume_name}")
+
+        last_error = "Unknown error"
+        for name in possible_names:
+            # Use alpine with root to clear container-owned files
+            cmd = [
```
mpawlow: ## (b) [Major] Volume-Clearing Logic Always Reports False Success for Wrong Volume Names

### Summary

The volume-name-guessing loop in `clear_opensearch_data_volume()` will always report success on the first candidate (`opensearch-data`) regardless of whether it is the real volume, because Docker auto-creates volumes when they are mounted. The actual data volume (e.g., `openrag_opensearch-data`) is never cleared.

### Problem

- Docker's behaviour: when you run `docker run -v <name>:/work alpine ...` with a volume name that does **not** exist, Docker silently creates a new empty volume
- The shell command `rm -rf /work/* /work/.[!.]* 2>/dev/null; echo done` always exits 0 and prints `"done"` — even on an empty, newly-created volume
- Therefore `success and "done" in stdout` is **always true** for the first candidate `opensearch-data`, regardless of whether that volume is the real one
- The real volume produced by Docker Compose (e.g., `openrag_opensearch-data`) is never reached or cleared
- As a side-effect, every call to `clear-os-data` creates a spurious empty Docker volume named `opensearch-data`

```python
# This check will always pass for the first candidate, even if the volume is brand-new and empty:
if success and "done" in stdout:
    yield True, f"OpenSearch data volume '{name}' cleared successfully"
    return
```

### Potential Solution

Query Docker for the list of existing volumes **before** attempting to clear any of them, so only a volume that actually exists is targeted:

```python
# List volumes and find the correct one
list_success, stdout, _ = await self._run_runtime_command(["volume", "ls", "--format", "{{.Name}}"])
if not list_success:
    yield False, "Could not list Docker volumes"
    return

existing_volumes = set(stdout.splitlines())
found_name = next((n for n in possible_names if n in existing_volumes), None)

if found_name is None:
    yield True, "OpenSearch data volume not found — already removed or never created"
    return

# Only clear the volume we know exists
cmd = ["run", "--rm", "-v", f"{found_name}:/work:Z", "alpine", "sh", "-c",
       "rm -rf /work/* /work/.[!.]* 2>/dev/null; echo done"]
success, stdout, stderr = await self._run_runtime_command(cmd)
if success and "done" in stdout:
    yield True, f"OpenSearch data volume '{found_name}' cleared successfully"
else:
    yield False, f"Failed to clear OpenSearch data volume '{found_name}': {stderr}"
```

rodageve (approved):

zzzming (approved):

src/tui/main.py
mpawlow: ## (c) [Major] `~/.openrag/data/opensearch-data/` Directory Still Created Despite Docker-Volume Migration

### Confidence

80%

### Summary

`src/tui/main.py` still creates `~/.openrag/data/opensearch-data/` as a host directory on startup, even though OpenSearch data is now managed entirely by a Docker named volume. This is misleading and may confuse users who inspect their home directory.

### Problem

- The `_create_openrag_directories()` function (called at TUI startup) explicitly lists and creates `~/.openrag/data/opensearch-data/`:
  ```python
  base_dir / "data" / "opensearch-data",
  ```
- The docstring comment on line 756 still documents this path as used "for OpenSearch index"
- Since OpenSearch now uses a Docker named volume (`opensearch-data` in `docker-compose.yml`), Docker does **not** mount this host directory into the container — the created directory is never used
- Users may be confused when they see a perpetually empty `~/.openrag/data/opensearch-data/` on their filesystem

### Code References

- `src/tui/main.py` — lines 756, 766 (`_create_openrag_directories` body)

### Potential Solution

Remove `base_dir / "data" / "opensearch-data"` from the `directories` list and update the docstring:

```python
directories = [
    base_dir / "documents",
    base_dir / "flows",
    base_dir / "keys",
    base_dir / "config",
    base_dir / "data",
    # opensearch-data removed: now managed as a Docker named volume
]
```

And remove line 756 from the docstring accordingly.

src/tui/managers/container_manager.py
```diff
@@ -1278,23 +1278,41 @@ async def clear_opensearch_data_volume(self) -> AsyncIterator[tuple[bool, str]]:
...
-            return
+        # The volume name is 'opensearch-data' as defined in docker-compose.yml.
+        # Docker Compose usually prefixes it with the project name (defaulting to the directory name).
+        volume_name = "opensearch-data"
+        
+        # Try both the plain name and common prefixed versions
+        possible_names = [volume_name]
+        
+        # Current folder name is a common prefix
+        project_name = Path.cwd().name.lower().replace("-", "").replace("_", "")
```
mpawlow: ## (d) [Minor] Duplicate Volume Name Candidates in `possible_names`

### Summary

When the current working directory name contains no hyphens or underscores (e.g., the repo is checked out as `openrag`), `project_name` and `project_name_simple` are identical, so `possible_names` contains the same prefixed volume name twice. This is harmless but wasteful and indicates incomplete reasoning about the prefix-generation logic.

### Problem

- `project_name` strips hyphens and underscores: `Path.cwd().name.lower().replace("-", "").replace("_", "")`
- `project_name_simple` does not strip them: `Path.cwd().name.lower()`
- For a directory named `openrag` (no hyphens or underscores), both produce `"openrag"`, so `possible_names` becomes:
  ```python
  ["opensearch-data", "openrag_opensearch-data", "openrag_opensearch-data"]
  ```
- The third entry is redundant; the code attempts the same Docker command twice

### Code References

- `src/tui/managers/container_manager.py` — lines 1288–1294

### Potential Solution

Use a single strategy for constructing the prefix, or deduplicate the list:

```python
possible_names = list(dict.fromkeys([
    volume_name,
    f"{Path.cwd().name.lower()}_{volume_name}",              # e.g. openrag_opensearch-data
    f"{Path.cwd().name.lower().replace('-', '')}_{volume_name}",  # e.g. openrag_opensearch-data (dashes stripped)
]))
```

Using `dict.fromkeys` preserves insertion order while eliminating duplicates.

### Alternative Solutions

1. Use `docker volume ls` (as suggested in issue **b**) to find the real volume name programmatically, eliminating the guessing loop entirely.

src/tui/main.py
mpawlow: ## (e) [Minor] Legacy Migration Still Copies `opensearch-data` to a Path That Is No Longer Mounted

- Note: This is Minor. Please feel free to ignore or optionally implemented

### Summary

`migrate_legacy_data_directories()` in `src/tui/main.py` still copies an old `./opensearch-data` directory to `~/.openrag/data/opensearch-data` during a one-time migration, but since OpenSearch now uses a Docker named volume, the migrated data is never mounted and users' old index data is silently discarded.

### Problem

- The migration entry at line 587:
  ```python
  (cwd / "opensearch-data", target_base / "data" / "opensearch-data", "OpenSearch data"),
  ```
  copies data that will never be used by the Docker-volume-based OpenSearch
- An old index at `./opensearch-data` is binary data tied to the exact OpenSearch version and cluster UUID — even if it were mounted, reusing it would likely require a full re-import
- Users migrating from very old versions are told their "OpenSearch data" has been migrated, giving false confidence that their indices are preserved

### Code References

- `src/tui/main.py` — line 565 (docstring), line 587 (migration tuple)

### Potential Solution

Remove the `opensearch-data` entry from the migrations list and add an explanatory print statement:

```python
migrations = [
    (cwd / "openrag-documents", target_base / "documents", "documents"),
    (cwd / "flows", target_base / "flows", "flows"),
    (cwd / "keys", target_base / "keys", "keys"),
    (cwd / "config", target_base / "config", "config"),
    # opensearch-data intentionally omitted: OpenSearch now uses a Docker named
    # volume; old index data is not portable and must be re-ingested.
    (cwd / "langflow-data", target_base / "data" / "langflow-data", "Langflow data"),
]
```

tests/conftest.py
```diff
@@ -34,24 +34,20 @@ async def onboard_system():
...
+    # Clean up OpenSearch indices to ensure fresh state for tests
+    try:
+        await clients.initialize()
+        await clients.opensearch.indices.delete(index="_all", ignore_unavailable=True)
+        print("[DEBUG] Wiped all OpenSearch indices via API")
+    except Exception as e:
+        print(f"[DEBUG] Could not wipe OpenSearch indices: {e}")
 
-    # Initialize clients
+    # Initialize clients again (just to be safe, though already done)
```
mpawlow: ## (f) [Minor] Redundant Double `clients.initialize()` in `conftest.py`

### Summary

`tests/conftest.py` calls `await clients.initialize()` twice in the same fixture — once before wiping OpenSearch indices and again immediately after — with a comment acknowledging the redundancy.

### Problem

- Line 44: `await clients.initialize()` (needed to get a client for `indices.delete`)
- Line 51: `await clients.initialize()` (comment reads: "just to be safe, though already done")
- The second call is genuinely unnecessary; `initialize()` was already called and no state changes between the two calls require re-initialization
- This adds latency to the test setup without benefit and signals incomplete cleanup of the refactor

### Potential Solution

Remove the second call and its comment:

```python
# Clean up OpenSearch indices to ensure fresh state for tests
try:
    await clients.initialize()
    await clients.opensearch.indices.delete(index="_all", ignore_unavailable=True)
    print("[DEBUG] Wiped all OpenSearch indices via API")
except Exception as e:
    print(f"[DEBUG] Could not wipe OpenSearch indices: {e}")

# Initialize clients (already done above, but re-run in case of exception path)
# NOTE: remove the line below — clients are already initialized
await clients.initialize()   # <-- DELETE THIS LINE
```

mpawlow (changes requested):
@lucaseduoli 

## Code Review 1

- See PR comments: (a) to (f)

mpawlow (approved):
@lucaseduoli 

## Code Review 2

- ✅  Approved / LGTM

PR #1366, opened 2026-04-09
feat: upgraded Langflow version to 1.9.0, upgraded Docling version

lucaseduoli:
This pull request updates Langflow version to 1.9.0, and updates the `DoclingManager` process management in the TUI to improve reliability and simplify output handling. The most significant change is redirecting the subprocess output to a log file instead of using pipes, which prevents crashes when the parent process exits unexpectedly. The code also upgrades dependencies and streamlines output capture by tailing the log file in a single thread.

**Process management and output handling improvements:**
- Redirected subprocess `stdout` and `stderr` to a log file (`docling-serve.log`) in the TUI directory, instead of using pipes. This prevents SIGPIPE crashes and loss of in-memory state when the parent process exits. [[1]](diffhunk://#diff-2637d8507ee72e8a499f1990d03d9ef87bd3971d03a643d43ed79a945f80c3f1L42-R47) [[2]](diffhunk://#diff-2637d8507ee72e8a499f1990d03d9ef87bd3971d03a643d43ed79a945f80c3f1L318-R334)
- Updated the output capture mechanism to use a single thread that tails the log file, replacing the previous dual-threaded approach for capturing `stdout` and `stderr` pipes.
- Adjusted final process output reading to fetch from the log file, since pipes are no longer used.

**Dependency updates:**
- Upgraded `docling-serve[ui]` from version `1.5.0` to `1.15.1` and `docling-core` from `2.48.1` to `2.71.0` in the subprocess command.

Duplicate of #1355


edwinjosechittilappilly (approved):
LGTM

Issue #1369, opened 2026-04-09
[Bug]: Security: Containerized Langflow application is running as root in Cloud / SaaS


mpawlow:
### OpenRAG Version

0.4.1

### Deployment Method

uv add (installed in project)

### Operating System

Ubuntu 24.04

### Python Version

Python 3.13.12

### Affected Area

Cloud / SaaS

### Bug Description

Security: Containerized Langflow application is running as root in Cloud / SaaS

### Steps to Reproduce

N/A

### Expected Behavior

Containerized Langflow application should be running as a non-root user

### Actual Behavior

Containerized Langflow application is running as root

### Relevant Logs

```shell
N/A
```

### Screenshots

N/A

#### Related Issues

- https://github.com/langflow-ai/openrag/issues/1320
- https://github.com/langflow-ai/openrag/issues/1322

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

mpawlow:
CC @rodageve

mpawlow:
Fixed.

PR #1372, opened 2026-04-10
fix: Security: Containerized Langflow application is running as root in Cloud / SaaS

mpawlow:
### Issue

- #1369

### Summary

- Removes the `USER root` escalation from `Dockerfile.langflow` and eliminates the associated custom `langflow-entrypoint.py` script that ran as root to fix bind-mount permissions before dropping privileges
- The Langflow container now runs exclusively as the non-root user (`uid=1000`) inherited from the base image throughout its entire lifecycle, eliminating the security risk of running as root in Cloud / SaaS deployments

### Security Fix: Remove Root User Execution from Langflow Container

- **Root cause**: `Dockerfile.langflow` previously issued a `USER root` instruction after copying the entrypoint script, causing the container process to start (and remain) as root in Cloud / SaaS environments
- **Previous workaround**: A custom `scripts/langflow-entrypoint.py` was introduced to:
  - Run as root at startup to `chmod 0o777` the `/app/langflow-data` bind-mount directory
  - Drop privileges to `uid=1000` / `gid=1000` (the `langflow` user) before exec-ing the main process
  - Restore `HOME` and `USER` environment variables after the privilege drop
  - This workaround was originally needed for macOS/Podman environments where the `virtiofs` layer does not faithfully propagate host-side `chmod` changes into the container
- **Fix applied**:
  - Removed the `USER root` instruction from `Dockerfile.langflow`
  - Removed the `COPY --chmod=755 scripts/langflow-entrypoint.py /usr/local/bin/langflow-entrypoint` layer
  - Removed the `ENTRYPOINT ["langflow-entrypoint"]` override
  - Deleted `scripts/langflow-entrypoint.py` entirely (43 lines removed)
  - The container now uses the default `CMD ["langflow", "run", "--host", "0.0.0.0", "--port", "7860"]` with no custom entrypoint


lucaseduoli (approved):
LGTM

Issue #1320, opened 2026-04-02
[Bug]: Security: Containerized Frontend application is running as root


mpawlow:
### OpenRAG Version

0.4.1

### Deployment Method

uv add (installed in project)

### Operating System

Ubuntu 24.04

### Python Version

Python 3.13.12

### Affected Area

Frontend (Next.js UI, components)

### Bug Description

Security: Containerized Frontend application is running as root 

### Steps to Reproduce

N/A

### Expected Behavior

Containerized Frontend application should be running as a non-root user

### Actual Behavior

Security: Containerized Frontend application is running as root 

### Relevant Logs

```shell
N/A
```

### Screenshots

N/A

### Additional Context

N/A

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

mpawlow:
CC @rodageve

PR #1373, opened 2026-04-10
fix: Security: Containerized Frontend application is running as root

mpawlow:
### Issue

- #1320

### Reference Pull Request

- https://github.com/langflow-ai/openrag/pull/1321
- ✅  Changes already reviewed, approved, tested and merged into **main**
- ✅  Commits were cherry-picked from **main**
   - ✅  Applied cleanly (no merge conflicts)  

### Summary

- Fixed the frontend Docker container running as root by switching to the built-in non-root `node` user for all build and runtime steps.

### Security: Run frontend container as non-root user

- Added `chown node:node /app` after `WORKDIR` so the working directory is owned by the non-root user before the user switch.
- Added `USER node` directive to switch to the non-root `node` user for all subsequent Dockerfile instructions.
- Updated `COPY` instructions to use `--chown=node:node` so copied files are owned by the `node` user rather than root.
- Removed trailing whitespace from `npm run build` line.
- Added missing newline at end of file.

### DIFF
--- a/Dockerfile.frontend
+++ b/Dockerfile.frontend
@@ -1,17 +1,21 @@
 FROM node:20.20.0-slim
 
-# Set working directory
+# Set working directory and ensure it is owned by the non-root node user
 WORKDIR /app
+RUN chown node:node /app
+
+# Switch to non-root user for all subsequent steps
+USER node
 
 # Copy frontend dependencies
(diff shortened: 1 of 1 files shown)


lucaseduoli (approved):
LGTM

Issue #1322, opened 2026-04-02
[Bug]: Security: Containerized Backend application is running as root


mpawlow:
### OpenRAG Version

0.4.1

### Deployment Method

uv add (installed in project)

### Operating System

Ubuntu 24.04

### Python Version

Python 3.13.12

### Affected Area

Frontend (Next.js UI, components)

### Bug Description

Security: Containerized Backend application is running as root 

### Steps to Reproduce

N/A

### Expected Behavior

Containerized Backend application should be running as a non-root user

### Actual Behavior

Security: Containerized Backend application is running as root 

### Relevant Logs

```shell
N/A
```

### Screenshots

N/A

#### Related Issues

- https://github.com/langflow-ai/openrag/issues/1320

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

PR #1374, opened 2026-04-10
fix(security): Containerized Backend application is running as root

### LINKED ISSUE #1384: [Feature]: Re-implement support for the JWT_SIGNING_KEY environment variable for auth as a replacement for generation and use of private / public keys
### Feature Area

Backend/API (Python/FastAPI)

### Problem Description

- In Cloud / SaaS, public / private keys are generated and need to be stored in the containers file system / volume mount (`/keys`)
- In Cloud / SaaS, authentication is already handled externally outside of the **OpenRAG Backend**

### Proposed Solution

- Fix and re-implement support for the JWT_SIGNING_KEY environment variable for auth
- When a signing key is provided, bypass the generation and use of the public / private keys for auth

### Use Case

Cloud / SaaS

### Alternatives Considered

_No response_

### Priority

Critical for my use case

### Additional Context

_No response_

### Contribution

### Checklist

mpawlow:
### Issue

- #1322

### Reference Pull Request

- https://github.com/langflow-ai/openrag/pull/1329
- ✅  Changes already reviewed, approved, tested and merged into **main**
- ✅  Commits were cherry-picked from **main**
   - ✅  Applied cleanly (no merge conflicts)  

### Summary

- The backend container previously ran as `root`, violating the principle of least privilege and failing common container security benchmarks (CIS, NSA hardening guides).
- This fix introduces a dedicated non-root user (`appuser`, UID/GID 1000) in the backend image, a new `entrypoint.sh` that drops privileges at container startup via `gosu`, and supporting changes to the `Makefile` and TUI to ensure host-side volume directories are pre-created with the correct ownership before the container starts.
- A belt-and-suspenders approach is used to handle both Docker (which does not remap UIDs on volume mounts) and Podman (which uses `:U` in `docker-compose.yml`): the entrypoint re-`chown`s mounted directories at startup; the `Makefile` and TUI independently pre-set ownership before `docker compose up`.
- Added support for an external `JWT_SIGNING_KEY` environment variable, allowing the backend to skip RSA key generation when a signing key is already provided externally.
- Added `host.docker.internal` host mapping to the Langflow service in `docker-compose.yml` to enable container-to-host networking.

### Dockerfile Changes (`Dockerfile.backend`)

- Installed `gosu` via `apt-get` to enable clean privilege dropping from `root` to `appuser` at container startup without a setuid shell.
- Created the `appuser` group (GID 1000) and user (UID 1000) using `groupadd`/`useradd`:
  - UID/GID 1000 is the conventional first non-root account and matches what Podman's `:U` volume flag maps to.
- Pre-created all runtime-writable directories (`keys/`, `data/`, `config/`, `flows/backup/`, `openrag-documents/`) inside the image and `chown`-ed the entire `/app` tree to `appuser`:
  - Ensures safe defaults when no host volume is mounted (e.g. CI runs, unit tests).
- Copied `entrypoint.sh` into the image and set it as the `ENTRYPOINT`:
  - `CMD ["python", "src/main.py"]` remains unchanged, so the entrypoint receives the original command as `"$@"`.
- Added `chmod +x /entrypoint.sh` during the build step.

### New File: `entrypoint.sh`

- Runs as `root` on container startup (before application code executes).
- Unconditionally re-`chown`s all volume-mounted directories to `appuser:appuser`:
  - `/app/keys`, `/app/flows`, `/app/config`, `/app/data`, `/app/openrag-documents`
  - The `chown` is a no-op when the directory is already owned by `appuser` (e.g. a fresh Podman start), so there is no runtime overhead in the common case.
  - Errors are suppressed (`2>/dev/null || true`) to avoid failing startup when an optional volume is not mounted.
- Calls `exec gosu appuser "$@"` to replace the root shell with the application process running as `appuser`, ensuring the final PID 1 is the application, not a wrapper shell.

### Makefile Changes

- Added `HOST_UID` and `HOST_GID` variables (evaluated once at parse time via `id -u` / `id -g`) for use in Docker-assisted `chown` commands.
- Added new `ensure-backend-volumes` target:
  - Creates `flows/`, `keys/`, `config/`, and `data/` directories on the host if they do not exist.
  - Re-owns them to the host user via an Alpine container running as root (mirrors what the backend entrypoint does inside the container, but in reverse — host-side).
  - Falls back to native `chown`/`chmod` if no container runtime is available.
  - Sets `chmod 775` on all four directories so both the host user and container's `appuser` (when GID matches) can write.
- Added `ensure-backend-volumes` as a prerequisite to the following targets:
  - `dev`, `dev-cpu`, `dev-local`, `dev-local-cpu`
  - `dev-local-build-lf`, `dev-local-build-lf-cpu`
  - `dev-branch`, `dev-branch-cpu`, `restart-dev`
  - `backend` (local development)
  - `test-ci`, `test-ci-local`
- Updated `factory-reset` to use Docker Alpine for JWT key removal when the host user does not own the key files (i.e. after the container has `chown`-ed them to `appuser`).
- Removed manual JWT key generation and permission-fixing steps from `test-ci` and `test-ci-local`:
  - Key generation is now handled by the container at startup (via `src/main.py` → `generate_jwt_keys`).
  - Added a post-startup step in both CI targets to reclaim key ownership back to the host user (via Alpine) so the test runner can read `private_key.pem`.

### TUI Changes (`src/tui/main.py`)

- Extended `setup_host_directories()` to apply `chmod 0o775` on all backend volume-mounted directories (`documents/`, `flows/`, `keys/`, `config/`, `data/`) before starting Docker services:
  - Ensures `appuser` (UID 1000) can write to these directories on Docker hosts that do not perform UID remapping.
- Added `_reclaim_host_ownership(directories)` helper function:
  - Detects directories whose host-side owner UID differs from the current process UID (i.e. the container previously `chown`-ed them to UID 1000).
  - Launches an Alpine container as root to `chown` each affected directory back to the host user — mirrors the `ensure-backend-volumes` Makefile logic.
  - Silently skips directories already owned by the current user or when no container runtime (`docker` or `podman`) is available.
  - Called from `setup_host_directories()` before the `chmod` loop, so subsequent `chmod` calls succeed even after a prior container run changed ownership.
- Wrapped `os.chmod(langflow_data_dir, 0o777)` in a `try/except PermissionError` to prevent a crash when the TUI is not the directory owner (consistent with the new handling for backend directories).

### CI Workflow Changes (`.github/workflows/`)

- Updated the "Cleanup root-owned files" step in both `test-e2e.yml` and `test-integration.yml` to also remove `keys/`, `data/`, `flows/`, and `openrag-documents/` directories:
  - These directories are now created and `chown`-ed by the backend container, so they must be explicitly removed by the Alpine cleanup step (which runs as root) to avoid stale ownership blocking subsequent CI runs.

### Backend

- Skipped RSA JWT key generation in `initialize_services()` when the `JWT_SIGNING_KEY` environment variable is set.
- Removed `os.chmod()` calls on existing RSA key files to avoid permission errors when running as a non-root user.
- Introduce OPENRAG_RUN_MODE environment variable
- Execute gosu only if running as root
- Rename Backend entry point script

### Docker

- Passed `JWT_SIGNING_KEY` environment variable through to the backend service in `docker-compose.yml`.
- Added `extra_hosts` entry (`host.docker.internal:host-gateway`) to the Langflow service for host networking support.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -233,3 +233,10 @@ SERVICE_NAME=openrag
 # Secret key for session management (auto-generated if not provided)
 SESSION_SECRET=
 
+# OPTIONAL: The OpenRAG application execution mode or deployment target.
+# - Used to configure the operation of the application based on it's execution context.
+# Values:
+# - "oss": Open Source Software (default)
+# - "saas": Public Cloud
+# - "on_prem": Private Cloud
+OPENRAG_RUN_MODE=oss
--- a/.env.example
+++ b/.env.example
@@ -233,6 +233,25 @@ SERVICE_NAME=openrag
 # Secret key for session management (auto-generated if not provided)
 SESSION_SECRET=
 
+# OPTIONAL: JWT signing key for token issuance and verification.
+# If not set, OpenRAG generates an RSA key pair (private_key.pem / public_key.pem)
+# under OPENRAG_KEYS_PATH at startup and uses RS256 signing.
+# Accepted values:
+#   - RSA private key in PEM format (PKCS#8 or PKCS#1) → RS256
+#   - Plain string → HS256 (symmetric; JWKS endpoint returns an empty key set)
+# When set, the on-disk key files are never read or created.
+#
+# PEM keys must be set as a double-quoted single-line value with literal \n
+# separators. python-dotenv expands \n in double-quoted strings to real newlines,
+# which is required by the PEM parser. Multiline values break Make's .env parsing,
+# and single-quoted or unquoted values do not expand \n.
+#
+# Generate and format a test key with:
+#   openssl genpkey -algorithm RSA -pkeyopt rsa_keygen_bits:2048 2>/dev/null \
+#     | awk '{printf "%s\\n", $0}' | sed 's/^/"/' | sed 's/$/"/'
+#
+# JWT_SIGNING_KEY=
+
 # OPTIONAL: The OpenRAG application execution mode or deployment target.
 # - Used to configure the operation of the application based on it's execution context.
 # Values:
--- a/src/main.py
+++ b/src/main.py
@@ -375,13 +375,7 @@ def generate_jwt_keys():
             )
             raise
     else:
-        # Ensure correct permissions on existing keys
-        try:
-            os.chmod(private_key_path, 0o600)
-            os.chmod(public_key_path, 0o644)
-            logger.info("RSA keys already exist, ensured correct permissions")
-        except OSError as e:
-            logger.error("Failed to set permissions on existing keys", error=str(e))
+        logger.info("RSA keys already exist")
 
 
 def _get_documents_dir():
@@ -1368,8 +1362,9 @@ async def initialize_services():
     await TelemetryClient.send_event(
         Category.SERVICE_INITIALIZATION, MessageId.ORB_SVC_INIT_START
     )
-    # Generate JWT keys if they don't exist
-    generate_jwt_keys()
+    # Generate JWT keys if they don't exist and a JWT signing key isn't specified
+    if not os.getenv("JWT_SIGNING_KEY"):
+        generate_jwt_keys()
 
     from config.settings import IBM_AUTH_ENABLED
 
@@ -1480,7 +1475,7 @@ async def create_app():
     app = FastAPI(title="OpenRAG API", version=OPENRAG_VERSION, debug=True)
     app.state.services = services  # Store services for cleanup
     app.state.background_tasks = set()
-    
+
     try:
         Instrumentator().instrument(app).expose(app)
     except Exception as e:
@@ -2040,14 +2035,14 @@ async def create_app():
             Category.APPLICATION_SHUTDOWN, MessageId.ORB_APP_SHUTDOWN
         )
         logger.info("Application shutdown initiated")
-        
+
         # Gracefully shutdown OpenSearch connection first
         try:
             from utils.opensearch_utils import graceful_opensearch_shutdown
             await graceful_opensearch_shutdown(clients.opensearch)
         except Exception as e:
             logger.error("Error during graceful OpenSearch shutdown", error=str(e))
-        
+
         await cleanup_subscriptions_proper(services)
         # Cleanup task service (cancels background tasks and process pool)
         await services["task_service"].shutdown()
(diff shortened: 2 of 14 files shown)


src/main.py
```diff
@@ -1368,8 +1362,9 @@ async def initialize_services():
     await TelemetryClient.send_event(
         Category.SERVICE_INITIALIZATION, MessageId.ORB_SVC_INIT_START
     )
-    # Generate JWT keys if they don't exist
-    generate_jwt_keys()
+    # Generate JWT keys if they don't exist and a JWT signing key isn't specified
+    if not os.getenv("JWT_SIGNING_KEY"):
```
edwinjosechittilappilly, line 1366: we dont use this key for SAAS I believe ? if not used you can put as not IBM AUTH MODE enabled flag.

.env.example
```diff
@@ -233,3 +233,29 @@ SERVICE_NAME=openrag
...
+#
+# JWT_SIGNING_KEY=
+
+# OPTIONAL: The OpenRAG application execution mode or deployment target.
+# - Used to configure the operation of the application based on it's execution context.
+# Values:
+# - "oss": Open Source Software (default)
+# - "saas": Public Cloud
+# - "on_prem": Private Cloud
+OPENRAG_RUN_MODE=oss
```
edwinjosechittilappilly, line 261: Do we need a new FLAG?
I like this IDEA of run mode.

also Can we pair it with IBM AUTH MODE enabled Flag also so it works with current flag for SaaS.
connect with @rodageve  for this new variable update also

src/main.py
mpawlow: - That is correct, the key is not used for **Cloud / SaaS** (the AMS auth layer is what is utilized in this context)
- The key is simply provided as a tactical work-around to generating the RSA private / public keys and storing it in the volume mounts (which was causing problems and preventing the **Backend** application from starting
- The next step is to bypass the RSA private / public keys generation + load path via a feature flag instead
- Using the existing `IBM_AUTH_ENABLED` env var makes sense

.env.example
mpawlow: - The new `OPENRAG_RUN_MODE` env var is not needed / used in this PR
- It was introduced as new infrastructure for future use (didn't want the new Python utility to get lost)
- However, I can remove it temporarily for cleanliness

src/main.py
mpawlow: Tracked By
- https://github.com/langflow-ai/openrag/issues/1393

edwinjosechittilappilly (approved):
LGTM

PR #1385, opened 2026-04-12
feat(backend): Re-implement support for the JWT_SIGNING_KEY environment variable for auth...

### LINKED ISSUE #1384: [Feature]: Re-implement support for the JWT_SIGNING_KEY environment variable for auth as a replacement for generation and use of private / public keys

mpawlow:
...as a replacement for generation and use of private / public keys

### Issue

- #1384
- #1322

### Summary

- Re-implements support for the `JWT_SIGNING_KEY` environment variable as an alternative to RSA key pair generation, allowing external JWT signing key injection (e.g., from a Kubernetes secret or platform secret store)
- Fixes a security issue where the containerized backend ran as root by hardening the entrypoint script, introducing conditional privilege drop via `gosu`, and adding an `OPENRAG_RUN_MODE` utility for deployment-context-aware behavior

### JWT Signing Key Support

- Skips RSA key generation in `initialize_services()` when the `JWT_SIGNING_KEY` environment variable is set, avoiding unnecessary key file creation
- Removed `os.chmod()` calls on existing RSA key files in `generate_jwt_keys()` to prevent permission errors when the container runs as a non-root user
- Passed `JWT_SIGNING_KEY` through to the backend service via `docker-compose.yml`
- Added `extra_hosts` entry (`host.docker.internal:host-gateway`) to the Langflow service in `docker-compose.yml` to enable container-to-host networking

### Security: Non-Root Container Hardening

- Renamed `entrypoint.sh` to `scripts/backend-entrypoint.sh` and updated `Dockerfile.backend` to reference the new path
- Updated `scripts/backend-entrypoint.sh` to conditionally execute `chown` and `gosu` only when the container is running as root (`id -u == 0`), allowing the same entrypoint to work correctly in both Docker (root-started) and Podman (rootless/UID-remapped) runtimes
- Added `src/utils/run_mode_utils.py` with `get_run_mode()`, `is_run_mode_oss()`, `is_run_mode_saas()`, and `is_run_mode_on_prem()` helpers that read the `OPENRAG_RUN_MODE` environment variable for deployment-context-aware logic
- Added `OPENRAG_RUN_MODE` to `.env.example` documenting the three supported values: `oss` (default), `saas`, and `on_prem`


edwinjosechittilappilly (approved):
lgtm

PR #1392, opened 2026-04-13
refactor(knn): consolidate knn_vector field mapping into shared helper

### LINKED ISSUE #1391: [Feature]: consolidate KNN vector method mapping into shared helper
## Problem

The same `knn_vector `method configuration literal is duplicated in three places in `src/`:
- `src/config/settings.py`: INDEX_BODY (legacy chunk_embedding field at initial index creation)
- `src/utils/embedding_fields.py`: ensure_embedding_field_exists (runtime per-model field creation via PUT _mapping)
- `src/utils/embeddings.py`: create_dynamic_index_body (alternate dynamic index creation path)

Each copy produces an identical block:
```
  {
      "type": "knn_vector",
      "dimension": <dims>,
      "method": {
          "name": "disk_ann",
          "engine": "jvector",
          "space_type": "l2",
          "parameters": {"ef_construction": KNN_EF_CONSTRUCTION, "m": KNN_M},
      },
  }
```

Any change to KNN method parameters requires editing three files in lockstep, which invites drift and increases defect surface area. 

## Proposed approach

Introduce a single helper — suggested location `src/utils/embedding_fields.py`, suggested name `build_knn_vector_field(dims: int) -> dict` which will returns the full field mapping dict. Update all three call sites to  import  and use it.

The helper should:
 - Accept `dims: int (required)`
 - Read `KNN_EF_CONSTRUCTION` and `KNN_M `from config.settings
 - Return the complete `knn_vecto`r field dict ready to splice into a mapping's properties
 - Leave room for future advanced.* parameters without changing the signature (e.g., accept an optional overrides dict or return a mutable dict that callers can extend  (there are several JVector optimisation opportunities)

### Acceptance criteria

- Single source of truth for the `knn_vector` field definition
- All three files above import and use the helper and no inline literal remains
- No change in runtime behavior, index mappings produced are byte-identical to current output for the same dims
- `KNN_EF_CONSTRUCTION` and `KNN_M` still resolve from config.settings (do not hardcode in the helper)
- Helper has a unit test that verifies the returned dict structure for at least two different dimension values

### Out of scope

- Changing any parameter values (m, ef_construction, hierarchy, PQ overrides) as these will impact existing installations
- Multi-model or per-dimension parameter logic (these will be addressed in follow up performance improvement issues)

zznate:
Extract the duplicated knn_vector field literal into build_knn_vector_field() in utils.embedding_fields as the single source of truth for OpenRAG's JVector/DiskANN method configuration. All four call sites now delegate to the helper:

- src/config/settings.py (INDEX_BODY legacy chunk_embedding field)
- src/utils/embedding_fields.py (ensure_embedding_field_exists)
- src/utils/embeddings.py (create_dynamic_index_body)
- scripts/migrate_embedding_model_field.py (ensure_new_field_exists)

The helper returns a fresh dict per call so callers may extend it in place without affecting other sites, which unblocks follow-up work on per-field advanced.* parameters (hierarchy_enabled, num_pq_subspaces, etc.) without re-introducing drift.

No runtime behavior change: the generated mapping is byte-identical to the previous inline literal for the same dimension. KNN_EF_CONSTRUCTION and KNN_M imports are removed from callers that no longer reference them directly.

Adds tests/unit/test_embedding_fields.py covering structure, dimension propagation, settings resolution (including monkeypatch override), call-site isolation, and INDEX_BODY parity.

Refs #1391


phact (approved):
LGTM pending conflict resolution

edwinjosechittilappilly:
should we make similar change in the Opensearch component also?

zznate:
@edwinjosechittilappilly yes - i'll tackle that under a separate issue. I wanted to sort this out first to drill into how we use opensearch and jvector as there are a few more efficiencies we could add.

PR #1404, opened 2026-04-15
feat: Add check before filter creation during onboarding

mfortman11:
Adds check for if a filter is created based on onboarding ID and by the `OpenRAG Docs` filter name to avoid duplicates

Addressing https://github.ibm.com/lakehouse/tracker/issues/69638

This pull request improves the onboarding process by adding safeguards to prevent creating duplicate "OpenRAG Docs" knowledge filters when sample data is ingested. The changes ensure that a new filter is only created if one does not already exist, even if the onboarding embedding step is resubmitted.

**Enhancements to duplicate filter prevention:**

* In `src/api/settings.py`, the onboarding logic now checks if an `openrag_docs_filter_id` already exists before attempting to create a new "OpenRAG Docs" filter, preventing duplicate filters when users re-submit the embedding step.
* The `_create_openrag_docs_filter` function now queries the knowledge filters index in OpenSearch to check for an existing filter named "OpenRAG Docs" before creating a new one; if found, it logs and returns the existing filter's ID, otherwise it proceeds with creation.


lucaseduoli (approved):
LGTM! tested and working

PR #1419, opened 2026-04-16
Fix: remove duplicate /templates prefix in helm checksum paths

Matroxt:
**Problem**

When running `helm template`, you'd get this error
```
Error: Error: template: templates/langflow/deployment.yaml:23:28: executing "templates/langflow/deployment.yaml"  
  at <include (print $.Template.BasePath "/templates/langflow/langflow-dotenv.yaml") .>: error calling include: template: no        
  template "templates/templates/langflow/langflow-dotenv.yaml" associated with template "gotpl"
```

**Reason**

The `$.Template.BasePath` variable already resolves to the `templates/` directory, so including `/templates/` in the path caused an incorrect checksum annotations for the backend and langflow deployments. 

**Fix**

This fix removes the redundant prefix so Helm can correctly locate and hash the dotenv files, ensuring pods are restarted when their configuration changes.
                                                                                                                                    
Also 
- Bumps the chart version to `0.2.1` because of the fix
- Adjust `appVersion` to `0.4.1` to reflect the current release.
- Remove duplicated `loglevel` key under the `backend` key for Helm values

### DIFF
--- a/kubernetes/helm/openrag/Chart.yaml
+++ b/kubernetes/helm/openrag/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: openrag
 description: A Helm chart for deploying OpenRAG - an open-source agentic RAG platform
 type: application
-version: 0.2.0
+version: 0.2.1
 appVersion: "0.1.52"
 keywords:
   - openrag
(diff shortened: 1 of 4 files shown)


Matroxt:
It got closed in https://github.com/langflow-ai/openrag/pull/1655, closing this PR.

PR #1425, opened 2026-04-16
fix: update models api and langflow models setting

lucaseduoli:
This pull request introduces significant improvements to the concurrency and robustness of the flows management system, primarily within the `flows_service.py` file. The key changes include parallelizing multiple flow operations using asyncio, adding per-flow locking to prevent race conditions, and improving error handling and logging. Additionally, the onboarding and model update logic has been streamlined, and test version parameters were updated for WatsonX API validation.

**Concurrency and Parallelization Improvements:**

* Refactored several flow operations (`backup_all_flows`, `ensure_flows_exist`, `check_flows_reset`, `change_langflow_model_value`) to run tasks concurrently using `asyncio.gather`, greatly improving performance when handling multiple flows. [[1]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL176-R194) [[2]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL822-R861) [[3]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL888-R942) [[4]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL936-L950)
* Introduced per-flow asyncio locks and lock management to prevent race conditions during concurrent updates to the same flow, especially in the `_update_provider_components` method. [[1]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dR81-R95) [[2]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL974-R1027)

**Robustness and Error Handling:**

* Enhanced error handling by collecting and logging exceptions for each flow operation, ensuring failures in one flow do not affect others and providing clearer diagnostics. [[1]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL936-L950) [[2]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL176-R194)

**Model Update and Onboarding Logic:**

* Updated the `_update_langflow_model_values` function to batch and run model updates concurrently, improving efficiency and logging a summary upon completion. [[1]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceR1487-L1496) [[2]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1517-R1529)
* Streamlined onboarding logic to only attempt to save configuration when necessary, and improved logging when saving fails. [[1]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceR1118-R1119) [[2]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1165-L1167)

**Testing and Configuration:**

* Updated WatsonX API test version parameters from `2024-09-16` to `2026-04-15` for both completion and embedding endpoints, ensuring tests use the latest supported API versions. [[1]](diffhunk://#diff-a50ca64f56c03024a1381b36e0d50f0e969973b08a80291d6632a02c695e325dL457-R457) [[2]](diffhunk://#diff-a50ca64f56c03024a1381b36e0d50f0e969973b08a80291d6632a02c695e325dL552-R552)

**Miscellaneous:**

* Simplified and clarified certain flow update logic, removed unnecessary comments, and improved code readability. [[1]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL1009-L1016) [[2]](diffhunk://#diff-a8d945bb6eab1d9d5fdbd2a0963a76d72a99cf6fdd3b3a36a16b66971899022dL1027-R1077)
* Removed custom OpenSearch container command from `docker-compose.yml`, likely to simplify container startup and rely on default entrypoint behavior.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,21 +10,6 @@ services:
     environment:
       - discovery.type=single-node
       - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}
-        # Run security setup in background after OpenSearch starts
-    command: >
-      bash -c "
-        # Ensure data directory has correct permissions
-        sudo chown -R opensearch:opensearch /usr/share/opensearch/data || true
-
-        # Start OpenSearch in background
-        /usr/share/opensearch/opensearch-docker-entrypoint.sh opensearch &
-
-        # Wait a bit for OpenSearch to start, then apply security config
-        sleep 10 && /usr/share/opensearch/setup-security.sh &
-
-        # Wait for background processes
-        wait
-      "
       - OPENSEARCH_JAVA_OPTS=-Xms1g -Xmx1g
     ports:
       - "9200:9200"
--- a/src/api/settings.py
+++ b/src/api/settings.py
@@ -1110,12 +1110,12 @@ async def onboarding(
                 current_config.providers.any_configured()):
                 await _update_langflow_global_variables(current_config, flows_service=flows_service)
 
+            if body.llm_provider or body.llm_model and not (body.embedding_provider or body.embedding_model):
+                await _update_langflow_model_values(current_config, flows_service, llm_model=body.llm_model, llm_provider=body.llm_provider)
+
             if body.embedding_provider or body.embedding_model:
                 await _update_mcp_servers_with_provider_credentials(current_config, session_manager=session_manager, flows_service=flows_service)
-
-            # Update model values if provider or model fields were provided
-            if body.llm_provider or body.llm_model or body.embedding_provider or body.embedding_model:
-                await _update_langflow_model_values(current_config, flows_service, body.llm_model, body.llm_provider, body.embedding_model, body.embedding_provider)
+                await _update_langflow_model_values(current_config, flows_service, embedding_model=body.embedding_model, embedding_provider=body.embedding_provider)
 
         except Exception as e:
             logger.error(
@@ -1479,57 +1479,58 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager 
         # Don't fail the entire settings update if MCP update fails
 
 
-async def _update_langflow_model_values(config, flows_service, llm_model, llm_provider, embedding_model, embedding_provider):
+async def _update_langflow_model_values(config, flows_service, llm_model=None, llm_provider=None, embedding_model=None, embedding_provider=None):
     """Update model values across Langflow flows for all configured providers"""
     try:
 
-        # 1. Update ONLY the current LLM provider
-        llm_provider = llm_provider.lower() if llm_provider else config.agent.llm_provider.lower()
-        llm_model = llm_model if llm_model else config.agent.llm_model
-        flows_service.change_langflow_model_value(
-            llm_provider,
-            llm_model=llm_model,
-            force_llm_update=True
-        )
-
-
-        if embedding_provider and embedding_model:
-            result = await flows_service.change_langflow_model_value(
-                embedding_provider,
-                embedding_model=embedding_model,
-                force_embedding_update=True
+        if not embedding_model and not embedding_provider:
+            # 1. Update ONLY the current LLM provider
+            current_llm_provider = llm_provider.lower() if llm_provider else config.agent.llm_provider.lower()
+            current_llm_model = llm_model if llm_model else config.agent.llm_model
+            await flows_service.change_langflow_model_value(
+                current_llm_provider,
+                llm_model=current_llm_model,
+                force_llm_update=True
             )
 
-            logger.info(
-                f"Successfully updated Langflow flows for embedding provider {embedding_provider}",
-                result=result
-            )
-        else: 
-            # 2. Update ALL configured embedding providers
-            embedding_providers = []
-            if config.providers.openai.configured:
-                embedding_providers.append("openai")
-            if config.providers.watsonx.configured:
-                embedding_providers.append("watsonx")
-            if config.providers.ollama.configured:
-                embedding_providers.append("ollama")
-
-            current_embedding_provider = config.knowledge.embedding_provider.lower()
-            for provider in embedding_providers:
-                # Use configured model for current provider, or None (first available) for others
-                embedding_model = (
-                    config.knowledge.embedding_model
-                    if provider == current_embedding_provider
-                    else None
+        if not llm_model and not llm_provider:
+            if embedding_provider and embedding_model:
+                result = await flows_service.change_langflow_model_value(
+                    embedding_provider,
+                    embedding_model=embedding_model,
+                    force_embedding_update=True
                 )
-                await flows_service.change_langflow_model_value(
-                        provider,
-                        embedding_model=embedding_model,
-                        force_embedding_update=True
-                    )
+
                 logger.info(
-                    f"Successfully updated Langflow flows for embedding provider {provider}"
+                    f"Successfully updated Langflow flows for embedding provider {embedding_provider}",
+                    result=result
                 )
+            else: 
+                # 2. Update ALL configured embedding providers
+                embedding_providers = []
+                if config.providers.openai.configured:
+                    embedding_providers.append("openai")
+                if config.providers.watsonx.configured:
+                    embedding_providers.append("watsonx")
+                if config.providers.ollama.configured:
+                    embedding_providers.append("ollama")
+
+                current_embedding_provider = config.knowledge.embedding_provider.lower()
+                for provider in embedding_providers:
+                    # Use configured model for current provider, or None (first available) for others
+                    embedding_model = (
+                        config.knowledge.embedding_model
+                        if provider == current_embedding_provider
+                        else None
+                    )
+                    await flows_service.change_langflow_model_value(
+                            provider,
+                            embedding_model=embedding_model,
+                            force_embedding_update=True
+                        )
+                    logger.info(
+                        f"Successfully updated Langflow flows for embedding provider {provider}"
+                    )
     except Exception as e:
         logger.error(f"Failed to update Langflow model values: {str(e)}")
         raise
--- a/src/api/settings.py
+++ b/src/api/settings.py
@@ -1163,6 +1163,10 @@ async def onboarding(
                     # Import the function here to avoid circular imports
                     from main import ingest_default_documents_when_ready
 
+                    if not config_manager.save_config_file(current_config):
+                        logger.error("Failed to save embedding model to config")
+                        return JSONResponse({"error": "Failed to save configuration"}, status_code=500)
+
                     ingestion_jwt = user.jwt_token if IBM_AUTH_ENABLED and user and user.jwt_token else None
 
                     task_id = await ingest_default_documents_when_ready(
--- a/src/api/settings.py
+++ b/src/api/settings.py
@@ -1115,9 +1115,7 @@ async def onboarding(
 
             # Update model values if provider or model fields were provided
             if body.llm_provider or body.llm_model or body.embedding_provider or body.embedding_model:
-                if not config_manager.save_config_file(current_config):
-                    logger.error("Failed to save embedding model to config")
-                await _update_langflow_model_values(current_config, flows_service)
+                await _update_langflow_model_values(current_config, flows_service, body.llm_model, body.llm_provider, body.embedding_model, body.embedding_provider)
 
         except Exception as e:
             logger.error(
@@ -1481,52 +1479,57 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager 
         # Don't fail the entire settings update if MCP update fails
 
 
-async def _update_langflow_model_values(config, flows_service):
+async def _update_langflow_model_values(config, flows_service, llm_model, llm_provider, embedding_model, embedding_provider):
     """Update model values across Langflow flows for all configured providers"""
     try:
-        import asyncio
-        tasks = []
 
         # 1. Update ONLY the current LLM provider
-        llm_provider = config.agent.llm_provider.lower()
-        tasks.append(
-            flows_service.change_langflow_model_value(
-                llm_provider,
-                llm_model=config.agent.llm_model,
-                force_llm_update=True
-            )
+        llm_provider = llm_provider.lower() if llm_provider else config.agent.llm_provider.lower()
+        llm_model = llm_model if llm_model else config.agent.llm_model
+        flows_service.change_langflow_model_value(
+            llm_provider,
+            llm_model=llm_model,
+            force_llm_update=True
         )
 
-        # 2. Update ALL configured embedding providers
-        embedding_providers = []
-        if config.providers.openai.configured:
-            embedding_providers.append("openai")
-        if config.providers.watsonx.configured:
-            embedding_providers.append("watsonx")
-        if config.providers.ollama.configured:
-            embedding_providers.append("ollama")
 
-        current_embedding_provider = config.knowledge.embedding_provider.lower()
-        for provider in embedding_providers:
-            # Use configured model for current provider, or None (first available) for others
-            embedding_model = (
-                config.knowledge.embedding_model
-                if provider == current_embedding_provider
-                else None
+        if embedding_provider and embedding_model:
+            result = await flows_service.change_langflow_model_value(
+                embedding_provider,
+                embedding_model=embedding_model,
+                force_embedding_update=True
             )
 
-            tasks.append(
-                flows_service.change_langflow_model_value(
-                    provider,
-                    embedding_model=embedding_model,
-                    force_embedding_update=True
+            logger.info(
+                f"Successfully updated Langflow flows for embedding provider {embedding_provider}",
+                result=result
+            )
+        else: 
+            # 2. Update ALL configured embedding providers
+            embedding_providers = []
+            if config.providers.openai.configured:
+                embedding_providers.append("openai")
+            if config.providers.watsonx.configured:
+                embedding_providers.append("watsonx")
+            if config.providers.ollama.configured:
+                embedding_providers.append("ollama")
+
+            current_embedding_provider = config.knowledge.embedding_provider.lower()
+            for provider in embedding_providers:
+                # Use configured model for current provider, or None (first available) for others
+                embedding_model = (
+                    config.knowledge.embedding_model
+                    if provider == current_embedding_provider
+                    else None
                 )
-            )
-
-        # Run all model updates simultaneously
-        await asyncio.gather(*tasks)
-        logger.info("Successfully updated Langflow flows for all configured providers concurrently")
-
+                await flows_service.change_langflow_model_value(
+                        provider,
+                        embedding_model=embedding_model,
+                        force_embedding_update=True
+                    )
+                logger.info(
+                    f"Successfully updated Langflow flows for embedding provider {provider}"
+                )
     except Exception as e:
         logger.error(f"Failed to update Langflow model values: {str(e)}")
         raise
--- a/src/api/settings.py
+++ b/src/api/settings.py
@@ -1113,7 +1113,8 @@ async def onboarding(
             if body.embedding_provider or body.embedding_model:
                 await _update_mcp_servers_with_provider_credentials(current_config, session_manager=session_manager, flows_service=flows_service)
 
-            await _update_langflow_model_values(current_config, flows_service, embedding_model=body.embedding_model, embedding_provider=body.embedding_provider, llm_model=body.llm_model, llm_provider=body.llm_provider)
+            if body.llm_provider or body.llm_model or body.embedding_provider or body.embedding_model:
+                await _update_langflow_model_values(current_config, flows_service, embedding_model=body.embedding_model, embedding_provider=body.embedding_provider, llm_model=body.llm_model, llm_provider=body.llm_provider)
 
         except Exception as e:
             logger.error(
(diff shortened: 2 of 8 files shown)


src/api/settings.py
```diff
@@ -1110,12 +1110,12 @@ async def onboarding(
                 current_config.providers.any_configured()):
                 await _update_langflow_global_variables(current_config, flows_service=flows_service)
 
+            if body.llm_provider or body.llm_model and not (body.embedding_provider or body.embedding_model):
```
mpawlow: ## (a) [High] Operator Precedence Bug Skips LLM Update When Embedding Params Are Also Present

### Summary

The condition has a Python operator-precedence error. `and` binds more tightly than `or`, so the expression evaluates differently than intended. When `body.llm_model` is set but `body.llm_provider` is not, and embedding params are also present in the same request, the LLM model update is silently skipped.

### Problem

- Python evaluates `A or B and C` as `A or (B and C)`.
- The written condition:
  ```python
  if body.llm_provider or body.llm_model and not (body.embedding_provider or body.embedding_model):
  ```
  is parsed as:
  ```python
  if body.llm_provider or (body.llm_model and not (body.embedding_provider or body.embedding_model)):
  ```
- Scenario that silently breaks: `body.llm_provider = None`, `body.llm_model = "ibm/granite-3-8b-instruct"`, `body.embedding_model = "ibm/granite-embedding-278m-multilingual"`.
  - `None or ("ibm/granite-3-8b-instruct" and not True)` → `None or False` → **False** → LLM update never runs.
- The condition is also asymmetric with the embedding branch directly below (line 1116), which has no such exclusion.

### Potential Solution

Wrap the left-hand disjunction in parentheses so both fields are evaluated together:

```python
if (body.llm_provider or body.llm_model) and not (body.embedding_provider or body.embedding_model):
    await _update_langflow_model_values(
        current_config, flows_service,
        llm_model=body.llm_model, llm_provider=body.llm_provider
    )
```

src/api/settings.py
```diff
@@ -1162,9 +1162,6 @@ async def onboarding(
 
                     ingestion_jwt = user.jwt_token if IBM_AUTH_ENABLED and user and user.jwt_token else None
 
-                    if not config_manager.save_config_file(current_config):
```
mpawlow, line 1165: ## (b) [Medium] Removed Config Save Before Ingestion Task May Cause Race Condition

### Summary

A `config_manager.save_config_file(current_config)` call was removed immediately before `ingest_default_documents_when_ready` is enqueued. The final save now occurs at line 1196, which is *after* the background task has been created. If the background task reads config from disk before the save completes, it may ingest with stale embedding model settings.

### Problem

- The removed block:
  ```python
  if not config_manager.save_config_file(current_config):
      logger.error("Failed to save embedding model to config")
  ```
  was placed at `src/api/settings.py` just before calling `ingest_default_documents_when_ready`.
- The final `config_manager.save_config_file(current_config)` call now lives at line 1196, *after* the ingestion task is created (line 1165).
- `ingest_default_documents_when_ready` enqueues a background `TaskService` task. If that task reads the YAML config file to determine the embedding model (rather than receiving it in-memory), it could read a prior version and embed with the wrong model.
- This is particularly dangerous in production where the startup path reads config from disk.

### Potential Solution

Re-add the save (and its error check) just before the ingestion task is created:

```python
if not config_manager.save_config_file(current_config):
    logger.error("Failed to save embedding model to config")
    return JSONResponse({"error": "Failed to save configuration"}, status_code=500)

task_id = await ingest_default_documents_when_ready(...)
```

### Alternative Solutions

- Pass the embedding model explicitly as an argument to `ingest_default_documents_when_ready` so it never reads from disk inside the task, making the ordering irrelevant.

src/api/settings.py
```diff
@@ -1482,47 +1479,58 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager
...
-            )
-            logger.info(
-                f"Successfully updated Langflow flows for embedding provider {provider}"
+                current_llm_provider,
+                llm_model=current_llm_model,
+                force_llm_update=True
             )
 
+        if not llm_model and not llm_provider:
+            if embedding_provider and embedding_model:
```
mpawlow: ## (c) [Medium] Partial Embedding Params Fall Through to "Update All Providers" Path

### Summary

Inside `_update_langflow_model_values`, the condition `if embedding_provider and embedding_model:` requires *both* to be truthy. If the caller provides only one of them (e.g., `embedding_provider` without `embedding_model`), the function silently falls to the `else` branch that updates *all* configured providers, which is likely unintended.

### Problem

- When `embedding_provider = "watsonx"` and `embedding_model = None`, both branches evaluate: `not llm_model` is True, `embedding_provider and embedding_model` is False → falls to `else`, triggering updates across *all* configured providers.
- This regression from the prior PR can produce unexpected overwrites in multi-provider setups.

### Potential Solution

Handle the partial-params case explicitly:

```python
if not llm_model and not llm_provider:
    if embedding_provider or embedding_model:
        # Use config fallback when one param is missing
        effective_provider = embedding_provider or config.knowledge.embedding_provider.lower()
        effective_model = embedding_model or config.knowledge.embedding_model
        await flows_service.change_langflow_model_value(
            effective_provider, embedding_model=effective_model, force_embedding_update=True
        )
    else:
        # No specific params — update all configured providers (original fallback)
        ...
```

### Alternative Solutions

- Validate upstream in `onboarding()` that both `embedding_provider` and `embedding_model` are either both present or both absent before calling `_update_langflow_model_values`, and return a `400` error if only one is supplied.

src/services/flows_service.py
```diff
@@ -1151,15 +1190,15 @@ async def _update_component_fields(
...
         # Update model field and call custom_component/update endpoint
         if "model" in template:
             if "options" not in template["model"]:
+                logger.warning(f"Model field not found in template for provider {provider_name}")
                 return False
 
-            # Update template via Langflow API to get latest options
+            # Enable the model in Langflow first
+            await self._enable_model_in_langflow(provider_name, model_value)
+                # Update template via Langflow API to get latest options
```
mpawlow: ## (d) [Low] Misaligned Comment Creates Visual Confusion in `_update_component_fields`

### Problem

- The comment is indented 4 spaces more than the surrounding lines.
- This is a copy-paste artifact from the old code. It has no functional effect, but it misleads readers into thinking the two statements are at different nesting levels, which could mask future refactoring errors.

### Potential Solution

Fix the comment indentation to match the surrounding code:

```python
            # Enable the model in Langflow first
            await self._enable_model_in_langflow(provider_name, model_value)
            # Update template via Langflow API to get latest options
            template = await self._update_component_langflow(template, template["model"]["value"]) or template
```

src/services/models_service.py
```diff
@@ -492,8 +495,9 @@ async def get_ibm_models(
 
                 # Fetch embedding models
                 embed_params = {
-                    "version": "2024-09-16",
+                    "version": "2026-04-15",
                     "filters": "function_embedding,!lifecycle_withdrawn",
+                    "tech_preview": False,
```
mpawlow: ## (e) [Low] Inconsistent `tech_preview` Filtering Between Text and Embedding Endpoints

### Summary

Embedding model requests pass `"tech_preview": False` as a query parameter, but text chat model requests do not use this parameter. If the WatsonX API does not recognise `tech_preview` as a valid query param, it could be ignored or cause a validation error. The post-response `input_tier == "tech_preview"` filter applied to both is sufficient and is the safe guard; the query param is redundant and inconsistent.

### Problem

```python
# Text chat — no tech_preview query param
text_params = {
    "version": "2026-04-15",
    "filters": "function_text_chat,!lifecycle_withdrawn",
}

# Embedding — adds tech_preview param
embed_params = {
    "version": "2026-04-15",
    "filters": "function_embedding,!lifecycle_withdrawn",
    "tech_preview": False,          # ← asymmetry
}
```

- If `tech_preview` is not in the WatsonX ML API spec for `/ml/v1/foundation_model_specs`, the parameter is silently ignored, making it dead code.
- The server-side filter `if model.get("input_tier") == "tech_preview": continue` is applied to both endpoints and is sufficient.

### Potential Solution

Remove the `"tech_preview": False` query parameter from the embedding request, relying solely on the consistent response-side filter already applied to both endpoints:

```python
embed_params = {
    "version": "2026-04-15",
    "filters": "function_embedding,!lifecycle_withdrawn",
}
```

### Alternative Solutions

- If `tech_preview` is a valid WatsonX API filter, also add it to the text chat params for consistency:
  ```python
  text_params = {
      "version": "2026-04-15",
      "filters": "function_text_chat,!lifecycle_withdrawn",
      "tech_preview": False,
  }
  ```

src/services/flows_service.py
```diff
@@ -1240,12 +1300,18 @@ async def _enable_model_in_langflow(self, provider_name: str, model_value: str):
 
             if response.status_code == 200:
                 logger.info(f"Successfully enabled model {model_value} for provider {provider_name}")
+                # Invalidate cache so it reflects the new state on next check if needed, 
+                # or we could update it locally. Clearing it is safer.
+                async with self._enabled_models_lock:
```
mpawlow, line 1304: ## (h) [Low] Enabled-Models Cache Not Invalidated on Failed Enable Attempt

### Summary

`_enable_model_in_langflow` caches the enabled models list to reduce API calls. The cache is cleared on a successful enable (HTTP 200), but if the enable request returns a non-200 response (warning path) the cache is left stale and will continue reporting the model as disabled for up to 60 seconds, causing repeated unnecessary enable attempts on subsequent calls.

### Problem

```python
if response.status_code == 200:
    logger.info(f"Successfully enabled model ...")
    async with self._enabled_models_lock:
        self._enabled_models_cache = None  # ← cleared only on success
    return True
else:
    logger.warning(
        f"Failed to enable model: HTTP {response.status_code} - {response.text}"
    )
    # ← cache NOT cleared; stale data persists until TTL expiry (60s)
```

- If the first call finds the model disabled in the cache (correctly) but the enable API returns a transient non-200, the cache continues to report the model as disabled. The next 60 seconds of calls will all log "already enabled" and skip the attempt because the cache still holds the prior (pre-enable) snapshot.
- More accurately: the prior successful cache read showed the model was NOT enabled, so after the failed enable, the cache holds data that shows it as not-enabled. On the next request within TTL, the cache will re-check the model and again try to enable it — this part is actually fine. The concern is the inverse: the cache retains the pre-enable state after a *successful* enable.

### Potential Solution

Always clear the cache after any enable attempt (success or failure) to force a fresh read on the next call:

```python
if response.status_code == 200:
    logger.info(f"Successfully enabled model {model_value} for provider {provider_name}")
else:
    logger.warning(f"Failed to enable model: HTTP {response.status_code} - {response.text}")

# Always invalidate cache after an enable attempt
async with self._enabled_models_lock:
    self._enabled_models_cache = None
```

docker-compose.yml
```diff
@@ -10,21 +10,6 @@ services:
     environment:
       - discovery.type=single-node
       - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}
-        # Run security setup in background after OpenSearch starts
-    command: >
```
mpawlow, line 14: ## (i) [Question] Did you intend to remove this block of code or was it only done for local testing environment reasons?

- i.e. The change seems unrelated to the PR

src/services/flows_service.py
```diff
@@ -77,6 +78,21 @@ async def resolve_ollama_url(self, endpoint: str, force_refresh: bool = False) -
     def __init__(self):
         # Cache for flow file mappings to avoid repeated filesystem scans
         self._flow_file_cache = {}
+        # Semaphores to prevent race conditions when updating the same flow concurrently
+        self._flow_locks = {}
```
mpawlow: ## (f) [Low] `_flow_locks` Dictionary Grows Without Bound

### Summary

The `_flow_locks` dictionary in `FlowsService.__init__` accumulates one `asyncio.Lock` per unique flow ID that is ever locked. There is no eviction policy. In long-running deployments with many distinct flow IDs (e.g., after frequent Langflow flow recreation), memory usage grows unboundedly.

### Problem

```python
async def _get_flow_lock(self, flow_id: str) -> asyncio.Lock:
    async with self._lock_creation_lock:
        if flow_id not in self._flow_locks:
            self._flow_locks[flow_id] = asyncio.Lock()
        return self._flow_locks[flow_id]
```

- Locks are never removed after the flow is deleted or no longer active.
- Each `asyncio.Lock` is a small Python object, so this is a low-severity memory leak. However, if `FlowsService` is long-lived, the dictionary could grow significantly in environments with high flow turnover.

### Potential Solution

Use a `weakref`-compatible structure or an LRU eviction cache (e.g., `cachetools.LRUCache`) to cap the number of retained locks:

```python
from cachetools import LRUCache
self._flow_locks = LRUCache(maxsize=128)
```

mpawlow (changes requested):
@lucaseduoli 

### Code Review 1

- See PR comments: (a), (b), (c), (d), (e), (f), (h), (i)
- Note: Please feel free to ignore or optionally implement **Minor** severity items

docker-compose.yml
lucaseduoli: It was intended. It does not exist on the release PR, since the entrypoint takes care of it. I removed it because it got left over

src/api/settings.py
```diff
@@ -1110,12 +1110,15 @@ async def onboarding(
                 current_config.providers.any_configured()):
                 await _update_langflow_global_variables(current_config, flows_service=flows_service)
 
+            if (body.llm_provider or body.llm_model) and not (body.embedding_provider or body.embedding_model):
```
mpawlow: ## (2a) [Low] LLM Update Silently Skipped When Both LLM and Embedding Params Are Present in Same Request

- I'm not sure that this is a valid scenario; but, worth briefly sanity checking 

### Summary

The fix for issue `a` (operator precedence) introduced a mutual-exclusion guard that prevents the LLM update from running when embedding params are also in the same request. When a single onboarding call supplies both `llm_provider`/`llm_model` AND `embedding_provider`/`embedding_model`, only the embedding update executes; the LLM model is never updated in Langflow.

### Problem

- The corrected condition at `settings.py:1113`:
  ```python
  if (body.llm_provider or body.llm_model) and not (body.embedding_provider or body.embedding_model):
      await _update_langflow_model_values(..., llm_model=..., llm_provider=...)
  ```
  The `and not (...)` guard was added to prevent the LLM path from triggering the legacy "update all embedding providers" fallback inside `_update_langflow_model_values`. However, it also blocks the legitimate scenario where both LLM and embedding fields are provided together.
- The embedding branch (line 1119) has no equivalent guard, so it always runs when embedding params are present.
- Net effect: a request containing `{llm_provider: "openai", llm_model: "gpt-4o", embedding_provider: "openai", embedding_model: "text-embedding-3-small"}` would update the embedding component but leave the LLM component pointing at whatever it was previously set to.
- The original pre-PR code always updated both LLM and embedding in a single call regardless of which params triggered it.


### Potential Solution

Remove the `and not (...)` guard from the LLM condition. The `_update_langflow_model_values` function already handles the split internally: when called with `llm_model`/`llm_provider` and no embedding params, it only executes the LLM block; the embedding block is gated on `not llm_model and not llm_provider`. Both calls can fire independently without interfering:

```python
if body.llm_provider or body.llm_model:
    await _update_langflow_model_values(
        current_config, flows_service,
        llm_model=body.llm_model, llm_provider=body.llm_provider
    )

if body.embedding_provider or body.embedding_model:
    await _update_mcp_servers_with_provider_credentials(...)
    await _update_langflow_model_values(
        current_config, flows_service,
        embedding_model=body.embedding_model, embedding_provider=body.embedding_provider
    )
```

mpawlow (approved):
@lucaseduoli 

## Code Review 2

- ✅  Approved / LGTM  🚀 
- See PR comment (2a) - not sure if it is a valid scenario; but, worth sanity checking

src/api/settings.py
```diff
@@ -1482,47 +1481,64 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager
...
+                current_llm_provider,
+                llm_model=current_llm_model,
+                force_llm_update=True
             )
 
-            await flows_service.change_langflow_model_value(
-                provider,
-                embedding_model=embedding_model,
+            logger.info(
+                f"Successfully updated Langflow flows for LLM provider {embedding_provider}",
```
mpawlow: ## (3a) [Low] Logging statement inaccuracy

Should be:
```python
f"Successfully updated Langflow flows for LLM provider {current_llm_provider}",
```

src/api/settings.py
```diff
@@ -1482,47 +1481,64 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager
...
+            effective_model = embedding_model or config.knowledge.embedding_model
+            result = await flows_service.change_langflow_model_value(
+                effective_provider,
+                embedding_model=effective_model,
                 force_embedding_update=True
             )
+
             logger.info(
-                f"Successfully updated Langflow flows for embedding provider {provider}"
+                f"Successfully updated Langflow flows for embedding provider {embedding_provider}",
```
mpawlow: ## (3b) [Low] Logging statement inaccuracy

Should be:

```python
f"Successfully updated Langflow flows for embedding provider {effective_provider}",
```

src/api/settings.py
```diff
@@ -1482,47 +1481,64 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager
...
-
-        current_embedding_provider = config.knowledge.embedding_provider.lower()
-        for provider in embedding_providers:
-            # Use configured model for current provider, or None (first available) for others
-            embedding_model = (
-                config.knowledge.embedding_model
-                if provider == current_embedding_provider
-                else None
+        if llm_model or llm_provider:
+            current_llm_provider = llm_provider.lower() or config.agent.llm_provider.lower()
```
mpawlow: ## (3c) [Medium] Possible AttributeError thrown if LLM provider is not specified

### Problem

- The outer guard is if llm_model or llm_provider: — so this block runs when either is truthy.
- If only body.llm_model is provided and body.llm_provider is None, llm_provider.lower() raises AttributeError: 'NoneType' object has no attribute 'lower' and the entire onboarding request crashes with a 500.
- The prior version was safe: llm_provider.lower() if llm_provider else config.agent.llm_provider.lower().

### Possible Solution
   
- Fix: revert to the conditional form, or use (llm_provider or config.agent.llm_provider).lower().

mpawlow (approved):
@lucaseduoli 
- Post-review observations after optimizations
- See: PR comments (3a), (3b), (3c), (3d), (3e), (3f)

src/api/settings.py
```diff
@@ -1113,9 +1113,7 @@ async def onboarding(
             if body.embedding_provider or body.embedding_model:
                 await _update_mcp_servers_with_provider_credentials(current_config, session_manager=session_manager, flows_service=flows_service)
 
-            # Update model values if provider or model fields were provided
-            if body.llm_provider or body.llm_model or body.embedding_provider or body.embedding_model:
-                await _update_langflow_model_values(current_config, flows_service)
+            await _update_langflow_model_values(current_config, flows_service, embedding_model=body.embedding_model, embedding_provider=body.embedding_provider, llm_model=body.llm_model, llm_provider=body.llm_provider)
```
mpawlow: ## (3d) [Medium] _update_langflow_model_values now runs on every settings save

### Problem

- Calls the function unconditionally, not gated by if any([body.*_provider, body.*_model]).
- When none of the four model/provider fields are in the request (e.g., a save that only toggles should_ingest_sample_data or updates onboarding state), the function falls through to the if not (embedding_model or ... or llm_provider): branch and does a full "update ALL
  configured embedding providers" sweep against Langflow.
- This is wasted work on every unrelated settings write, and it can overwrite manually-tweaked embedding nodes in Langflow with the config-default model.
 
### Possible Solution 

- Fix: keep an outer guard at the call site:
```python
  if body.llm_provider or body.llm_model or body.embedding_provider or body.embedding_model:
      await _update_langflow_model_values(...)
```

src/api/settings.py
```diff
@@ -1482,47 +1481,64 @@ async def _update_mcp_servers_with_provider_credentials(config, session_manager
...
-            await flows_service.change_langflow_model_value(
-                provider,
-                embedding_model=embedding_model,
+            logger.info(
+                f"Successfully updated Langflow flows for LLM provider {embedding_provider}",
+                result=result
+            )
+
+        if embedding_model or embedding_provider:
+            effective_provider = embedding_provider or config.knowledge.embedding_provider.lower()
```
mpawlow: ## (3e) [Medium] embedding_provider is not lowercased when passed directly

### Problem

- The fallback branch lowercases; the direct branch does not. If the frontend sends "OpenAI" or "WatsonX", effective_provider retains the original casing and downstream change_langflow_model_value → _get_provider_name_display comparisons (which use lowercase keys like "openai",
  "watsonx", "ollama") will miss.
- The LLM path at line 1489 attempts to .lower() both, though broken for None (flagged previously).

### Possible Solution 
 
- Fix: effective_provider = (embedding_provider or config.knowledge.embedding_provider).lower().

mpawlow: ## (3f) [High] Cross-contamination risk: partial params fall back to unrelated config values

### Problem

```python
  if embedding_model or embedding_provider:
      effective_provider = embedding_provider or config.knowledge.embedding_provider.lower()
      effective_model = embedding_model or config.knowledge.embedding_model
```
 
- If the caller sends embedding_provider = "watsonx" but omits embedding_model, effective_model falls back to the currently configured model from config.knowledge.embedding_model — which may be an OpenAI model like "text-embedding-3-small". Pairing "watsonx" with an OpenAI model
  ID would push an invalid configuration into Langflow.
- Same hazard in the LLM branch: llm_provider="anthropic" with no llm_model would pull the last OpenAI/WatsonX model from config.agent.llm_model.

### Possible Solution
   
- Fix: only fall back to config values when the provider matches:
```python
  if embedding_provider and embedding_provider.lower() != config.knowledge.embedding_provider.lower():
      effective_model = embedding_model  # do not fall back; force caller to specify
  else:
      effective_model = embedding_model or config.knowledge.embedding_model
```
   
- Or reject partial input upstream at the /onboarding endpoint with a 400.

PR #1450, opened 2026-04-21
fix: Add deleting state to ensure proper loading state

mfortman11:
**Issue**                                                                                                                                                                                                          
                                                                                                                                                                                                                 
  The delete dialog's loading spinner never showed because of two bugs:                                                                                                                                          
                                                                                                                                                                                                                 
  1. useMutation has a single state machine. Calling mutateAsync in parallel (one per file) caused the first completed request to flip isPending back to false, even with others still in-flight.                
  2. The dialog's handleConfirm read isLoading from a stale closure — captured before the button was clicked when it was still false — so the guard meant to keep the dialog open did nothing.    
  
  This resulted in QA being able to click it twice resulting in a "No document found" error message as shown [here](https://github.ibm.com/lakehouse/tracker/issues/70225)           
                                                                                                                                                                                                                 
  **Fix**                                                                                                                                                                                                            
                                                                                                                                                                                                                 
  Replaced deleteDocumentMutation.isPending with a local isBulkDeleting state. It's set to true before the parallel deletes fire and cleared in a finally block, giving the spinner a stable, accurate signal    
  regardless of how many mutations are running.

### DIFF
--- a/frontend/app/knowledge/page.tsx
+++ b/frontend/app/knowledge/page.tsx
@@ -97,6 +97,7 @@ function SearchPage() {
   const { parsedFilterData, queryOverride } = useKnowledgeFilter();
   const [selectedRows, setSelectedRows] = useState<File[]>([]);
   const [showBulkDeleteDialog, setShowBulkDeleteDialog] = useState(false);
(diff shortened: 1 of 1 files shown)


Wallgau (approved):
LGTM

PR #1453, opened 2026-04-22
fix: add load dotenv on encryption file

lucaseduoli:
* added .env loading inside encryption

* add load dotenv on bootstrap at the top of the file

* added lifestan in fastapi

* Added checksum

* Pass startup and shutdown events to lifespan

* load dotenv on top of everything

* Add timestamp to log


edwinjosechittilappilly (approved):
LGTM

PR #1459, opened 2026-04-23
fix: fixes the jwt in delete existing default docs

mfortman11:
The issue

  - _delete_existing_default_docs forces AnonymousUser + session_manager._anonymous_jwt regardless of caller.                                                                                 
  - create_user_opensearch_client (config/settings.py:813-818) wraps any token lacking a Basic  / Bearer  prefix as Authorization: Bearer <token>.                                            
  - Under IBM auth, the real caller's User.jwt_token is actually "Basic {lh_credentials}" — what the Lakehouse gateway requires. The anonymous JWT is just a signed OpenRAG token, which the  
  gateway rejects with plain-text Bad Request\n.                                                                                                                                              
  - Timing and symptom shape match: single round-trip ~19ms between the "Refreshing default OpenRAG docs" log and the error; no JSON body; "Deleted existing default OpenRAG docs" success log
   never appears.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -760,7 +760,9 @@ async def _materialize_default_docs_url_as_text_file(
     return temp_file.name
 
 
-async def _delete_existing_default_docs(session_manager, connector_type: str):
+async def _delete_existing_default_docs(
+    session_manager, connector_type: str, jwt_token=None, user_id=None
+):
     """Delete previously ingested default OpenRAG docs before reingestion."""
     from session_manager import AnonymousUser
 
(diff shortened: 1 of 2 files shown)


src/main.py
```diff
@@ -771,16 +773,18 @@ async def _delete_existing_default_docs(session_manager, connector_type: str):
...
         session_manager.get_user_opensearch_client(
-            anonymous_user.user_id, effective_jwt
+            anonymous_user.user_id, None
         )
         if hasattr(session_manager, "_anonymous_jwt"):
             effective_jwt = session_manager._anonymous_jwt
 
     opensearch_client = session_manager.get_user_opensearch_client(
-        anonymous_user.user_id, effective_jwt
+        effective_user_id, effective_jwt
```
edwinjosechittilappilly: Latest docs needs to be the default user always. if we do this only the current user would be able to see.
Just checking if its a known change.

edwinjosechittilappilly (approved):
LGTM, can you add an issue to check the anonymous user jwt pathway in Saas, i anticipate we might have a bug there too

Issue #1335, opened 2026-04-06
[Bug]: Uploading documents larger than 100 KB fails when using watsonx.ai.


Revathypv:
### OpenRAG Version

0.4.1.dev11

### Deployment Method

uv add (installed in project)

### Operating System

macOS 15.7.5

### Python Version

3.13

### Affected Area

Onboarding (setup wizard, initial configuration), Ingestion (document processing, upload, Docling)

### Bug Description

User is getting error when uploading documents >100kb using only watsonx.ai model provider

### Steps to Reproduce

1. Update watsonx.ai as model provider in the settings page. Remove other models if already configured
2. Upload attached documents

[AMCOR_2022_8K_2022-07-01.pdf](https://github.com/user-attachments/files/26512116/AMCOR_2022_8K_2022-07-01.pdf)

[ULTABEAUTY_2024Q2_EARNINGS.pdf](https://github.com/user-attachments/files/26512120/ULTABEAUTY_2024Q2_EARNINGS.pdf)

### Expected Behavior

User should be able to upload documents successfully

### Actual Behavior

```
Failure Log (1 of 1 pending)

> AMCOR_2022_8K_2022-07-01.pdf

Error running graph: Error building Component OpenSearch (Multi-Model Multi-Embedding): Failure during generate. (POST https://us-south.ml.cloud.ibm.com/ml/v1/text/embeddings?version=2026-02-03) Status code: 400, body: {"errors":[{"code":"invalid_input_argument","message":"Invalid input argument for Model 'ibm/slate-125m-english-rtrvr-v2': This model's maximum context length is 512 tokens. However, you requested 571 tokens in the input for embedding generation. Please reduce the length of the input.","more_info":"https://cloud.ibm.com/apidocs/watsonx-ai#text-embeddings"}],"trace":"b2831c3cc6a051b9c23aee42f0fa1375","status_code":400}
```

<img width="1728" height="915" alt="Image" src="https://github.com/user-attachments/assets/f9b62c8a-734c-4b56-bf7b-ad0403e43219" />

### Relevant Logs

```shell

```

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before.
- [x] I have provided all the requested information.

aimurphy:
Sorry for the delay. Added troubleshooting for this in https://github.com/langflow-ai/openrag/pull/1461

PR #1461, opened 2026-04-24
docs: Add troubleshooting for reducing chunk size

aimurphy:
For #1335 

Added troubleshooting and explanation that chunk size must be less than the model's supported token length.


mendonk (approved):
+1

PR #1462, opened 2026-04-24
fix: Ingest document from chat is not working (SaaS)

mpawlow:
### Issue

- GH-69225

### Target Deployment

- **Cloud / SaaS**
- [release-saas-0.1](https://github.com/langflow-ai/openrag/tree/release-saas-0.1) branch

### Related Pull Request

- https://github.com/langflow-ai/openrag/pull/1463

### Summary

- Gated the chat input file upload ("+" button) behind a new `OPENRAG_INGEST_VIA_CHAT` feature flag, defaulting to hidden, so operators can opt-in to chat-based ingestion without changing existing behaviour.

### Backend

- Added `OPENRAG_INGEST_VIA_CHAT` env var to `src/config/settings.py` (defaults `false`).
- Exposed `ingest_via_chat` field on `SettingsResponse` in `src/api/settings.py`, populated from the new env var.
- Documented the env var in `.env.example`.

### Frontend

- Added `ingest_via_chat` field to the `Settings` interface in `useGetSettingsQuery.ts`.
- Added `ingestViaChat` prop to `ChatInput` and conditionally rendered the "+" upload button only when the prop is `true`.
- Passed `settings?.ingest_via_chat ?? false` from `chat/page.tsx` to `ChatInput`.
- Reformatted `tsconfig.json` (array expansions, `jsx: preserve`).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,11 +3,6 @@
 # If unset or false, Langflow pipeline will be used (default: upload -> ingest -> delete)
 DISABLE_INGEST_WITH_LANGFLOW=false
 
-# OPTIONAL: Show the "+" file upload button in the chat input, allowing users to
-# attach files directly in the chat interface for additional context during conversation.
-# Default: false
-# OPENRAG_INGEST_VIA_CHAT=false
-
 # Set to false to skip ingesting OpenRAG sample docs during onboarding
 # Default: true
 INGEST_SAMPLE_DATA=true
--- a/.env.example
+++ b/.env.example
@@ -3,6 +3,11 @@
 # If unset or false, Langflow pipeline will be used (default: upload -> ingest -> delete)
 DISABLE_INGEST_WITH_LANGFLOW=false
 
+# OPTIONAL: Show the "+" file upload button in the chat input, allowing users to
+# attach files directly in the chat interface for additional context during conversation.
+# Default: false
+# OPENRAG_INGEST_VIA_CHAT=false
+
 # Set to false to skip ingesting OpenRAG sample docs during onboarding
 # Default: true
 INGEST_SAMPLE_DATA=true
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -11,7 +11,7 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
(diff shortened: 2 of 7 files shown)


frontend/tsconfig.json
```diff
@@ -11,15 +15,17 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "react-jsx",
```
mpawlow, line 14: ### [Question] Is this change appropriate?

- Need a sanity check here on the value change from `react-jsx` -> `preserve`
- The odd thing is that I did not manually modify this file
- At runtime, during OpenRAG local dev env start-up, it appears that Next.js dynamically updated the file with these formatting changes
- `Problem`: These formatting changes caused the Biome check to fail 
   - ✅  fixed the formatting errors in this file and Biome check passes now

.env.example
```diff
@@ -3,6 +3,11 @@
 # If unset or false, Langflow pipeline will be used (default: upload -> ingest -> delete)
 DISABLE_INGEST_WITH_LANGFLOW=false
 
+# OPTIONAL: Show the "+" file upload button in the chat input, allowing users to
+# attach files directly in the chat interface for additional context during conversation.
+# Default: false
+# OPENRAG_INGEST_VIA_CHAT=false
+
```
mfortman11: I won't add this here no need so give details on of FFs

frontend/tsconfig.json
```diff
@@ -11,7 +11,7 @@
...
+    "jsx": "preserve",
```
mfortman11, line 14: confirmed next makes this change so it should be fine

mfortman11 (approved):
LGTM

frontend/tsconfig.json
mpawlow: Thanks!

.env.example
mpawlow: Will add an additional commit to remove the comments from `.env.example` accordingly

mpawlow: Additional commit added  🚀

frontend/tsconfig.json
mpawlow: ✅  Confirmed offline with @mfortman11 that the change is correct / fine.

PR #1466, opened 2026-04-24
feat: user can rename file name (excluding url)

Wallgau:
**User flow:**
**Knowledge table:** 

- Row actions include Rename (opens a dialog hosted on the knowledge page so the grid refetch does not unmount the modal).
- Rename updates the display filename on indexed chunks in OpenSearch. document_id is not changed (copy in the dialog explains that).
- Partial renames: If only some chunks update, the API returns 422 with counts; the UI shows a warning, keeps the dialog open, shows a short summary, and the primary action becomes Retry until complete.
- Rename is hidden for connector types treated as non-file uploads: url, openrag_docs, system_default (case-insensitive after trim).
- Frontend
- useGetSearchQuery: Chunks and aggregated files carry optional document_id; when merging chunks into one file row, the first non-empty document_id wins so the grid can pass a stable id into rename.
- useRenameDocument: POST to /api/documents/rename with current_filename, new_filename, optional document_id. Treats 422 + partial: true as a normal resolved response (no throw). Types cover partial, idempotent, resumed, counts, and echoed document_id.

**knowledge/page.tsx:** 

- Local state for the rename dialog; handleRenameSave branches toasts for partial / idempotent / resumed / full success; after a successful mutateAsync, queryClient.invalidateQueries({ queryKey: ["search"] }) then refreshTasks() so the list refreshes immediately with no delayed invalidation in the hook.
- knowledge-actions-dropdown: Passes documentId and onOpenRename up to the page; Rename disabled if onOpenRename is missing.

**Backend**

- POST /documents/rename (RenameDocumentBody: current_filename, new_filename, optional document_id).
- rename_document_chunks_core: Owner- and filename-alias–scoped match; optional document_id to disambiguate when several documents share a name; collision check so another document cannot already own the target name (excluding the same document_id when applicable).
- Multi-pass update_by_query with refresh and conflict handling; on incomplete rename returns 422 with partial, updated_chunks, remaining_old_chunks, matched_chunks, and document_id for retries.
- Resume path: If no chunks match the current filename anymore but document_id is supplied, logic can finish chunks that still do not match the new filename aliases (e.g. after a partial rename).
- Idempotent success when everything already shows the new name for that document_id.
- check_filename_exists: Optional content_hash query param; if present, can treat duplicate as content hash match on document_id (in addition to filename alias logic); responses can include match_type.

https://github.com/user-attachments/assets/b489b5c9-bea1-42a4-8ebc-455c40d625f8

https://github.com/user-attachments/assets/d1b77b08-7284-429a-95a0-30b4ca9b6df7


Wallgau:
created another PR to deal with conflicts and reduce the amount of code.

Wallgau:
please review this one instead: https://github.com/langflow-ai/openrag/pull/1532

PR #1468, opened 2026-04-25
fix: Knowledge search fails with embedding model error

mpawlow:
### Issue

- #69807

### Summary

- Fixed knowledge search failing with an embedding model error when a previously configured provider was removed after documents were ingested.
- Search now degrades gracefully to keyword-only results instead of hard-failing, and the UI surfaces a non-fatal warning banner explaining the degraded state.

### Backend: Graceful Embedding Failure & Provider-Removal Guard

- Added `strict` mode to `get_litellm_model_name` in `models_service.py` that raises `UnknownEmbeddingProvider` immediately when no configured provider claims a model, preventing LiteLLM from entering a costly retry loop on an unroutable model.
- Refactored `search_service.py` to gather embedding results with `return_exceptions=True`, tracking per-model failures in `failed_models` and continuing with the subset of models that succeeded.
- Introduced keyword-only fallback when all embeddings fail: removed the embedding-field `exists` filter from the query so documents without a matching embedding field are still returned, and bumped `multi_match` boost to compensate for the absent semantic signal.
- Added `warnings` to the search response payload (`code: "embedding_unavailable"`) to signal degraded semantic search to the client.
- Added `force_remove` flag to `SettingsUpdateBody` in `settings.py` and a pre-removal check (`_affected_embedding_models`) that returns a structured `409` response listing affected models and chunk counts before allowing a provider to be removed.

### Frontend: 409 Force-Remove Flow & Warning Banner

- Introduced `UpdateSettingsError` typed error class and `isEmbeddingProviderInUseError` type guard in `useUpdateSettingsMutation.ts` to preserve the structured 409 payload from the backend.
- Added `force_remove` to `UpdateSettingsRequest` and wired it through the Ollama, OpenAI, and watsonx settings dialogs; on a 409, `affectedModels` state is populated and the dialog footer switches to a force-confirmation view listing impacted models and chunk counts.
- Updated `useGetSearchQuery.ts` return type from `File[]` to `SearchResult` (`{ files, warnings }`) and propagated the change through `knowledge/page.tsx` and `knowledge/chunks/page.tsx`.
- Rendered `embedding_unavailable` warnings in the knowledge page as amber banner alerts that explain the degraded state and guide users to re-ingest or re-configure the missing provider.

---

### API Changes

- Removed eager `update_model_registry()` call from `update_settings` and passed `models_service` into `_run_async_post_save_langflow_updates` so the registry refresh happens in the background after save completes.
- Added `models_service` parameter to `_run_async_post_save_langflow_updates`; registry refresh only runs when a provider update was detected.

### Model Registry Changes

- Registered Ollama models with no `capabilities` field as potential embedding models, defaulting `nomic-embed-text` variants as the default embedding model.
- Preserved graceful error handling so models that cannot actually embed fail at the LiteLLM call level rather than silently dropping from the registry.

### DIFF
--- a/src/services/search_service.py
+++ b/src/services/search_service.py
@@ -116,6 +116,7 @@ class SearchService:
         # Get available embedding models from corpus
         query_embeddings = {}
         available_models = []
+        failed_models: list = []
 
         opensearch_client = self.session_manager.get_user_opensearch_client(
             user_id, jwt_token
@@ -198,9 +199,15 @@ class SearchService:
                 attempts = 0
                 last_exception = None
 
-                # Use centralized utility for LiteLLM model formatting
+                # Use centralized utility for LiteLLM model formatting.
+                # strict=True: if no configured provider claims this model
+                # (e.g. the provider was removed after ingest), raise
+                # immediately rather than entering a ~3s retry loop on an
+                # unroutable model name.
                 if self.models_service:
-                    formatted_model = await self.models_service.get_litellm_model_name(model_name)
+                    formatted_model = await self.models_service.get_litellm_model_name(
+                        model_name, strict=True
+                    )
                 else:
                     # Fallback if service not injected (tests/etc)
                     formatted_model = model_name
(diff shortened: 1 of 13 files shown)


mpawlow:
### Manual Test Plan

- [test-plan-1468.md](https://github.com/user-attachments/files/27090441/test-plan-1468.md)

mpawlow:
## UI / UX Changes

### Ingest a variety of documents using OpenAI and watsonx.ai model providers

image

### Remove watsonx.ai model provider

- User prompted to force remove the model provider if existing documents ingested with it are detected in Knowledge

image

### Search for term in document ingested via removed model provider

- Fallback keyword search works
- Warning shown as banner

image

- Fallback keyword search works
- Warning shown as banner

image

### Search for term in document ingested via current model provider

- Normal search still works
- Warning shown as banner

image

src/services/search_service.py
```diff
@@ -198,9 +199,15 @@ async def embed_with_model(model_name):
                 attempts = 0
                 last_exception = None
 
-                # Use centralized utility for LiteLLM model formatting
+                # Use centralized utility for LiteLLM model formatting.
```
mpawlow, line 202: ### Request for special focus or emphasis on reviewing changes in this class / file

- e.g. Due to risk of regression
- e.g. Due to sensitive code area

lucaseduoli (approved):
LGTM

PR #1473, opened 2026-04-27
fix: changed frontend checks to pass correct embedding model to backend

lucaseduoli:
This pull request simplifies and clarifies the onboarding logic in the `OnboardingCard` component, specifically around how embedding models are selected and validated. The changes remove unnecessary conditions and streamline how the embedding model is set and validated during onboarding.

Validation and completion logic simplification:

* The check for required fields now only requires `settings.embedding_model` to be present for embedding providers, removing the dependency on `showProviderConfiguredMessage`.
* The onboarding completion status for embedding providers now depends solely on the presence of `settings.embedding_model`, rather than also considering `showProviderConfiguredMessage`.

Embedding model selection logic:

* The logic for setting `onboardingData.embedding_model` now prioritizes `settings.embedding_model`, and only falls back to `currentSettings.knowledge.embedding_model` if the provider matches, further simplifying the previous conditional structure.

### DIFF
--- a/frontend/app/onboarding/_components/onboarding-card.tsx
+++ b/frontend/app/onboarding/_components/onboarding-card.tsx
@@ -382,9 +382,7 @@ const OnboardingCard = ({
 
     if (
       !currentProvider ||
-      (isEmbedding &&
-        !settings.embedding_model &&
-        !showProviderConfiguredMessage) ||
+      (isEmbedding && !settings.embedding_model) ||
       (!isEmbedding && !settings.llm_model)
     ) {
       toast.error("Please complete all required fields");
@@ -400,16 +398,15 @@ const OnboardingCard = ({
     // Set the provider field
     if (isEmbedding) {
       onboardingData.embedding_provider = currentProvider;
-      // If provider is already configured, use the existing embedding model from settings
-      // Otherwise, use the embedding model from the form
-      if (
+      if (settings.embedding_model) {
+        onboardingData.embedding_model = settings.embedding_model;
+      } else if (
         showProviderConfiguredMessage &&
-        currentSettings?.knowledge?.embedding_model
+        currentSettings?.knowledge?.embedding_model &&
+        currentProvider === currentSettings?.knowledge?.embedding_provider
       ) {
         onboardingData.embedding_model =
           currentSettings.knowledge.embedding_model;
-      } else {
-        onboardingData.embedding_model = settings.embedding_model;
       }
     } else {
       onboardingData.llm_provider = currentProvider;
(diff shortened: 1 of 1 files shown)


lucaseduoli:
Tested in cloud and working!

frontend/app/onboarding/_components/onboarding-card.tsx
```diff
@@ -400,16 +398,15 @@ const OnboardingCard = ({
     // Set the provider field
     if (isEmbedding) {
       onboardingData.embedding_provider = currentProvider;
-      // If provider is already configured, use the existing embedding model from settings
-      // Otherwise, use the embedding model from the form
-      if (
+      if (settings.embedding_model) {
+        onboardingData.embedding_model = settings.embedding_model;
+      } else if (
```
mpawlow: ## (a) [Low] Dead Code — `else if` Fallback in `handleComplete` Is Unreachable

### Summary

After the validation guard is tightened to require `settings.embedding_model`, the new `else if` branch that reads from `currentSettings.knowledge.embedding_model` is dead code that can never execute.

### Problem

- In lines 383–390, `handleComplete` **returns early** whenever `settings.embedding_model` is falsy for an embedding provider. After this gate, when `isEmbedding === true`, `settings.embedding_model` is guaranteed to be a non-empty string.
- The `if` branch is taken whenever `settings.embedding_model` is truthy — guaranteed by the validation gate.
- The `else if` branch requires `settings.embedding_model` to be falsy — which, by the validation gate, is impossible at this point in the code.
- The `else if` block is therefore unreachable and can never execute. It also introduces a path where `onboardingData.embedding_model` could remain `undefined` (if somehow neither branch runs), which would silently send no model to the backend.

### Code References

- `frontend/app/onboarding/_components/onboarding-card.tsx` lines 383–390 (validation gate)
- `frontend/app/onboarding/_components/onboarding-card.tsx` lines 401–409 (dead `else if` block)

### Potential Solution

- Remove the `else if` branch entirely, since the validation gate already guarantees `settings.embedding_model` is set when this code is reached:

```tsx
if (isEmbedding) {
  onboardingData.embedding_provider = currentProvider;
  onboardingData.embedding_model = settings.embedding_model;
} else {
  onboardingData.llm_provider = currentProvider;
  onboardingData.llm_model = settings.llm_model;
}
```

- This also removes the ambiguity where `onboardingData.embedding_model` could be left unset.

mpawlow (approved):
### Code Review 1

- ✅  Approved / LGTM  🚀 
- One **Low** priority issue identified
   - See PR comment: (a)
   - Please feel free to ignore or optionally implement

PR #1475, opened 2026-04-27
fix: ensure model selections are reflected in LF

mfortman11:
Model selections weren't being reflected in LF


lucaseduoli (approved):
LGTM

PR #1477, opened 2026-04-27
fix: ensure model selections are reflected in LF

mfortman11:
make sure the model selection works and is shown in LF


lucaseduoli (approved):
lgtm

PR #1478, opened 2026-04-27
fix: changed frontend checks to pass correct embedding model to backend

lucaseduoli:
This pull request refines the onboarding logic for selecting embedding and LLM models in the `OnboardingCard` component. The changes ensure that model selection and validation are more accurate, especially when handling providers that are already configured.

**Improvements to model selection and validation logic:**

* The validation for required fields now only checks for the presence of `embedding_model` when `isEmbedding` is true, removing the dependency on `showProviderConfiguredMessage` for this check.
* The logic for setting `embedding_model` in the onboarding data now prefers the value from `settings.embedding_model`. It only falls back to the provider-configured model if the provider matches and the model is available, preventing accidental mismatches.
* The condition for marking the onboarding step as complete is simplified to require only `settings.embedding_model` when `isEmbedding` is true, rather than also checking `showProviderConfiguredMessage`.

Duplicate of #1473

### DIFF
--- a/frontend/app/onboarding/_components/onboarding-card.tsx
+++ b/frontend/app/onboarding/_components/onboarding-card.tsx
@@ -382,9 +382,7 @@ const OnboardingCard = ({
 
     if (
       !currentProvider ||
-      (isEmbedding &&
-        !settings.embedding_model &&
-        !showProviderConfiguredMessage) ||
+      (isEmbedding && !settings.embedding_model) ||
       (!isEmbedding && !settings.llm_model)
     ) {
       toast.error("Please complete all required fields");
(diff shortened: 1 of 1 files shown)


mpawlow (approved):
### Code Review 1

- ✅ LGTM / Approved  🚀 
- Approval based on changes already reviewed, approved & merged into `release-saas-0.1` for the following Pull Request:
   - https://github.com/langflow-ai/openrag/pull/1473
- ✅ All changes from the source Pull Request are cleanly included in the target Pull Request
   - No commits are missing.
   - No regressions were introduced by the rebase.
   - All pre-existing issues (resolved or unresolved) were carried forward verbatim.

PR #1474, opened 2026-04-27
fix: pass parameters as global variable headers to MCP server on agent flow

lucaseduoli:
This pull request refactors how MCP server updates are handled and improves the propagation of user identity information (owner, name, email) to Langflow chat and upload endpoints. The main changes are a simplification of MCP server update logic (removing provider credential propagation and focusing on server URLs), and ensuring user identity is consistently included in chat-related headers.

**MCP Server Update Refactor:**

* Replaced all uses of `_update_mcp_servers_with_provider_credentials` with a new `_update_mcp_server_urls` function, which now only updates MCP server URLs (removing the logic for propagating provider credentials and anonymous user info). This affects startup, onboarding, settings reapplication, and post-save hooks. [[1]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1119-R1119) [[2]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1439-R1439) [[3]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1453-R1463) [[4]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1647-R1625) [[5]](diffhunk://#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312L1173-R1193) [[6]](diffhunk://#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312L1342-R1304)

* Removed the `build_mcp_global_vars_from_config` function and all related code for building and sending provider credentials to MCP servers. [[1]](diffhunk://#diff-62094c4bdc99f526cd5a3f567d7fba2be315cd159906e3279fcded67c2896894L76-L131) [[2]](diffhunk://#diff-36bf71c2e2702cd04dd28bc3792fe0935b2c5a3588c300c8506b4504f71b39ddL379-R380)

**User Identity Propagation in Chat/Upload:**

[...]

PR #1479, opened 2026-04-27
fix: pass parameters as global variable headers to MCP server on agent flow

lucaseduoli:
This pull request makes significant changes to how MCP (Model Control Plane) server updates and user ownership information are handled throughout the system. The main improvements are: user and owner metadata are now consistently passed to Langflow chat and document upload endpoints, MCP server update logic is simplified to focus on updating server URLs rather than injecting provider credentials, and related utility code is removed or refactored to match the new approach.

**MCP server update refactor:**

* All logic for updating MCP servers with provider credentials is replaced with a streamlined approach that only updates MCP server URLs using `update_all_mcp_server_urls`, removing the previous use of `build_mcp_global_vars_from_config` and related credential injection. [[1]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1458-R1468) [[2]](diffhunk://#diff-cae7dfc71e73f99be2acece4b754e742fd7b262340229af93b03d2c826e5d3ceL1652-R1630) [[3]](diffhunk://#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312L1182-R1202) [[4]](diffhunk://#diff-2e5ad92c43aa96cc3a9cef6c6aec998b216f1379c43b1f651013d25e55989312L1350-R1312) [[5]](diffhunk://#diff-36bf71c2e2702cd04dd28bc3792fe0935b2c5a3588c300c8506b4504f71b39ddL379-R380) [[6]](diffhunk://#diff-62094c4bdc99f526cd5a3f567d7fba2be315cd159906e3279fcded67c2896894L76-L131)

**User/owner metadata propagation:**

* The `owner`, `owner_name`, and `owner_email` fields are now included in all relevant chat and upload context endpoints and passed through to the Langflow chat service, ensuring user identity is consistently available for downstream processing. [[1]](diffhunk://#diff-ffceb2a90b7e00399a465d8e90d36bec5d978025ff8784fef9c9fee4b972b7c5R105-R107) [[2]](diffhunk://#diff-ffceb2a90b7e00399a465d8e90d36bec5d978025ff8784fef9c9fee4b972b7c5R125-R127) [[3]](diffhunk://#diff-a4f7adb8dc47fe3831c1836b8dbc26ebdde5b05d83708db56e480e063bf024c7R132-R134) [[4]](diffhunk://#diff-a4f7adb8dc47fe3831c1836b8dbc26ebdde5b05d83708db56e480e063bf024c7R150-R152) [[5]](diffhunk://#diff-84150c6a4d4a4e82e41dd5b9e4146e956f5f7bff1bc379245b40e012188b0bb2R127-R142) [[6]](diffhunk://#diff-a656d24ea3c40e5b3ba0c8c3e11af85a45cad1ba6f2ffd9ddf83af25408a6739R60-R62) [[7]](diffhunk://#diff-a656d24ea3c40e5b3ba0c8c3e11af85a45cad1ba6f2ffd9ddf83af25408a6739R322-R335)

**Header injection for owner metadata:**

* Owner information is injected into Langflow requests using custom HTTP headers (`X-Langflow-Global-Var-OWNER`, etc.), supporting both authenticated and no-auth modes. [[1]](diffhunk://#diff-a656d24ea3c40e5b3ba0c8c3e11af85a45cad1ba6f2ffd9ddf83af25408a6739L71-R78) [[2]](diffhunk://#diff-a656d24ea3c40e5b3ba0c8c3e11af85a45cad1ba6f2ffd9ddf83af25408a6739R322-R335)

**Code cleanup and removal of unused utilities:**

* The `build_mcp_global_vars_from_config` utility and all related credential-injection code are removed from `utils/langflow_headers.py` and service layers, reflecting the new, simplified update logic. [[1]](diffhunk://#diff-62094c4bdc99f526cd5a3f567d7fba2be315cd159906e3279fcded67c2896894L76-L131) [[2]](diffhunk://#diff-36bf71c2e2702cd04dd28bc3792fe0935b2c5a3588c300c8506b4504f71b39ddL379-R380)

**Minor refactoring and import updates:**

* Unused imports and background task tracking are removed from the authentication service, and imports are updated to match the new logic. [[1]](diffhunk://#diff-36bf71c2e2702cd04dd28bc3792fe0935b2c5a3588c300c8506b4504f71b39ddL8-R8) [[2]](diffhunk://#diff-36bf71c2e2702cd04dd28bc3792fe0935b2c5a3588c300c8506b4504f71b39ddL40-R40)

These changes simplify the MCP update process, make user/owner metadata handling more robust and consistent, and remove legacy credential-injection code that is no longer needed.

Duplicate of #1474


src/api/chat.py
```diff
@@ -118,6 +122,9 @@ async def langflow_endpoint(
                 previous_response_id=body.previous_response_id,
                 stream=False,
                 filter_id=body.filter_id,
+                owner=user.user_id if not is_no_auth_mode() else None,
```
mpawlow: ## (a) [Major] OWNER header is empty string in no-auth mode — regression from prior "anonymous" identity

### Summary

In no-auth mode, `OWNER` is intentionally set to `None` by callers (resulting in an empty-string header), while `OWNER_NAME` and `OWNER_EMAIL` are still set to the `AnonymousUser` values (`"Anonymous User"` / `"anonymous@localhost"`). Previously, the startup code explicitly set `OWNER = "anonymous"`. Any Langflow flow logic that reads the `OWNER` global variable to do routing, filtering, or access control now receives `""` instead of `"anonymous"`, which is a silent behavioral regression.

### Problem

- In `chat.py` and `v1/chat.py`, the `owner` field is conditionally `None` in no-auth mode:
  ```python
  owner=user.user_id if not is_no_auth_mode() else None,
  owner_name=user.name,   # "Anonymous User" — always passed
  owner_email=user.email, # "anonymous@localhost" — always passed
  ```
- In `chat_service.py`, a `None` owner becomes an empty string header:
  ```python
  "X-Langflow-Global-Var-OWNER": str(owner) if owner else "",
  ```
- The previous code (now deleted from `auth_service.py`, `main.py`, and `settings.py`) explicitly set `OWNER = "anonymous"` (the `AnonymousUser.user_id`) for no-auth mode. With that code removed, downstream Langflow flows that branch on `OWNER` now receive `""` instead of `"anonymous"`.
- The inconsistency is also visible in `upload.py` where `owner_name` and `owner_email` are conditionally set via `user.name if user else None`, meaning `AnonymousUser` values will always propagate for the name and email — but the owner ID will not.

### Code References

- `src/api/chat.py` lines 105–107, 125–127
- `src/api/v1/chat.py` lines 132–134, 150–152
- `src/api/upload.py` lines 129–131
- `src/services/chat_service.py` lines 74–78
- `src/session_manager.py` lines 36–44 (`AnonymousUser` class — `user_id = "anonymous"`)

### Potential Solution

Pass `user.user_id` (which is `"anonymous"` for `AnonymousUser`) unconditionally as the owner, without the `is_no_auth_mode()` gate:

```python
owner=user.user_id,         # "anonymous" in no-auth mode, real ID otherwise
owner_name=user.name,
owner_email=user.email,
```

This matches the previous behavior and avoids the empty-string regression. If the intent is to suppress the owner in no-auth mode, document that explicitly and ensure the flow handles `""` gracefully.

### Alternative Solutions

- Keep `None` for no-auth `owner` but change `chat_service.py` to omit the header entirely when `owner` is `None` rather than sending an empty string. This ensures the header is absent rather than blank:
  ```python
  extra_headers = {}
  if owner:
      extra_headers["X-Langflow-Global-Var-OWNER"] = str(owner)
  if owner_name:
      extra_headers["X-Langflow-Global-Var-OWNER_NAME"] = str(owner_name)
  if owner_email:
      extra_headers["X-Langflow-Global-Var-OWNER_EMAIL"] = str(owner_email)
  ```

src/api/settings.py
```diff
@@ -1455,40 +1455,17 @@ async def _run_async_post_save_langflow_updates(
         logger.error(f"Failed to update Langflow settings asynchronously: {str(e)}")
 
 
-async def _update_mcp_servers_with_provider_credentials(config, session_manager = None, flows_service=None):
-    # Update MCP servers with provider credentials
+async def _update_mcp_server_urls(config, session_manager=None, flows_service=None):
```
mpawlow, line 1458: ## (b) [Normal] `_update_mcp_server_urls` in `settings.py` accepts parameters it never uses

### Confidence

97

### Summary

The new `_update_mcp_server_urls` function accepts `config`, `session_manager`, and `flows_service` parameters but uses none of them. All call sites still pass these arguments, creating misleading dead code in the function signature.

### Problem

- The function signature is:
  ```python
  async def _update_mcp_server_urls(config, session_manager=None, flows_service=None):
  ```
- The function body does not reference `config`, `session_manager`, or `flows_service` at all — it only instantiates `LangflowMCPService()` and calls `update_all_mcp_server_urls()`.
- Callers such as `onboarding`, `_run_async_post_save_langflow_updates`, and `reapply_all_settings` continue to pass these arguments, giving the false impression that they influence behavior.

### Code References

- `src/api/settings.py` line 1458 (function definition with unused params)
- `src/api/settings.py` line 1114 (caller in `onboarding` still passes `session_manager`)
- `src/api/settings.py` line 1434 (caller in `_run_async_post_save_langflow_updates`)
- `src/api/settings.py` line 1629 (caller in `reapply_all_settings`)

### Potential Solution

Remove unused parameters from the function signature and update all call sites:

```python
async def _update_mcp_server_urls():
    """Update MCP server URLs (patch localhost and convert to streamable HTTP)."""
    try:
        from services.langflow_mcp_service import LangflowMCPService
        mcp_service = LangflowMCPService()
        result = await mcp_service.update_all_mcp_server_urls()
        logger.info("Updated MCP server URLs after settings change", **result)
    except Exception as mcp_error:
        logger.warning(f"Failed to update MCP server URLs after settings change: {str(mcp_error)}")
```

Then update all call sites to simply `await _update_mcp_server_urls()`.

src/services/langflow_mcp_service.py
```diff
@@ -218,12 +119,8 @@ def _is_convertible_to_streamable_http(self, server_config: Dict[str, Any]) -> b
         for arg in args:
             if not isinstance(arg, str):
                 continue
-            if not (arg.startswith("http://") or arg.startswith("https://")):
-                continue
-            # Check URL path endings
-            if arg.rstrip("/").endswith("/mcp") or arg.rstrip("/").endswith(
-                "/streamable"
-            ):
+            if arg.startswith("http://") or arg.startswith("https://"):
```
mpawlow, line 123: ## (c) [Normal] `_is_convertible_to_streamable_http` conversion criteria is now too broad

### Summary

The previous implementation required that a stdio server's URL argument end in `/mcp` or `/streamable` before converting it to streamable HTTP mode. The new implementation converts any stdio server containing any `http://` or `https://` URL argument. This risks incorrectly converting stdio servers that carry a URL in args for a different purpose.

### Problem

- MCP stdio server args may legitimately carry HTTP URLs as target parameters (e.g., a fetch tool, a web scraping server) that are not MCP HTTP endpoints. Converting those to streamable HTTP mode would replace the entire server configuration with `{"url": <that_URL>}`, likely breaking the server.
- The comment "We convert any stdio server with a URL to streamable HTTP" confirms this is intentional but does not address the risk of false positives.

### Potential Solution

Restore a path-suffix check as an additional guard, or limit conversion to URLs that contain `/mcp` or `/api/v2/mcp`:

```python
MCP_URL_PATTERNS = ("/mcp", "/streamable", "/api/v2/mcp")

for arg in args:
    if not isinstance(arg, str):
        continue
    if arg.startswith("http://") or arg.startswith("https://"):
        if any(arg.rstrip("/").endswith(pat) or pat in arg for pat in MCP_URL_PATTERNS):
            return True
```

tests/integration/core/test_mcp_url_ingest.py
```diff
@@ -0,0 +1,156 @@
...
+    response = await clients.langflow_request(
+        "GET",
+        f"/api/v1/flows/{LANGFLOW_CHAT_FLOW_ID}",
+    )
+    response.raise_for_status()
+    flow_text = json.dumps(response.json())
+
+    assert "opensearch_url_ingestion_flow" in flow_text
+    assert OPENRAG_MCP_SERVER_NAME in flow_text
+    assert "DynamicCreateData-ElAqI" in flow_text
```
mpawlow: ## (d) [Major] Integration test hard-codes internal Langflow flow node IDs

### Summary

`test_loaded_agent_flow_routes_request_globals_into_mcp_headers` asserts specific auto-generated Langflow node IDs (`DynamicCreateData-ElAqI`, `MCP-k64rA`) and an exact edge assertion. These will produce false failures whenever the flow JSON is regenerated, even if the flow's behavior is correct.

### Problem

- The test asserts:
  ```python
  assert "DynamicCreateData-ElAqI" in flow_text
  assert "MCP-k64rA" in flow_text
  assert '"target": "MCP-k64rA"' in flow_text
  assert '"fieldName": "headers"' in flow_text
  ```
- Langflow node IDs (e.g., `MCP-k64rA`, `DynamicCreateData-ElAqI`) are auto-generated and can change whenever a flow is re-saved or exported from the Langflow UI.
- The test intent appears to be verifying that the agent flow correctly routes global variables into MCP headers — the hardcoded IDs make it a brittle structural check rather than a behavioral one.
- If the `openrag_agent.json` flow is updated (e.g., to use a different Langflow component type), all these assertions break even if the new flow has equivalent behavior.

### Potential Solution

Test the behavior rather than the internal structure. Instead of asserting specific node IDs, assert the behavioral properties:

```python
# Assert that some node connects to an MCP server component via a "headers" field
assert '"fieldName": "headers"' in flow_text
assert OPENRAG_MCP_SERVER_NAME in flow_text

# Assert that global variable dynamic inputs are wired up
for global_var_name in [...]:
    assert f"dynamic_X-Langflow-Global-Var-{global_var_name}" in flow_text
```

Drop the node-ID assertions (`DynamicCreateData-ElAqI`, `MCP-k64rA`, and the edge target check) entirely. If specific component connectivity must be verified, derive the IDs dynamically from the flow's node list rather than hardcoding them.

src/services/langflow_mcp_service.py
```diff
@@ -466,15 +239,16 @@ async def patch_mcp_server_args_with_jwt(
...
-                "Exception while patching MCP server",
+                "Exception while patching MCP server URL",
                 server_name=server_name,
                 error=str(e),
             )
             return False
 
-    async def update_mcp_servers_with_jwt(self, jwt_token: str) -> Dict[str, Any]:
-        """Fetch all MCP servers and ensure each includes the JWT header in args.
+    async def update_all_mcp_server_urls(self) -> Dict[str, Any]:
```
mpawlow, line 251: ## (e) [Minor] "Updated" metric in `update_all_mcp_server_urls` conflates patched and skipped servers

### Summary

`update_all_mcp_server_urls` counts servers where `patch_mcp_server_url` returned `True` as "updated", but `patch_mcp_server_url` returns `True` for both "successfully patched" and "nothing to do / skipped" cases. The logged summary is therefore misleading.

### Problem

- `patch_mcp_server_url` returns `True` in at least three "skip" scenarios:
  - Could not extract URL from stdio args
  - URL already points to `LANGFLOW_URL` (unchanged)
  - No URL found in args to patch
- All these `True` returns are counted in the `updated` tally in `update_all_mcp_server_urls`:
  ```python
  ok = await self.patch_mcp_server_url(name)
  if ok:
      updated += 1   # "ok" means both "patched" and "skipped"
  ```
- The log `"MCP servers URL update completed", updated=N` implies N servers were modified, but N actually means "N servers were processed without error."

### Code References

- `src/services/langflow_mcp_service.py` lines 248–275 (`update_all_mcp_server_urls`)
- `src/services/langflow_mcp_service.py` lines 155–246 (`patch_mcp_server_url` — `return True` at lines 182, 192–193, 213–214)

### Potential Solution

Use a tri-state return value or a tuple to distinguish patched, skipped, and failed:

```python
async def patch_mcp_server_url(self, server_name: str) -> str:
    """Returns 'patched', 'skipped', or 'failed'."""
    ...
```

Then aggregate counts separately in `update_all_mcp_server_urls`:

```python
summary = {"patched": patched, "skipped": skipped, "failed": failed, "total": len(servers)}
```

mpawlow (changes requested):
### Code Review 1

- See PR comments: (a) to (e)

src/services/langflow_mcp_service.py
```diff
@@ -285,162 +153,70 @@ def _patch_url_with_langflow_url(self, url: str) -> str:
...
-                    updated_args = self._upsert_global_var_headers_in_args(
-                        args, sanitized
+                    # Cannot extract URL — leave as-is
+                    logger.debug(
+                        "Could not extract URL from stdio args, skipping",
+                        server_name=server_name,
                     )
-                    payload = {"command": command, "args": updated_args}
-                    mode = "stdio"
+                    return True
```
mpawlow: ## (2a) [Normal] Two early-exit "skip" branches in `patch_mcp_server_url` return `True` instead of `"skipped"` - counted as failures

### Summary

The return-value refactoring in `patch_mcp_server_url` (from `bool` to `"patched"` / `"skipped"` / `"failed"` strings) was applied inconsistently. Two early-exit branches that represent "nothing to do" still return `True` (the old boolean), causing `update_all_mcp_server_urls` to classify them as failures in the summary metrics.

### Problem

- `patch_mcp_server_url` has three early-exit "skip" paths:
  1. **Cannot extract URL from stdio args** (line 185): returns `True`
  2. **URL already matches LANGFLOW_URL — nothing to patch** (line 196): returns `True`
  3. **No URL to patch in stdio args** (line 217): returns `"skipped"` ✓
- In `update_all_mcp_server_urls`, the aggregation logic is:
  ```python
  ok = await self.patch_mcp_server_url(name)
  if ok == "patched":
      patched_count += 1
  elif ok == "skipped":
      skipped_count += 1
  else:
      failed_count += 1   # True falls here — counted as failure!
  ```
- `True == "patched"` is `False` and `True == "skipped"` is `False`, so both early-exit `True` returns land in the `else` branch and are counted as `failed`.
- The logged summary will show spurious failure counts for servers whose URLs are already correct or whose args cannot be parsed — misleading operators and potentially triggering alerts.

### Code References

- `src/services/langflow_mcp_service.py` line 185 (returns `True` — should be `"skipped"`)
- `src/services/langflow_mcp_service.py` line 196 (returns `True` — should be `"skipped"`)
- `src/services/langflow_mcp_service.py` lines 265–270 (`update_all_mcp_server_urls` aggregation)

### Potential Solution

Change both `return True` early-exit branches to `return "skipped"`:

```python
# Line 185 area — cannot extract URL from stdio args
return "skipped"

# Line 196 area — URL already points to LANGFLOW_URL, nothing to do
return "skipped"
```

src/services/langflow_mcp_service.py
```diff
@@ -285,162 +153,70 @@ def _patch_url_with_langflow_url(self, url: str) -> str:
             logger.debug(f"Patched URL: {url}")
         return url
 
-    async def patch_mcp_server_args_with_global_vars(
-        self, server_name: str, global_vars: Dict[str, Any]
-    ) -> bool:
-        """Patch a single MCP server to include/update multiple X-Langflow-Global-Var-* headers.
+    async def patch_mcp_server_url(self, server_name: str) -> bool:
```
mpawlow: ## (2b) [Minor] `patch_mcp_server_url` return type annotation `-> bool` contradicts its string return values

### Summary

The method signature declares `-> bool` but the function now returns string literals `"patched"`, `"skipped"`, or `"failed"`. Static type checkers will flag all callers and return sites, and the docstring's "Returns 'patched', 'skipped', or 'failed'" directly contradicts the annotation.

### Problem

- Method signature:
  ```python
  async def patch_mcp_server_url(self, server_name: str) -> bool:
      """...
      Returns 'patched', 'skipped', or 'failed'.
      """
  ```
- The `-> bool` annotation was not updated when the return type was changed from boolean to string literals.
- Any mypy / pyright / IDE type check run against this code will produce errors, and `True` values returned by the two un-refactored early-exit paths (issue 2a) will appear to satisfy the `-> bool` annotation, masking the bug until runtime.

### Potential Solution

Update the return type annotation to `Literal["patched", "skipped", "failed"]` (requires `from typing import Literal`):

```python
from typing import Literal

async def patch_mcp_server_url(self, server_name: str) -> Literal["patched", "skipped", "failed"]:
```

Or use `str` as a simpler alternative:

```python
async def patch_mcp_server_url(self, server_name: str) -> str:
```

mpawlow (approved):
### Code Review 2

- ✅  Approved / LGTM  🚀 
- See PR comments: (2a) and (2b) for potential post-evaluation

PR #1481, opened 2026-04-27
fix: Ensure the list of files deleted doesn't clear before delete finished

mfortman11:
Fixes https://github.ibm.com/lakehouse/tracker/issues/70731

Bulk delete — setSelectedRows([]) and gridRef.current?.api.deselectAll() moved into the finally block so selection clears atomically with the dialog closing, preventing a flash of empty content.

Per-row delete — replaced deleteDocumentMutation.isPending with a dedicated isDeleting state and consolidated dialog close into finally, matching the bulk pattern.

### DIFF
--- a/frontend/app/knowledge/page.tsx
+++ b/frontend/app/knowledge/page.tsx
@@ -629,12 +629,6 @@ function SearchPage() {
           } not deleted (0 chunks matched).`,
         );
       }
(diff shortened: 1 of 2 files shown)


Wallgau (approved):
Nice! LGTM!

PR #1482, opened 2026-04-27
fix: sync ingest embedding from settings and persist ingest UI

Wallgau:
issue: https://github.ibm.com/lakehouse/tracker/issues/70634
In ingest-settings, the child both rendered the form and ran an effect that watched server data (settings / model list). When that data loaded or changed, the effect called onSettingsChange to push the parent’s embeddingModel toward whatever the API said
So the embedding field was not only “what the user picked”, it was continuously reconciled from below against async query results.
Changes to have:
One owner of the truth: The page / hook that holds useState for ingest settings decides when to apply workspace defaults (e.g. only while still on a placeholder, or after persistence loads). The form component becomes controlled: it displays settings and emits changes; it does not sneak in server-driven patches through an effect.

Clear product rules: Rules like “seed from knowledge.embedding_model once” or “do not override after the user or localStorage set a value” live in one place (useIngestEmbeddingFromSettings + persistence), not inside every field component.

IngestSettings can still fetch settings and model lists to build the dropdown (labels, provider, options). That is read-only UI support data. The important split is: no effect that mutates parent state from queries that sync belongs next to the state it updates.


lucaseduoli (changes requested):
After changing some settings in the settings page, it does not reflect on the Sharepoint settings:

Image

Image

Wallgau:
> After changing some settings in the settings page, it does not reflect on the Sharepoint settings:
> 
> Image Image

so those 2 settings should reflect? I assumed not because why having 2 settings place then... maybe the user wants to set up the connector differently? Or should we always override from the knowledge settings?

lucaseduoli (changes requested):
Image

Image

Even after changing the embedding model, it ingests with the previous model

Wallgau:
created a new PR for this issue

PR #1483, opened 2026-04-27
fix: set default ibm model

mfortman11:
sets the default ibm model to one that works w/ openrag tools


src/config/model_constants.py
```diff
@@ -35,3 +35,5 @@
 ANTHROPIC_DEFAULT_LANGUAGE_MODEL = "claude-sonnet-4-5-20250929"
 
 OLLAMA_DEFAULT_LANGUAGE_MODEL_PATTERN = "gpt-oss"
+
+WATSONX_DEFAULT_LANGUAGE_MODEL = "openai/gpt-oss-120b"
```
mpawlow, line 39: ## (a) [Major] WatsonX Default Model Bypasses Provider Routing in `get_litellm_model_name`

### Summary

The constant `WATSONX_DEFAULT_LANGUAGE_MODEL = "openai/gpt-oss-120b"` contains an `openai/` prefix that causes `get_litellm_model_name` to skip provider-based routing entirely, potentially sending WatsonX requests to the OpenAI API instead.

### Problem

- `get_litellm_model_name` has an early-return guard:
  ```python
  KNOWN_PREFIXES = ["openai", "ollama", "watsonx", "anthropic"]

  if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
      return model_name
  ```
- Because `"openai/gpt-oss-120b"` starts with `"openai/"`, the function immediately returns `"openai/gpt-oss-120b"` — it never consults the model registry, and never prepends the `"watsonx/"` prefix that would route the call through WatsonX
- LiteLLM interprets `"openai/gpt-oss-120b"` as an OpenAI-provider model and routes it to the OpenAI API, not the configured WatsonX endpoint
- This behaviour is inconsistent with how other WatsonX models are handled (e.g. `"ibm/granite-..."` would correctly resolve to `"watsonx/ibm/granite-..."`)

### Code References

- `src/config/model_constants.py:39` — `WATSONX_DEFAULT_LANGUAGE_MODEL = "openai/gpt-oss-120b"`
- `src/services/models_service.py:117–118` — prefix early-return in `get_litellm_model_name`
- `src/services/models_service.py:488` — comparison `model_id == WATSONX_DEFAULT_LANGUAGE_MODEL`
- `src/services/search_service.py:203` — call site where `get_litellm_model_name` is invoked without an explicit provider

### Potential Solution

- Verify the exact `model_id` string returned by the WatsonX `foundation_model_specs` API for this model, and use that raw ID (without any provider prefix) as the constant value:
  ```python
  WATSONX_DEFAULT_LANGUAGE_MODEL = "gpt-oss-120b"  # raw Watson model_id
  ```
- Update `get_litellm_model_name` to check the registry first before applying the early-return prefix guard, so that models registered under `"watsonx"` always receive the correct `"watsonx/"` prefix regardless of their internal ID structure

### Alternative Solutions

- **Keep the constant as-is but fix `get_litellm_model_name`** — extend the early-return logic to verify the prefix matches the registered provider; if the registry says `"watsonx"` but the prefix is `"openai"`, override and format as `"watsonx/openai/gpt-oss-120b"`:
   ```python
   if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
       registered_provider = ModelsService._model_provider_registry.get(model_name)
       if registered_provider and not model_name.startswith(registered_provider + "/"):
           return f"{registered_provider}/{model_name}"
       return model_name
   ```

src/services/models_service.py
```diff
@@ -521,9 +522,11 @@ async def get_ibm_models(
                             {
                                 "value": model_id,
                                 "label": model_name or model_id,
-                                "default": i == 0,  # First model is default
+                                "default": model_id == WATSONX_DEFAULT_LANGUAGE_MODEL,
```
mpawlow, line 533: ## (b) [Normal] No Fallback Default When `WATSONX_DEFAULT_LANGUAGE_MODEL` Is Absent from API Response

### Summary

If the Watson API does not return `"openai/gpt-oss-120b"` (e.g., model retired, renamed, or unavailable on the account), no language model will have `"default": True`, leaving the UI with an unselected language model for all IBM Watson users.

### Problem

- Previously, `"default": i == 0` guaranteed that exactly one model was always marked as default (the first one returned)
- The new code, `"default": model_id == WATSONX_DEFAULT_LANGUAGE_MODEL`, produces zero defaults if the specified model ID does not appear in the API response
- There is no warning log or fallback to indicate this state has occurred
- The `"default"` flag is consumed downstream to pre-populate the model selector in the UI; an empty default state may require users to manually choose a model every time, or could surface as a confusing blank selection

### Code References

- `src/services/models_service.py:484–489` — the new language model default assignment
- `src/services/models_service.py:492` — sort after loop (sort is no-op if no default is set)
- `src/config/model_constants.py:39` — the constant being matched against

### Potential Solution

Add a fallback after the loop: if no model was marked as default, fall back to the first model and emit a warning:

```python
language_models.sort(key=lambda x: (not x.get("default", False), x["value"]))

if language_models and not any(m.get("default") for m in language_models):
    logger.warning(
        "WATSONX_DEFAULT_LANGUAGE_MODEL '%s' not found in API response; "
        "falling back to first available model as default.",
        WATSONX_DEFAULT_LANGUAGE_MODEL,
    )
    language_models[0]["default"] = True
```

### Alternative Solutions

1. **Raise an explicit warning only** (no fallback) — log clearly that the configured default model was not found, allowing the caller to decide how to handle an unset default, rather than silently applying one
2. **Validate the model ID at startup** — when WatsonX credentials are first configured, fetch and cache the list of available model IDs and verify `WATSONX_DEFAULT_LANGUAGE_MODEL` is in the list; fail loudly at configuration time rather than silently at runtime

src/services/models_service.py
```diff
@@ -547,7 +550,7 @@ async def get_ibm_models(
                     embed_models = embed_data.get("resources", [])
                     logger.info(f"Retrieved {len(embed_models)} embedding models from Watson API")
 
-                    for i, model in enumerate(embed_models):
+                    for model in embed_models:
```
mpawlow: ## (c) [Minor] IBM Embedding Models Have No Default — Inconsistent with Other Providers

- Note: This is Minor. Please feel free to optionally implement or ignore.

### Summary

All IBM Watson embedding models are now unconditionally set to `"default": False`. Unlike OpenAI (which has `OPENAI_DEFAULT_EMBEDDING_MODEL = "text-embedding-3-small"`), there is no `WATSONX_DEFAULT_EMBEDDING_MODEL` constant, leaving WatsonX embedding model selection entirely up to the user.

### Problem

- Before this PR, the first embedding model in the WatsonX API response was marked as default (`"default": i == 0`)
- After this PR, no embedding model ever has `"default": True`
- There is no `WATSONX_DEFAULT_EMBEDDING_MODEL` constant in `model_constants.py` or `embedding_constants.py`, and no comment explaining the intentional omission
- This is asymmetric with `get_openai_models` (which uses `OPENAI_DEFAULT_EMBEDDING_MODEL`) and may degrade UX for Watson users who need to embed documents, as they must manually select an embedding model where they previously did not

### Code References

- `src/config/embedding_constants.py:3` — `OPENAI_DEFAULT_EMBEDDING_MODEL = "text-embedding-3-small"` (no WatsonX equivalent)

### Potential Solution

Define a `WATSONX_DEFAULT_EMBEDDING_MODEL` constant in `model_constants.py` or `embedding_constants.py` and apply the same explicit-match pattern used for language models:

```python
# model_constants.py
WATSONX_DEFAULT_EMBEDDING_MODEL = "<verified-watson-embedding-model-id>"
```

```python
# models_service.py — embedding loop
embedding_models.append({
    "value": model_id,
    "label": model_name or model_id,
    "default": model_id == WATSONX_DEFAULT_EMBEDDING_MODEL,
})
```

### Alternative Solutions

1. **Keep `"default": False` but document it** — add a comment stating that no default embedding model is defined for WatsonX because no single model is broadly recommended, directing users to choose explicitly; this makes the omission intentional and visible
2. **Restore the first-model fallback for embeddings only** — since an embedding default is purely cosmetic (pre-populates the UI), reverting to `"default": i == 0` for embeddings avoids regression while the team decides on a canonical default

mpawlow (changes requested):
## Code Review 1

- See PR comments: (a) to (c)

src/services/models_service.py
```diff
@@ -132,8 +133,13 @@ async def get_litellm_model_name(
         if not model_name:
             return ""
 
-        # Skip formatting if already has a known provider prefix
+        # If the model name already has a known provider prefix, trust the registry
+        # over the prefix — e.g. "openai/gpt-oss-120b" registered under "watsonx"
+        # must be routed as "watsonx/openai/gpt-oss-120b", not sent to OpenAI.
         if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
+            registered_provider = ModelsService._model_provider_registry.get(model_name)
```
mpawlow, line 140: ## (2a) [Normal] Cold-Start Registry Miss Leaves Prefixed Models Incorrectly Routed

### Summary

The routing fix for `get_litellm_model_name` only works when the model registry is already populated. On a cold start (empty registry), `"openai/gpt-oss-120b"` is still returned unchanged, causing it to be routed to OpenAI rather than WatsonX.

### Problem

- The updated prefix guard path reads:
  ```python
  if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
      registered_provider = ModelsService._model_provider_registry.get(model_name)
      if registered_provider and not model_name.startswith(registered_provider + "/"):
          return f"{registered_provider}/{model_name}"
      return model_name
  ```
- When the registry is empty (e.g. first call after service startup), `registered_provider` is `None`
- The condition `if registered_provider and ...` is `False`, so execution falls through to `return model_name` — returning `"openai/gpt-oss-120b"` unchanged
- This causes LiteLLM to route the call to the OpenAI API, not WatsonX
- By contrast, the non-prefix path that follows already handles cold starts correctly:
  ```python
  if provider_lower is None:
      provider_lower = ModelsService._model_provider_registry.get(model_name)
      if provider_lower is None:
          await self.update_model_registry()   # ← triggers a refresh
          provider_lower = ModelsService._model_provider_registry.get(model_name)
  ```
- The prefix path never calls `update_model_registry()`, creating an asymmetry: non-prefixed models get an automatic registry refresh on cache miss; prefixed models do not

### Code References

- `src/services/models_service.py:148–153` — non-prefix path with registry refresh (the correct pattern to follow)

### Potential Solution

Mirror the registry-refresh pattern from the non-prefix path inside the prefix guard:

```python
if any(model_name.startswith(p + "/") for p in KNOWN_PREFIXES):
    registered_provider = ModelsService._model_provider_registry.get(model_name)
    if registered_provider is None:
        await self.update_model_registry()
        registered_provider = ModelsService._model_provider_registry.get(model_name)
    if registered_provider and not model_name.startswith(registered_provider + "/"):
        return f"{registered_provider}/{model_name}"
    return model_name
```

### Alternative Solutions

1. **Eagerly populate the registry at WatsonX credential configuration time** — call `update_model_registry()` immediately when WatsonX credentials are saved rather than lazily on the first model-name lookup; this eliminates cold-start gaps for all code paths, not just the prefix guard
2. **Use the explicit `provider` parameter at call sites** — where the WatsonX provider is already known (e.g. when resolving a model stored in the WatsonX settings), pass `provider="watsonx"` explicitly to `get_litellm_model_name`; the explicit-provider path bypasses the registry entirely and doesn't suffer from cold-start issues

mpawlow (approved):
## Code Review 2

- ✅  Approved / LGTM  🚀 
- See PR comment: (2a)

PR #1489, opened 2026-04-28
fix: remove mcp url ingest code path

phact:


phact:
probably close in favor of https://github.com/langflow-ai/openrag/pull/1474

lucaseduoli:
@phact I think we can close this one

PR #1492, opened 2026-04-28
fix: sync ingest embedding from settings and persist ingest UI

Wallgau:
Summary
Hydrate connector ingest settings from GET /api/settings knowledge values (for existing ingest UI fields only) and keep them session-owned via useSessionIngestSettings.
Make cloud picker ingest settings controlled from upload page, and fix embedding model select rendering so user-selected models are always reflected (even if not in fetched options list yet).
Ensure connector sync respects per-request ingest settings end-to-end in both Langflow and non-Langflow paths, with connector-only embedding model override behavior and reusable UI-settings→tweaks mapping.
Key changes
Frontend ingest state + UX

Added frontend/lib/ingest-settings-knowledge.ts to map knowledge settings to ingest panel shape.
Added frontend/hooks/useSessionIngestSettings.ts to hydrate once from settings and preserve in-session edits.
Updated frontend/app/upload/[provider]/page.tsx to use controlled ingest settings and improved sync error handling.
Updated frontend/components/cloud-picker/unified-cloud-picker.tsx + types.ts for controlled ingest props (ingestSettings, onIngestSettingsChange).
Updated frontend/components/cloud-picker/ingest-settings.tsx to avoid clobbering user model selection and to ensure current selected model is always represented in Select options.
Backend connector ingest propagation

Added optional settings to connector sync API request model and passed through router/service layers:
src/api/connectors.py
src/api/connector_router.py
src/connectors/service.py
src/connectors/langflow_connector_service.py
src/models/processors.py
Non-Langflow path now consumes connector ingest settings in ConnectorFileProcessor:
applies embeddingModel
applies chunkSize/chunkOverlap through TaskProcessor.process_document_standard(...).
Added resplit_chunks_character_windows helper for non-Langflow chunk sizing in src/utils/document_processing.py.
Langflow-specific behavior

Added LangflowFileService.merge_ui_ingest_settings_into_tweaks(...) with documentation.
Connector path strips embeddingModel from tweaks and passes it via selected_embedding_model to avoid conflict with global header override.
run_ingestion_flow now prioritizes:
explicit selected_embedding_model
OpenAI embeddings tweak model
configured knowledge embedding model

Uploading Screen Recording 2026-04-27 at 10.55.20 PM.mov…

### DIFF
--- a/src/services/langflow_file_service.py
+++ b/src/services/langflow_file_service.py
@@ -42,14 +42,11 @@ class LangflowFileService:
 
         - ``chunkSize`` / ``chunkOverlap`` / ``separator`` update the flow's
           ``SplitText-QIKhg`` node when any of those keys are present.
-        - ``embeddingModel`` maps to ``OpenAIEmbeddings-joRJ6`` ``model`` for
-          flows that use that OpenAI embeddings component id.
-
-        **Connector ingest:** callers should remove ``embeddingModel`` from
-        ``settings`` before merging and pass the user's model via
-        ``run_ingestion_flow(..., selected_embedding_model=...)`` instead, so
-        the global Langflow header override does not fight a conflicting
-        embedding tweak.
+        - ``embeddingModel`` is intentionally not mapped to a component tweak.
+          The embedding model should be supplied via
+          ``run_ingestion_flow(..., selected_embedding_model=...)`` so Langflow
+          resolves it through the global variable override, without relying on
+          provider-specific component ids.
         """
         final_tweaks = dict(tweaks) if tweaks else {}
         if not settings:
@@ -70,13 +67,6 @@ class LangflowFileService:
                 ]
             if settings.get("separator"):
                 final_tweaks["SplitText-QIKhg"]["separator"] = settings["separator"]
-
-        if settings.get("embeddingModel"):
-            if "OpenAIEmbeddings-joRJ6" not in final_tweaks:
-                final_tweaks["OpenAIEmbeddings-joRJ6"] = {}
-            final_tweaks["OpenAIEmbeddings-joRJ6"]["model"] = settings[
-                "embeddingModel"
-            ]
 
         return final_tweaks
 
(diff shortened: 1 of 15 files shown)


src/services/langflow_file_service.py
```diff
@@ -33,6 +33,53 @@ def _is_transient_request_error(error: Exception) -> bool:
...
+            if settings.get("chunkSize"):
+                final_tweaks["SplitText-QIKhg"]["chunk_size"] = settings["chunkSize"]
+            if settings.get("chunkOverlap"):
+                final_tweaks["SplitText-QIKhg"]["chunk_overlap"] = settings[
+                    "chunkOverlap"
+                ]
+            if settings.get("separator"):
+                final_tweaks["SplitText-QIKhg"]["separator"] = settings["separator"]
+
+        if settings.get("embeddingModel"):
```
lucaseduoli: embedding model shouldn't be hard coded to this OpenAIEmbeddings. You should just set the SELECTED_EMBEDDING_MODEL variable

src/services/langflow_file_service.py
```diff
@@ -156,6 +204,12 @@ async def run_ingestion_flow(
         
         config = get_openrag_config()
         embedding_model = config.knowledge.embedding_model
+        if selected_embedding_model:
+            embedding_model = selected_embedding_model
+        else:
+            openai_tweak = tweaks.get("OpenAIEmbeddings-joRJ6") if isinstance(tweaks, dict) else None
```
lucaseduoli: Don't hard code the OpenAIEmbeddings, use the env variable if it works

lucaseduoli: The old code is outdated, you can remove the part of the tweaks that reference that openai embedding

lucaseduoli (approved):
It's working! Only thing I noticed is that I cant select the text-embedding-3-large while the small is the default, but I think that can be resolved afterwards. Please fix the nitpicks and test again to see if they didn't break anything

PR #1504, opened 2026-04-28
fix: Unable to overwrite documents in Knowledge

mpawlow:
### Issue

- #70921

### Summary

- Fixed a bug where overwriting an existing document in Knowledge failed because the query cache invalidation assumed a flat array format but the search results now use a `SearchResult` object structure with a `files` property.

### Frontend

- Updated `queryClient.setQueriesData` callback in `knowledge-dropdown.tsx` to handle the `SearchResult` structure `{ files: [], warnings: [] }` when filtering out the file being overwritten
- Added a fallback path to handle the legacy flat array format for backwards compatibility
- Changed the `oldData` type annotation from `[]` to `any` to accommodate both data shapes

### DIFF
--- a/frontend/components/knowledge-dropdown.tsx
+++ b/frontend/components/knowledge-dropdown.tsx
@@ -393,12 +393,24 @@ export function KnowledgeDropdown() {
 
     if (pendingFile) {
       // Remove the old file from all search query caches before overwriting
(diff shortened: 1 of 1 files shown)


lucaseduoli (approved):
perfect

mfortman11 (approved):

PR #1507, opened 2026-04-29
fix: adjust patch function for mcp

lucaseduoli:
This pull request updates the return value of the `patch_mcp_server_url` method in `src/services/langflow_mcp_service.py` to provide more informative status feedback. Instead of returning a boolean, the method now returns a string status, such as `"skipped"`, to clarify the outcome of the patch operation.

Patch method return value update:

* Changed the return type of `patch_mcp_server_url` from `bool` to `str` to allow returning descriptive status messages instead of just `True`/`False`.
* Updated return values within `patch_mcp_server_url` to return `"skipped"` instead of `True` when the patch operation is not performed, making the method's outcome more explicit and easier to interpret. [[1]](diffhunk://#diff-5709c20045ba762b6893b80b5944afefabf8f6d62ec3473a595db44daf8c032fL185-R185) [[2]](diffhunk://#diff-5709c20045ba762b6893b80b5944afefabf8f6d62ec3473a595db44daf8c032fL196-R196)

### DIFF
--- a/src/services/langflow_mcp_service.py
+++ b/src/services/langflow_mcp_service.py
@@ -153,7 +153,7 @@ class LangflowMCPService:
             logger.debug(f"Patched URL: {url}")
         return url
 
-    async def patch_mcp_server_url(self, server_name: str) -> bool:
+    async def patch_mcp_server_url(self, server_name: str) -> str:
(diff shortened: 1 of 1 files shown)


mpawlow (approved):
### Code Review 1

- ✅  Approved / LGTM  🚀

PR #1512, opened 2026-04-30
feat: Add error handling for failed file entries in TaskNotificationMenu

ricofurtado:
This pull request primarily improves the handling and display of task errors in the UI and updates TypeScript and package configuration for better compatibility and stricter typing. The most significant changes include surfacing failed file entries in the task notification menu, updating the TypeScript configuration to use the latest JSX transform, and marking several dependencies as peer dependencies in the lockfile for improved package management.

**UI Improvements:**

* The `TaskNotificationMenu` component now displays detailed error information for tasks with failed file entries by rendering the `TaskErrorContent` component when applicable.

**TypeScript and Build Configuration:**

* Updated `tsconfig.json` to use `"jsx": "react-jsx"` for the new JSX transform, and reformatted the file for better readability. 

## Summary by CodeRabbit

* **New Features**
  * Active Tasks now display error details inline for tasks with failures; error panels are visible and can be expanded directly within the task list.

* **Chores**
  * Frontend TypeScript config switched to the newer JSX transform to improve JSX handling and build consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1512)

### DIFF
--- a/frontend/components/task-notification-menu.tsx
+++ b/frontend/components/task-notification-menu.tsx
@@ -387,6 +387,11 @@ export function TaskNotificationMenu() {
                               mode="recent"
                               showHeader={false}
                               defaultExpanded={true}
+                              expandTrigger={
+                                selectedTaskId === task.task_id
+                                  ? selectedTaskTrigger
+                                  : 0
+                              }
                             />
                           </div>
                         )}
--- a/frontend/components/task-notification-menu.tsx
+++ b/frontend/components/task-notification-menu.tsx
@@ -380,6 +380,16 @@ export function TaskNotificationMenu() {
                             </Button>
                           </div>
                         )}
+                        {hasFailedFileEntries(task) && (
+                          <div className="mt-3">
+                            <TaskErrorContent
+                              task={task}
+                              mode="recent"
+                              showHeader={false}
+                              defaultExpanded={true}
+                            />
+                          </div>
+                        )}
                       </CardContent>
                     )}
                   </Card>
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -1,11 +1,7 @@
 {
   "compilerOptions": {
     "target": "ES2017",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
     "strict": true,
@@ -23,9 +19,7 @@
       }
     ],
     "paths": {
-      "@/*": [
-        "./*"
-      ]
+      "@/*": ["./*"]
     }
   },
   "include": [
@@ -35,7 +29,5 @@
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],
-  "exclude": [
-    "node_modules"
-  ]
+  "exclude": ["node_modules"]
 }
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -1,7 +1,11 @@
 {
   "compilerOptions": {
     "target": "ES2017",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
     "allowJs": true,
     "skipLibCheck": true,
     "strict": true,
@@ -11,7 +15,7 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
     "incremental": true,
     "plugins": [
       {
@@ -19,7 +23,9 @@
       }
     ],
     "paths": {
-      "@/*": ["./*"]
+      "@/*": [
+        "./*"
+      ]
     }
   },
   "include": [
@@ -29,5 +35,7 @@
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],
-  "exclude": ["node_modules"]
+  "exclude": [
+    "node_modules"
+  ]
 }
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1533,6 +1533,7 @@
       "integrity": "sha512-6TyEnHgd6SArQO8UO2OMTxshln3QMWBtPGrOCgs3wVEmQmwyuNtB10IZMfmYDE0riwNR1cu4q+pPcxMVtaG3TA==",
       "devOptional": true,
       "license": "Apache-2.0",
+      "peer": true,
       "dependencies": {
         "playwright": "1.57.0"
       },
@@ -2903,6 +2904,7 @@
       "integrity": "sha512-yCAeZl7a0DxgNVteXFHt9+uyFbqXGy/ShC4BlcHkoE0AfGXYv/BUiplV72DjMYXHDBXFjhvr6DD1NiRVfB4j8g==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "undici-types": "~6.21.0"
       }
@@ -2918,6 +2920,7 @@
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.1.12.tgz",
       "integrity": "sha512-cMoR+FoAf/Jyq6+Df2/Z41jISvGZZ2eTlnsaJRptmZ76Caldwy1odD4xTr/gNV9VLj0AWgg/nmkevIyUfIIq5w==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "csstype": "^3.0.2"
       }
@@ -2928,6 +2931,7 @@
       "integrity": "sha512-qXRuZaOsAdXKFyOhRBg6Lqqc0yay13vN7KrIg4L7N4aaHN68ma9OK3NE1BoDFgFOTfM7zg+3/8+2n8rLUH3OKQ==",
       "devOptional": true,
       "license": "MIT",
+      "peer": true,
       "peerDependencies": {
         "@types/react": "^19.0.0"
       }
@@ -2992,6 +2996,7 @@
       "integrity": "sha512-klQbnPAAiGYFyI02+znpBRLyjL4/BrBd0nyWkdC0s/6xFLkXYQ8OoRrSkqacS1ddVxf/LDyODIKbQ5TgKAf/Fg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.56.1",
         "@typescript-eslint/types": "8.56.1",
@@ -3579,6 +3584,7 @@
       "integrity": "sha512-UVJyE9MttOsBQIDKw1skb9nAwQuR5wuGD3+82K6JgJlm/Y+KI92oNsMNGZCYdDsVtRHSak0pcV5Dno5+4jh9sw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -4081,6 +4087,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001737",
         "electron-to-chromium": "^1.5.211",
@@ -4938,6 +4945,7 @@
       "integrity": "sha512-VmQ+sifHUbI/IcSopBCF/HO3YiHQx/AVd3UVyYL6weuwW+HvON9VYn5l6Zl1WZzPWXPNZrSQpxwkkZ/VuvJZzg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@eslint-community/eslint-utils": "^4.8.0",
         "@eslint-community/regexpp": "^4.12.1",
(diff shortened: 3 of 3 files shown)


mpawlow:
@coderabbitai full review

frontend/components/task-notification-menu.tsx
```diff
@@ -380,6 +380,16 @@ export function TaskNotificationMenu() {
                             </Button>
                           </div>
                         )}
+                        {hasFailedFileEntries(task) && (
```
mpawlow, line 383: ### (a) [Minor] UX: Active tasks force all error logs open simultaneously

- Note: This is Minor. Please feel free to ignore or optionally implement.

#### Problem

- In the **Recent Tasks** and **Past Tasks** sections, `TaskErrorContent` is rendered with `defaultExpanded={shouldExpandDetails}`, where `shouldExpandDetails` is computed to be `true` for at most **one** task at a time — the explicitly selected task or the most recent failure.
- The new Active Tasks block unconditionally passes `defaultExpanded={true}`, meaning **every** active task that has even one failed file entry starts with its full error log expanded.
- In concurrent-upload scenarios (e.g. multiple connectors syncing simultaneously, each with partial file failures), every card in the Active Tasks list will render with an open accordion. The notification panel becomes dominated by expanded failure logs, making it difficult to see the task status overview or reach the Cancel button without scrolling.
- The expansion state is local to each `TaskErrorContent` instance, so when the parent component remounts (e.g. polling updates cause a list reconciliation), all logs re-expand, potentially overriding a deliberate user collapse.

#### Potential Solution

Compute a `shouldExpandDetails` equivalent for active tasks and pass it:

```tsx
const mostRecentActiveFailureTaskId =
  activeTasks.find((t) => hasFailedFileEntries(t))?.task_id ?? null;

// Inside activeTasks.map(...):
const shouldExpandDetails =
  selectedTaskId === task.task_id ||
  (!selectedTaskId && task.task_id === mostRecentActiveFailureTaskId);

{hasFailedFileEntries(task) && (
  <div className="mt-3">
    <TaskErrorContent
      task={task}
      mode="recent"
      showHeader={false}
      defaultExpanded={shouldExpandDetails}
      expandTrigger={shouldExpandDetails ? selectedTaskTrigger : 0}
    />
  </div>
)}
```

This aligns active task behavior with recent/past task sections and ensures only one log is open at a time by default.

#### Alternative Solutions

- **Alt 1:** Always start collapsed (`defaultExpanded={false}`) for active tasks; the progress grid already shows the `{failed}` count in red, giving a clear signal without the full log taking space.
- **Alt 2:** Keep `defaultExpanded={true}` but cap it to the first active task with failures, equivalent to a null-coalescing guard: `defaultExpanded={task.task_id === activeTasks.find(hasFailedFileEntries)?.task_id}`.

frontend/components/task-notification-menu.tsx
```diff
@@ -380,6 +380,16 @@ export function TaskNotificationMenu() {
...
+                          <div className="mt-3">
+                            <TaskErrorContent
```
mpawlow, line 385: ### (b) [Normal] `expandTrigger` prop omitted for active task error content

#### Problem

- Every other invocation of `TaskErrorContent` with `showHeader={false}` in the same file passes an `expandTrigger` prop tied to `selectedTaskTrigger`:
  ```tsx
  expandTrigger={shouldExpandDetails ? selectedTaskTrigger : 0}
  ```
- The new Active Tasks block omits `expandTrigger` entirely, so `TaskErrorContent` uses its default of `0` (constant) for this prop.
- Consequence: if a user navigates to an active task via `selectedTaskId` (e.g. clicking a notification banner), the re-expansion mechanism that `expandTrigger` drives will not fire for active tasks. The `useEffect` in `TaskErrorContent` depends on `expandTrigger` changing to force re-expand, so programmatic re-triggers are silently swallowed.
- This creates a subtle but observable inconsistency: the same "select this task" action expands the error log in the Recent/Past sections but has no visible effect in the Active section.

#### Code References

- `frontend/components/task-notification-menu.tsx`, lines 470–478 (recent equivalent with `expandTrigger`)
- `frontend/components/task-notification-menu.tsx`, lines 555–563 (past equivalent with `expandTrigger`)
- `frontend/components/task-error-content.tsx`, lines 33–37 (`useEffect` dependency on `expandTrigger`)

#### Potential Solution

Add the missing prop, consistent with the recent/past task pattern:

```tsx
<TaskErrorContent
  task={task}
  mode="recent"
  showHeader={false}
  defaultExpanded={shouldExpandDetails}
  expandTrigger={shouldExpandDetails ? selectedTaskTrigger : 0}
/>
```

mpawlow, line 385: ### (c) [Minor] Error log shown in Active Tasks creates visual redundancy and potential UX confusion

#### Problem

- Active tasks (status `pending | running | processing`) already surface per-file failure counts through the progress breakdown grid:
  ```tsx
  <span className="text-red-600">
    {progress.detailed.failed} failed
  </span>
  ```
  This gives the user an immediate count of failures while the task is still running.
- The new `TaskErrorContent` block renders a second representation of the same information (per-file error messages) immediately below, while the spinner icon in the card header continues to show the task as still **running**.
- Presenting detailed failure log content alongside "running" status indicators can mislead users into thinking the entire task has failed, when it may still be actively processing remaining files. The `TaskErrorContent` component uses red destructive styling with an `XCircle` icon in its header-enabled form — even without the header, the accordion trigger reads `{failedCount} failed` in `text-destructive`, creating a failed-task visual language inside a card that simultaneously shows a blue animated spinner.
- There is no parallel in the codebase: terminal tasks (failed/error status) receive `TaskErrorContent` without a progress section; completed-with-partial-failures tasks receive it after status-based routing. Active tasks are unique in receiving both.

#### Code References

- `frontend/components/task-notification-menu.tsx`, lines 332–392 (full active task card body with both blocks)
- `frontend/components/task-notification-menu.tsx`, lines 337–363 (existing progress grid already shows failed count)
- `frontend/components/task-error-content.tsx`, lines 96–103 (accordion trigger uses `text-destructive` styling)

#### Potential Solution

- Consider showing the `TaskErrorContent` block only once the task has reached a terminal state, which is already handled by the Recent/Past sections. For in-progress tasks, the existing progress grid (which already shows failure counts) is sufficient.
- If real-time per-file error detail during active tasks is a product requirement, add a visual cue (e.g. a note saying "task still in progress") to disambiguate running-with-failures from fully-failed, and consider collapsing by default (see Issue a).

#### Alternative Solutions

- **Alt 1 (preferred):** Remove the new `hasFailedFileEntries` block from the Active Tasks section entirely. When the task transitions to `completed` or `failed`, the Recent/Past sections pick it up and surface the error content there.
- **Alt 2:** Keep the block but wrap it with a label such as "Partial failures so far" to distinguish it from a terminal failure display, and default it to collapsed.

frontend/tsconfig.json
```diff
@@ -11,15 +15,17 @@
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
```
mpawlow, line 14: ### (d) [Minor] `tsconfig.json` `jsx` mode conflicts with Next.js recommended setting

- I've been noticing the value flip-flopping between other PRs
   - It might be related to installed Node.js runtime version
   - Probably need to get a consensus from the team regarding what the value should be

#### Problem

- The `jsx` compiler option was changed from `"preserve"` to `"react-jsx"`.
- Next.js [recommends `"jsx": "preserve"`](https://nextjs.org/docs/pages/api-reference/config/typescript) for all Next.js TypeScript projects. The reason is that Next.js (via SWC or Babel) is responsible for transforming JSX at build time, and TypeScript is configured with `"noEmit": true` — TypeScript performs type-checking only and never emits transformed code.
- With `"noEmit": true` the runtime impact of this change is zero: TypeScript will not produce output files regardless of the `jsx` setting. However, the TypeScript language server and IDE tooling will now model JSX using the automatic runtime (`react/jsx-runtime`) rather than the classic `React.createElement` path.
- If any other tooling in the pipeline (Babel plugins, custom webpack transforms) reads `tsconfig.json`'s `jsx` value, the new setting could produce unexpected interactions.
- The change is otherwise benign for a React 17+ / React 19 project and `"react-jsx"` is arguably more accurate for type-checking. But deviating from the documented Next.js convention without a comment documenting the rationale is a maintainability concern.

#### Potential Solution

- Revert to `"jsx": "preserve"` unless there is a specific IDE type-checking failure that this change is meant to fix. If the change is intentional, add a comment to `tsconfig.json` (or the PR description) explaining why the Next.js default was overridden.

mpawlow (changes requested):
@ricofurtado 

## Code Review 1

- See PR comments: (a) to (d)
- All **Minor** (optional consideration) except (b) which is worth addressing
- Will defer to @mfortman11 here for overall UX / UI feedback

frontend/components/task-notification-menu.tsx
ricofurtado: Implemented

frontend/tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
   "compilerOptions": {
     "target": "ES2017",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": [
```
mpawlow: ### (2b) [Normal] `tsconfig.json` JSON reformatting conflicts with Biome formatter and may fail CI

### Problem

- The PR reformats `tsconfig.json` arrays from single-line to multi-line
- Biome is configured as the project formatter (`biome.json`) and is set to process all file types (`"ignoreUnknown": false`). Biome formats JSON files by default, applying its own line-length heuristic (default print width: 80 chars).
- CodeRabbit's earlier automated review also explicitly flagged this: *"Fix Biome formatting to unblock the CI pipeline."*


### Potential Solution

Run Biome's formatter on the file before committing:
```bash
cd frontend && npx biome format --write tsconfig.json
```
Then commit the result. Biome will normalize the arrays back to single-line (or whatever its canonical output is), and subsequent `biome check .` runs will pass.

frontend/components/task-notification-menu.tsx
mpawlow: ### (2a) [Minor] `expandTrigger` fix causes unintended accordion re-expansion on task deselection

### Problem

- The fix wires `expandTrigger` to `selectedTaskTrigger` when `selectedTaskId === task.task_id`, and to `0` otherwise:
  ```tsx
  expandTrigger={
    selectedTaskId === task.task_id
      ? selectedTaskTrigger
      : 0
  }
  ```
- The `TaskErrorContent` `useEffect` re-expands the accordion whenever either `defaultExpanded` **or** `expandTrigger` changes and `defaultExpanded` is truthy:
  ```tsx
  useEffect(() => {
    if (defaultExpanded) {
      setIsExpanded(true);
    }
  }, [defaultExpanded, expandTrigger]);
  ```
- Because `defaultExpanded={true}` is hardcoded (always truthy), **any** change in `expandTrigger` — including the transition from a non-zero value back to `0` — triggers the effect and forcibly re-opens the accordion.
- Concrete failure scenario:
  - User selects active task A → `expandTrigger` becomes `selectedTaskTrigger` (e.g., `1`)
  - User manually collapses task A's error log (`isExpanded = false`)
  - User then selects active task B → `selectedTaskId` moves to B, `selectedTaskTrigger` increments to `2`
  - For task A: `expandTrigger` changes from `1` to `0` (a real dependency change)
  - The `useEffect` fires; `defaultExpanded` is `true`, so `setIsExpanded(true)` is called
  - Task A's error log re-opens, overriding the user's explicit collapse action
- The Recent/Past task sections avoid this by coupling `defaultExpanded` to `shouldExpandDetails`:
  ```tsx
  defaultExpanded={shouldExpandDetails}
  expandTrigger={shouldExpandDetails ? selectedTaskTrigger : 0}
  ```
  When `shouldExpandDetails` becomes `false`, the `useEffect` fires but `if (defaultExpanded)` is `false`, so the accordion is left in whatever state the user set.

### Code References

- `frontend/components/task-notification-menu.tsx`, lines 389–394 (new `expandTrigger` prop)
- `frontend/components/task-error-content.tsx`, lines 32–37 (`useState` + `useEffect` expansion logic)
- Compare with Recent Tasks pattern at lines 429–433 and Past Tasks pattern at lines 509–512

### Potential Solution

Align active tasks with the pattern used in Recent/Past sections by computing `shouldExpandDetails` before the map and coupling `defaultExpanded` to it:

```tsx
// Compute once before activeTasks.map(...)
const mostRecentActiveFailureTaskId =
  activeTasks.find((t) => hasFailedFileEntries(t))?.task_id ?? null;

// Inside activeTasks.map(...)
const shouldExpandDetails =
  selectedTaskId === task.task_id ||
  (!selectedTaskId && task.task_id === mostRecentActiveFailureTaskId);

{hasFailedFileEntries(task) && (
  <div className="mt-3">
    <TaskErrorContent
      task={task}
      mode="recent"
      showHeader={false}
      defaultExpanded={shouldExpandDetails}
      expandTrigger={shouldExpandDetails ? selectedTaskTrigger : 0}
    />
  </div>
)}
```

This eliminates the re-expansion regression because when `shouldExpandDetails` becomes `false`, `defaultExpanded` is also `false`, so the `useEffect` fires but does not forcibly re-open the accordion.

mpawlow (changes requested):
@ricofurtado 

## Code Review 2

- See PR comments (2a), (2b)

frontend/tsconfig.json
ricofurtado: Done

mpawlow (approved):
## Code Review 3

- ✅  Approved / LGTM  🚀

PR #1516, opened 2026-05-04
fix: removes duplicated dependency from pyproject.toml

philnash:
prometheus-fastapi-instrumentator>=7.1.0 was in the list twice


lucaseduoli:
Merged in PR https://github.com/langflow-ai/openrag/pull/1518

PR #1520, opened 2026-05-04
refactor: dynamically load large library imports

mfortman11:
This pull request optimizes how the `MarkdownRenderer` and its dependency `CodeComponent` are loaded in the application to improve performance and user experience. The most significant changes are the conversion of these components to use dynamic imports with server-side rendering (SSR) disabled, which helps reduce initial bundle size and avoids SSR issues with client-only code.

**Dynamic Import Optimization:**

* Converted `MarkdownRenderer` to use dynamic imports with SSR disabled in both `assistant-message.tsx` and `onboarding-step.tsx`, ensuring that the component is only loaded on the client side to prevent SSR-related issues. [[1]](diffhunk://#diff-8c3036b8736dc6bef2884d3a9aa93835a386dff9a18a23f401d9995022563738R3-R13) [[2]](diffhunk://#diff-920b384f6a57a28a832a77415326ce6b0a1a3199225864ac7927c00707ba6988R2-R14)
* Updated `markdown-renderer.tsx` so that `CodeComponent` is dynamically imported with SSR disabled, and added a loading placeholder for better user experience while the component loads.

### DIFF
--- a/frontend/app/chat/_components/assistant-message.tsx
+++ b/frontend/app/chat/_components/assistant-message.tsx
@@ -1,7 +1,16 @@
 import { GitBranch } from "lucide-react";
 import { motion } from "motion/react";
+import dynamic from "next/dynamic";
 import DogIcon from "@/components/icons/dog-icon";
-import { MarkdownRenderer } from "@/components/markdown-renderer";
+
+const MarkdownRenderer = dynamic(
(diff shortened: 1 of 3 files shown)


Wallgau (approved):
LGTM

PR #1558, opened 2026-05-07
refactor: Cleanup useChatStreaming logic

mfortman11:
- `frontend/lib/chat-stream-parsers.ts` (new, 230 lines) — 5 exported functions: `parseOpenAIChatChunk`, `parseRealtimeChunk`, `parseOpenRAGChunk`, `detectImplicitToolCall`, `detectRAGFromContent`                       
- `frontend/hooks/useChatStreaming.ts` (716 → 324 lines) — loop body replaced with 4 function calls; all accumulator logic removed                                                                               
                                                                                                                                                                                                                 
  The hook now reads like an orchestrator: fetch → stream loop → parse → update state. Each protocol is independently testable by calling its function with a raw JSON object and asserting on the mutated       
  accumulators.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Improvements**
  * More reliable streaming across multiple AI providers, reducing incomplete or stalled assistant responses.
  * More accurate detection and handling of tool/function calls and retrieval (RAG) behavior during chats.
  * Better accumulation and delivery of final response content and token usage for clearer conversation results.
  * Improved end-of-stream validation to avoid spurious “no response” messages.

* **New Features**
  * Broader streaming format support and heuristic detection to surface retrieval/tool results more reliably.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


lucaseduoli (approved):
LGTM! thank you for this refactor

frontend/lib/chat-stream-parsers.ts
```diff
@@ -0,0 +1,359 @@
...
+        });
+      } else if (fn.arguments) {
+        const idx =
+          typeof toolCall.index === "number"
+            ? (toolCall.index as number)
+            : undefined;
+        const target =
+          idx !== undefined && idx >= 0 && idx < calls.length
+            ? calls[idx]
+            : calls[calls.length - 1];
```
Wallgau: If the chunk has an explicit idx but that call is not in calls yet, falling back to calls[calls.length - 1] can attach arguments to the wrong call and produce wrong tool-call content.
so falling back to calls[calls.length - 1] only works if undefined, maybe use
if (idx === undefined) {
  target = calls[calls.length - 1];
} else if (idx >= 0 && idx < calls.length) {
  target = calls[idx];
} else {
  target = undefined;
}

Wallgau (changes requested):

Wallgau (approved):
Great work!

PR #1560, opened 2026-05-07
feat: Add orphan reconcile and bulk-delete helper

edwinjosechittilappilly:
Introduce a reconcile pass and supporting bulk-delete utility to remove OpenSearch chunks for files that no longer exist remotely.

- Add reconcile_orphans_for_connector_type (src/api/connectors.py): lists all active connections for a connector type, strictly gates on unauthenticated connections or listing errors (abort with 0 deletes), aggregates paginated remote file IDs, computes orphans (indexed IDs not present remotely) and invokes delete_chunks_by_document_ids. Integrated into connector_sync and sync_all_connectors (skips reconcile when sync is capped).
- Add delete_chunks_by_document_ids (src/api/documents.py): issues a single delete_by_query with terms(document_id, ...) and conflicts="proceed", returns deleted count, and short-circuits on empty input.
- Update connector metadata painless params (src/connectors/service.py): include filename param and assign ctx._source.filename = params.filename in the update script so renamed files update indexed chunks.
- Add unit tests: cover delete_chunks_by_document_ids, reconcile_orphans_for_connector_type (gating, pagination, multi-connection union, error handling), and connector metadata filename behavior.

Behavior notes: reconcile is conservative to avoid false-positive deletions; bulk delete is defensive and returns 0 on unexpected responses or failures.

## Summary by CodeRabbit

* **New Features**
  * Connector synchronization now automatically detects and removes documents from the search index when source files have been deleted, keeping your index clean and search results accurate.

* **Bug Fixes**
  * Improved filename metadata handling during connector updates to prevent unintended data loss from null or empty values.

### DIFF
--- a/src/api/connectors.py
+++ b/src/api/connectors.py
@@ -90,6 +90,95 @@ async def get_synced_file_ids_for_connector(
         return [], []
 
 
+async def reconcile_orphans_for_connector_type(
+    connector_type: str,
+    user_id: str,
+    connector_service,
+    session_manager,
+    jwt_token: Optional[str],
+    existing_file_ids: List[str],
+) -> int:
+    """Delete OpenSearch chunks whose document_id is no longer present in the
+    union of remote file IDs across all active connections of this connector_type
+    for this user.
+
+    Strict gating: returns 0 (no deletes) if any active connection cannot be
+    listed cleanly — an unauthenticated connector or a listing exception aborts
+    the pass to avoid false-positive deletions.
+    """
+    if not existing_file_ids:
+        return 0
+
+    connections = await connector_service.connection_manager.list_connections(
+        user_id=user_id, connector_type=connector_type
+    )
+    active = [c for c in connections if c.is_active]
+    if not active:
+        logger.info(
+            "Skipping orphan reconcile — no active connections",
+            connector_type=connector_type,
+        )
+        return 0
+
+    remote_ids: set = set()
+    for conn in active:
+        try:
+            connector = await connector_service.get_connector(conn.connection_id)
+            if not connector or not connector.is_authenticated:
+                logger.info(
+                    "Skipping orphan reconcile — connection unauthenticated",
+                    connector_type=connector_type,
+                    connection_id=conn.connection_id,
+                )
+                return 0
+            page_token = None
+            while True:
+                page = await connector.list_files(page_token=page_token)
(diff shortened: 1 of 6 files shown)


src/api/connectors.py
ricofurtado, line 39: A question here: we're only relying on the OpenSearch to enforce user-level filtering internally. Which is fine. Are there any case where we don't use JWT, or DLS config is not applied? 
I think for the future RBAC implementations, it would be safer to also include user_id in the filter.

src/api/documents.py
```diff
@@ -99,6 +99,21 @@ async def delete_documents_by_filename_core(
         )
 
 
+async def delete_chunks_by_document_ids(
```
ricofurtado, line 102: Unlikely but still possible: If delete_chunks_by_document_ids deletes only by document_id, a document ID collision across connector types, users, or connections could delete too much.
We should consider namespace it like: document_id:connector_type:user/owner:connection_id
in the future

src/api/connectors.py
```diff
@@ -287,6 +376,18 @@ async def connector_sync(
                     connector_type=connector_type,
                     file_count=len(existing_file_ids),
                 )
+                # Reconcile orphans (files deleted at the source) before re-syncing.
```
ricofurtado, line 379: After reconcile runs, the code still calls sync_specific_files(..., existing_file_ids, ...) with the original list, including IDs that were just detected as orphaned.
Maybe reconcile should return the orphan IDs, or compute remaining_file_ids = existing_file_ids - orphan_ids, then sync only remaining IDs. The same issue exists in sync_all_connectors

src/api/connectors.py
ricofurtado, line 217: Old stuff, not in your PR, but it got me confused because the docstring is misleading:
The function name says sync from all active connections, but the code selects only the first working connection and then syncs all existing IDs through that one connection. also the return is 
```
"connections_synced": len(active_connections)
"message": f"Started syncing files from {len(active_connections)} 
```

ps. Same issue in `sync_all_connectors`. for each connector type

ricofurtado (changes requested):

src/api/connectors.py
edwinjosechittilappilly: Nope we rely on OpenSearch on DLS (Document level security) completely.

PR #1576, opened 2026-05-11
fix: get all file info at the same time displayed in the table

Wallgau:
Summary
When a document’s status switches to active on the Knowledge page, every column (size, type, embedding, owner, and the rest) now shows the right values at the same time, instead of some staying empty or wrong for a few seconds.

Problem
buildKnowledgeTableRows merged each search row with the in-flight task overlay using { ...file, ...taskFile, status }. Overlay rows mirror task API data: no real mimetype / file_size, and embedding_* can be unset. Object spread overwrote good search fields with those placeholders. status was corrected separately, so the row could flip to active while other columns were cleared until the overlay disappeared or search refetched.

Solution
frontend/lib/knowledge-table-state.ts: When a search file matches a task overlay, merge only the fields we intend from the task side (filename, source_url, connector_type, status, error, embedding_model / embedding_dimensions with ?? fallbacks). Leave the rest of the row from search so size, mimetype, owners, chunk stats stay intact.

frontend/app/knowledge/page.tsx: Remove refetchInterval: 5000 from useGetSearchQuery. Search still updates from normal query lifecycle and task-driven invalidation after ingestion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stopped periodic refetches for knowledge search to avoid unexpected background refreshes.
  * Unified toast/error behavior so notifications reflect the primary search/list state.
  * Ensure deletions refresh both search results and file listings so the UI stays in sync.

* **Improvements**
  * Sanitized and standardized filter handling so wildcard queries behave consistently and warnings show only for non-wildcard searches.
  * More accurate merging of backend and task-file data so filenames, status, source URLs, connector types, and embeddings prefer task-file values with safe fallbacks.
  * Immediate refreshes when task overlays change for faster UI updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1576)


frontend/contexts/task-context.tsx
```diff
@@ -378,19 +382,34 @@ export function TaskProvider({ children }: { children: React.ReactNode }) {
...
-            setFiles((prevFiles) =>
-              prevFiles.filter(
-                (file) =>
-                  file.task_id !== currentTask.task_id ||
-                  (completedHasFailures && file.status === "failed"),
-              ),
-            );
-            refetchSearch();
-          }, 500);
+          void (async () => {
```
mfortman11: let's make this a named function instead of a IIFE

frontend/lib/knowledge-table-state.ts
```diff
@@ -78,14 +88,15 @@ export function buildKnowledgeTableRows(
...
-    return (
-      taskFile.status !== "active" &&
-      !backendFiles.some(
-        (backendFile) =>
-          getKnowledgeFileIdentity(backendFile) ===
-          getKnowledgeFileIdentity(taskFile),
-      )
+    const identity = getKnowledgeFileIdentity(taskFile);
+    const inBackend = backendFiles.some(
+      (backendFile) => getKnowledgeFileIdentity(backendFile) === identity,
```
mfortman11: _nit_: `.some()` inside of a `.filter()` is pretty inefficient (O(n^2)) we could define the backend identities as a set outside of this function

mfortman11 (changes requested):

mfortman11 (approved):
Much improved LGTM

PR #1622, opened 2026-05-18
fix: Ensure SUCCESS status requires fetchable result from Docling

ricofurtado:
## Summary by CodeRabbit

* **New Features**
  * Two-step connector sync with a preview/confirmation dialog to show files/orphans before syncing.
  * Improved file upload flow that returns either immediate results or background-task IDs, updating chat upload UX.
  * Richer cloud-picker metadata and expanded MIME type labels for clearer file info.

* **Bug Fixes**
  * Polling now verifies conversion results before reporting success to avoid false success reports.
  * Connector sync now reconciles and safely deletes orphaned remote documents to prevent unintended removals.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1622?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/services/docling_polling_service.py
+++ b/src/services/docling_polling_service.py
@@ -54,6 +54,11 @@ class DoclingPollingService:
     ) -> DoclingPollResult:
         """Loop on Docling status until terminal or until max_seconds elapses.
 
+        A SUCCESS status is treated as ready only after the result endpoint
+        returns a payload with usable ``document.json_content``. This prevents
+        handing Langflow a task that Docling accepted but failed to convert
+        into a consumable document.
+
         Transient errors (network, 5xx, NOT_FOUND seen briefly before the task
         is registered server-side) are absorbed up to ``transient_retry_budget``
         before being surfaced as failures. The interval grows by
@@ -81,8 +86,25 @@ class DoclingPollingService:
             elapsed = time.monotonic() - start
 
             if snapshot.state == DoclingTaskState.SUCCESS:
+                try:
+                    await self.docling_service.fetch_task_result(task_id)
+                except Exception as e:
+                    detail = f"Docling result unavailable after SUCCESS status: {str(e)}"
+                    logger.warning(
+                        "Docling task reached SUCCESS but result fetch failed",
+                        task_id=task_id,
+                        detail=detail,
+                        elapsed_seconds=round(elapsed, 2),
+                    )
+                    return DoclingPollResult(
+                        outcome=PollOutcome.FAILED,
+                        detail=detail,
+                        last_snapshot=snapshot,
+                        elapsed_seconds=elapsed,
+                    )
+
                 logger.debug(
-                    "Docling task reached SUCCESS",
+                    "Docling task reached SUCCESS and result is available",
                     task_id=task_id,
                     elapsed_seconds=round(elapsed, 2),
                 )
--- a/tests/unit/test_docling_polling_service.py
+++ b/tests/unit/test_docling_polling_service.py
@@ -42,6 +42,7 @@ def no_sleep():
 @pytest.mark.asyncio
 async def test_returns_success_immediately_when_already_done(polling_service, mock_docling_service):
     mock_docling_service.check_task_status.return_value = _snap(DoclingTaskState.SUCCESS)
+    mock_docling_service.fetch_task_result.return_value = {"body": "ok"}
 
     result = await polling_service.poll_until_ready(
         task_id="t1", poll_interval=1.0, max_seconds=10.0
@@ -49,6 +50,7 @@ async def test_returns_success_immediately_when_already_done(polling_service, mo
 
     assert result.outcome == PollOutcome.SUCCESS
     assert mock_docling_service.check_task_status.call_count == 1
+    mock_docling_service.fetch_task_result.assert_awaited_once_with("t1")
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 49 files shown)


src/services/docling_polling_service.py
```diff
@@ -81,8 +86,25 @@ async def poll_until_ready(
             elapsed = time.monotonic() - start
 
             if snapshot.state == DoclingTaskState.SUCCESS:
+                try:
+                    await self.docling_service.fetch_task_result(task_id)
```
mpawlow, line 90: ### (a) [Normal] Fetched result is discarded — causes double HTTP call to Docling

#### Problem

- `fetch_task_result` is called inside `poll_until_ready` solely to prove the result is accessible, but its return value is never stored or propagated
- `DoclingPollResult` has no field to carry the document content, so the caller cannot reuse the fetched payload
- After `poll_until_ready` returns `SUCCESS`, `langflow_file_service.py` passes `task_id` to Langflow via `X-Langflow-Global-Var-DOCLING_TASK_ID` (line 216–217), and Langflow's DoclingRemote component will call the Docling result endpoint a second time — making the preflight fetch redundant
- The double-fetch adds latency on every successful document conversion and increases load on the Docling service

#### Code References

- `src/services/docling_polling_service.py:111–115` — `DoclingPollResult` returned with no result field
- `src/services/langflow_file_service.py:216–217` — task_id forwarded to Langflow which re-fetches
- `src/services/langflow_file_service.py:681` — `docling_task_id=task_id` passed into Langflow ingestion flow

#### Potential Solution

- Add a `result` field to `DoclingPollResult` and store the fetched content on success:

```python
@dataclass
class DoclingPollResult:
    outcome: PollOutcome
    detail: Optional[str] = None
    last_snapshot: Optional[DoclingStatusSnapshot] = None
    elapsed_seconds: float = 0.0
    result: Optional[dict] = None  # populated on SUCCESS
```

```python
doc = await self.docling_service.fetch_task_result(task_id)
return DoclingPollResult(
    outcome=PollOutcome.SUCCESS,
    last_snapshot=snapshot,
    elapsed_seconds=elapsed,
    result=doc,
)
```

- Then pass `poll_result.result` directly to Langflow instead of re-fetching via `task_id`.

src/services/docling_polling_service.py
```diff
@@ -81,8 +86,25 @@ async def poll_until_ready(
...
+                except Exception as e:
```
mpawlow, line 91: ### (b) [Normal] Transient network errors during result fetch immediately fail the task

#### Problem

- The `except Exception` block at line 91 catches every exception from `fetch_task_result`, including transient network errors (`httpx.RequestError`, timeouts) that do not indicate a real conversion failure
- `check_task_status` applies a configurable `transient_retry_budget` for 5xx/network errors; `fetch_task_result` has no equivalent — a single dropped packet after a valid `SUCCESS` status permanently fails the file ingestion
- This creates an asymmetry: the polling loop survives transient errors during status checks but is fatally intolerant of them on the final result fetch

#### Code References

- `src/services/docling_polling_service.py:131–148` — existing `NOT_FOUND` transient retry logic for comparison
- `src/services/docling_service.py:283` — `DoclingServeError` raised for network errors in `fetch_task_result`

#### Potential Solution

- Distinguish content failures (missing `json_content`) from transient failures, and retry on the latter:

```python
from services.docling_service import DoclingServeError

try:
    doc = await self.docling_service.fetch_task_result(task_id)
except DoclingServeError as e:
    if "missing document.json_content" in str(e):
        # Permanent content failure — return FAILED immediately
        ...
    else:
        # Transient — re-enter polling loop or retry with remaining budget
        ...
except Exception as e:
    # Unexpected error — still fail, but log at error level
    ...
```

#### Alternative Solutions

- Add a narrow retry loop (e.g., 3 attempts with a short delay) specifically wrapping the `fetch_task_result` call before returning `FAILED`, without changing the exception type discrimination

tests/unit/test_docling_polling_service.py
mpawlow, line 109: ### (c) [Normal] Two existing tests are missing `fetch_task_result` mock setup

#### Confidence

85

#### Problem

- `test_tolerates_brief_not_found_then_succeeds` (line 109) and `test_backoff_grows_interval_up_to_cap` (line 183) both end in a `DoclingTaskState.SUCCESS` path but set no `return_value` or `side_effect` on `mock_docling_service.fetch_task_result`
- The tests pass only because `AsyncMock()` returns a non-raising coroutine by default — if `fetch_task_result` were ever changed to raise by default (e.g. via `spec=`), both tests would silently switch from asserting `SUCCESS` to getting `FAILED`
- Neither test asserts that `fetch_task_result` was called, so the new code path introduced in this PR goes unverified in those scenarios

#### Code References

- `tests/unit/test_docling_polling_service.py:109–126` — `test_tolerates_brief_not_found_then_succeeds`, no `fetch_task_result` mock
- `tests/unit/test_docling_polling_service.py:183–206` — `test_backoff_grows_interval_up_to_cap`, no `fetch_task_result` mock or assertion

#### Potential Solution

Add the mock and assertion to both tests, matching the pattern already used in the updated tests:

```python
mock_docling_service.fetch_task_result.return_value = {"body": "ok"}
...
mock_docling_service.fetch_task_result.assert_awaited_once_with("t1")
```

mpawlow (changes requested):
@ricofurtado 
## Code Review 1

- See PR comments: (a) to (c)

src/services/docling_polling_service.py
ricofurtado: Good one! implemented. I created a DoclingTransientError ex class and implemented a retry.

src/services/docling_polling_service.py
ricofurtado: This would require some refactoring in Langflow flow+component. But I agree that's expensive double http call, especially with big files

tests/unit/test_docling_polling_service.py
ricofurtado: Checked

ricofurtado:
Closed. Added this solution to a broader PR

PR #1655, opened 2026-05-21
fix: static scan issues in helm chart

rodageve:
Description:
- Fixes static scan issues identified in helm chart #70873

## Summary by CodeRabbit

* **Chores**
  * Added explicit ephemeral-storage requests/limits for backend init containers to control startup resource usage.
  * Disabled automatic service account token mounting to improve security posture.
  * Updated checksum-based annotations so pods redeploy reliably when dotenv/config changes, ensuring consistent restarts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1655?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/kubernetes/helm/openrag/templates/backend/deployment.yaml
+++ b/kubernetes/helm/openrag/templates/backend/deployment.yaml
@@ -39,10 +39,7 @@ spec:
         - name: load-default-documents
           image: curlimages/curl:8.5.0
           resources:
-            limits:
-              ephemeral-storage: "500Mi"
-            requests:
-              ephemeral-storage: "500Mi"
+            {{- toYaml .Values.backend.initContainers.loadDefaultDocuments.resources | nindent 12 }}
           command: ["/bin/sh", "-c"]
           args:
             - |
@@ -115,10 +112,7 @@ spec:
         - name: load-default-flows
           image: curlimages/curl:8.5.0
           resources:
-            limits:
-              ephemeral-storage: "500Mi"
-            requests:
-              ephemeral-storage: "500Mi"
+            {{- toYaml .Values.backend.initContainers.loadDefaultFlows.resources | nindent 12 }}
           command: ["/bin/sh", "-c"]
           args:
             - |
--- a/kubernetes/helm/openrag/templates/backend/deployment.yaml
+++ b/kubernetes/helm/openrag/templates/backend/deployment.yaml
@@ -184,6 +184,8 @@ spec:
         {{- if .Values.langflow.enabled }}
         - name: wait-for-langflow
           image: busybox:1.36
+          resources:
+            {{- toYaml .Values.backend.initContainers.waitForLangflow.resources | nindent 12 }}
           command:
             - sh
             - -c
--- a/kubernetes/helm/openrag/templates/backend/deployment.yaml
+++ b/kubernetes/helm/openrag/templates/backend/deployment.yaml
@@ -23,7 +23,7 @@ spec:
         checksum/config-flows: {{ include (print $.Template.BasePath "/configmaps/flow-ids-configmap.yaml") . | sha256sum }}
         checksum/config-app: {{ include (print $.Template.BasePath "/configmaps/app-config-configmap.yaml") . | sha256sum }}
         checksum/secret-backend: {{ include (print $.Template.BasePath "/secrets/backend-secret.yaml") . | sha256sum }}
-        checksum/secret-dotenv: {{ include (print $.Template.BasePath "/templates/backend/backend-dotenv.yaml") . | sha256sum }}
+        checksum/secret-dotenv: {{ include (print $.Template.BasePath "/backend/backend-dotenv.yaml") . | sha256sum }}
     spec:
       serviceAccountName: {{ include "openrag.serviceAccountName" . }}
       {{- with .Values.podSecurityContext }}
@@ -38,6 +38,11 @@ spec:
         {{- if and .Values.backend.persistence.shared.enabled .Values.backend.persistence.documents.loadDefaults .Values.backend.persistence.documents.git.enabled }}
         - name: load-default-documents
           image: curlimages/curl:8.5.0
+          resources:
+            limits:
+              ephemeral-storage: "500Mi"
+            requests:
+              ephemeral-storage: "500Mi"
           command: ["/bin/sh", "-c"]
           args:
             - |
@@ -109,6 +114,11 @@ spec:
         {{- if and .Values.backend.persistence.shared.enabled .Values.langflow.flows.loadDefaults .Values.langflow.flows.git.enabled }}
         - name: load-default-flows
           image: curlimages/curl:8.5.0
+          resources:
+            limits:
+              ephemeral-storage: "500Mi"
+            requests:
+              ephemeral-storage: "500Mi"
           command: ["/bin/sh", "-c"]
           args:
             - |
--- a/kubernetes/helm/openrag/values.yaml
+++ b/kubernetes/helm/openrag/values.yaml
@@ -233,6 +233,23 @@ backend:
     limits:
       cpu: "4"
       memory: "16Gi"
+  # Init container resource limits
+  # Configure resource requests/limits for init containers that load default documents and flows
+  # Adjust ephemeral-storage based on the size of documents/flows being downloaded
+  initContainers:
+    loadDefaultDocuments:
+      resources:
+        requests:
+          ephemeral-storage: "500Mi"  # Ephemeral storage for downloading default documents
+        limits:
+          ephemeral-storage: "500Mi"
+    loadDefaultFlows:
+      resources:
+        requests:
+          ephemeral-storage: "500Mi"  # Ephemeral storage for downloading default flows
+        limits:
+          ephemeral-storage: "500Mi"
+
 
   # Logging and Service
   logFormat: "json"                       # json or text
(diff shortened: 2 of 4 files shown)


kubernetes/helm/operator/values.yaml
```diff
@@ -21,7 +21,7 @@ serviceAccount:
   # Specifies whether a service account should be created
   create: true
   # Automatically mount a ServiceAccount's API credentials?
-  automount: true
+  automount: false
```
mpawlow: ### (a) [Critical] Operator `automount: false` disables in-cluster API access, breaking the operator

#### Problem

- The operator `values.yaml` sets `serviceAccount.automount: false`, which propagates to `automountServiceAccountToken: false` in the `ServiceAccount` manifest via `kubernetes/helm/operator/templates/serviceaccount.yaml`
- The operator deployment (`kubernetes/helm/operator/templates/deployment.yaml`) does **not** override `automountServiceAccountToken` at the pod spec level, so the service account's `false` value takes effect
- Without the token mounted at `/var/run/secrets/kubernetes.io/serviceaccount/token`, `ctrl.GetConfigOrDie()` in `kubernetes/operator/cmd/main.go` (which calls `rest.InClusterConfig()`) fails immediately
- `rest.InClusterConfig()` reads the token from that path; if the file is absent, it returns an error
- `ctrl.GetConfigOrDie()` calls `os.Exit(1)` on error — the operator pod will crash-loop on every startup
- Leader election (enabled by default via `cfg.LeaderElectionEnabled`) and all reconciliation loops also require API access, reinforcing that this token is non-optional

#### Code References

- `kubernetes/helm/operator/templates/serviceaccount.yaml` — line 13 (`automountServiceAccountToken: {{ .Values.serviceAccount.automount }}`)
- `kubernetes/helm/operator/templates/deployment.yaml` — no `automountServiceAccountToken` field present at pod spec level
- `kubernetes/operator/cmd/main.go` — `ctrl.GetConfigOrDie()` call (in-cluster config depends on mounted token)

#### Potential Solution

Revert `automount` back to `true` in `values.yaml`. The static scanner likely flagged the *explicit* `true` value (which is redundant since it's the default), not the concept of token mounting itself. The correct remediation for that specific scan rule is to **remove the `automount` key entirely** from `values.yaml`, letting the Kubernetes default apply silently:

```yaml
serviceAccount:
  create: true
  # automount key removed — token mounting uses the Kubernetes default (true)
  annotations: {}
  name: ""
```

Then remove the corresponding `automountServiceAccountToken` field from the `ServiceAccount` template:

```yaml
{{- if .Values.serviceAccount.create -}}
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ include "openrag-operator.serviceAccountName" . }}
  ...
{{- end }}
```

This satisfies scanners that flag the explicit `automount: true` while preserving operator functionality.

#### Alternative Solutions

- **Alt 1 (Override at pod level):** Keep `automount: false` on the `ServiceAccount` but add `automountServiceAccountToken: true` to the pod spec in `kubernetes/helm/operator/templates/deployment.yaml`. This is more explicit and provides defense-in-depth documentation at the pod level, but is less intuitive.

  ```yaml
  spec:
    serviceAccountName: {{ include "openrag-operator.serviceAccountName" . }}
    automountServiceAccountToken: true  # required for in-cluster API access
  ```

- **Alt 2 (Projected volume):** Keep `automount: false` and mount the token explicitly via a projected `ServiceAccountToken` volume in the deployment. This gives fine-grained control over token audience and expiry, but adds significant template complexity for no material gain over Alt 1.

kubernetes/helm/openrag/templates/backend/deployment.yaml
mpawlow, line 185: ### (b) [Normal] `wait-for-langflow` initContainer missing ephemeral-storage resource limits

#### Problem

- The PR adds `ephemeral-storage` limits and requests to the `load-default-documents` and `load-default-flows` initContainers but leaves the `wait-for-langflow` initContainer (using `busybox:1.36`) with no `resources` block at all
- If the static analysis tool flagged all containers/initContainers without ephemeral-storage limits, `wait-for-langflow` will still trigger the same finding after this PR merges

#### Potential Solution

- Add a `resources` block to the `wait-for-langflow` initContainer. Because it only runs a `nc` (netcat) poll loop, an extremely small ephemeral-storage allocation is sufficient:

```yaml
- name: wait-for-langflow
  image: busybox:1.36
  resources:
    limits:
      ephemeral-storage: "10Mi"
    requests:
      ephemeral-storage: "10Mi"
  command:
    - sh
    - -c
    - |
      ...
```

#### Alternative Solutions

- **Alt 1:** Add the `resources` block with the same `500Mi` value used for the other two initContainers for consistency, even though it is over-provisioned for a simple polling container.

kubernetes/helm/openrag/templates/backend/deployment.yaml
```diff
@@ -38,6 +38,11 @@ spec:
         {{- if and .Values.backend.persistence.shared.enabled .Values.backend.persistence.documents.loadDefaults .Values.backend.persistence.documents.git.enabled }}
         - name: load-default-documents
           image: curlimages/curl:8.5.0
+          resources:
```
mpawlow, line 41: ### (c) [Normal] Ephemeral-storage limit hardcoded at `500Mi` — not configurable via `values.yaml`

#### Problem

- The `ephemeral-storage` limits and requests for `load-default-documents` and `load-default-flows` are hardcoded to `"500Mi"` directly in the deployment template
- Other resource limits (`backend.resources`, `langflow.resources`, etc.) are all driven from `values.yaml` entries, making them operator-configurable
- Deployments fetching large document sets or flows from GitHub may legitimately need more than `500Mi` of ephemeral scratch space; others may need less
- Hardcoding the value prevents tuning without modifying the chart template itself

#### Code References

- `kubernetes/helm/openrag/templates/backend/deployment.yaml` — lines 118–121 (`load-default-flows` limits/requests)

#### Potential Solution

Add configurable values to `values.yaml` and reference them in the template:

```yaml
# In values.yaml — under backend.persistence.documents and langflow.flows respectively
backend:
  persistence:
    documents:
      initContainer:
        resources:
          limits:
            ephemeralStorage: "500Mi"
          requests:
            ephemeralStorage: "500Mi"
langflow:
  flows:
    initContainer:
      resources:
        limits:
          ephemeralStorage: "500Mi"
        requests:
          ephemeralStorage: "500Mi"
```

Then in the template:

```yaml
resources:
  limits:
    ephemeral-storage: {{ .Values.backend.persistence.documents.initContainer.resources.limits.ephemeralStorage | quote }}
  requests:
    ephemeral-storage: {{ .Values.backend.persistence.documents.initContainer.resources.requests.ephemeralStorage | quote }}
```

#### Alternative Solutions

- **Alt 1 (Simpler):** Add a single shared `initContainers.ephemeralStorage` value that applies to all curl-based initContainers, reducing duplication at the cost of less granular control.

mpawlow (dismissed):
@rodageve 

## Code Review 1

- See PR comments: (a) to (c)

kubernetes/helm/operator/values.yaml
rodageve: Thanks, my bad, I checked in the change to operator helm values.yaml vs the regular OpenRag helm chart values.yaml.

zzzming (approved):

kubernetes/helm/openrag/values.yaml
```diff
@@ -559,6 +582,7 @@ serviceAccount:
   create: true
   name: ""
   annotations: {}
+  automount: false
```
mpawlow, line 585: ### (2a) [Normal] `automount: false` added to `values.yaml` but never rendered — security fix is a no-op

#### Problem

- `serviceAccount.automount: false` was added to `kubernetes/helm/openrag/values.yaml` (line 585)
- The `serviceaccount.yaml` template for the openrag chart does **not** reference `{{ .Values.serviceAccount.automount }}` anywhere — it never emits `automountServiceAccountToken`
- Kubernetes defaults to `true` when `automountServiceAccountToken` is absent from the `ServiceAccount` manifest
- The intended security hardening (disabling SA token auto-mounting) is therefore **never applied** to any pod; all containers running under this SA still receive a mounted token

#### Code References

- `kubernetes/helm/openrag/templates/serviceaccount.yaml` — lines 1–14: never emits `automountServiceAccountToken`

#### Potential Solution

Add `automountServiceAccountToken` rendering to the openrag `serviceaccount.yaml` template, mirroring the operator chart pattern:

```yaml
{{- if .Values.serviceAccount.create }}
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ include "openrag.serviceAccountName" . }}
  namespace: {{ include "openrag.namespace" . }}
  labels:
    {{- include "openrag.labels" . | nindent 4 }}
  {{- with .Values.serviceAccount.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
automountServiceAccountToken: {{ .Values.serviceAccount.automount }}
{{- end }}
```

#### Alternative Solutions

- Set `automountServiceAccountToken: false` at the pod spec level in both `backend/deployment.yaml` and `langflow/deployment.yaml` directly (does not require using the values key, avoids the template change):
  ```yaml
  spec:
    serviceAccountName: {{ include "openrag.serviceAccountName" . }}
    automountServiceAccountToken: false
  ```
  Pod-level override takes precedence over the SA-level setting, so this is equally effective and provides defense-in-depth even if the SA-level field is missing.

mpawlow (commented):
## Code Review 2 

- ✅  Approved / LGTM  🚀 
- See PR comment (2a) for optional consideration

PR #1660, opened 2026-05-22
fix: Ensure success status requires fetchable result in DoclingPolling

ricofurtado:
…gService
This pull request improves the robustness of the Docling polling service by ensuring that a task is only considered successful if its result can actually be fetched and is usable. It also adds new tests to verify this behavior, preventing downstream consumers from receiving incomplete or unusable data.

**Enhancements to Docling polling logic:**

* Updated `poll_until_ready` in `src/services/docling_polling_service.py` to treat a SUCCESS status as ready only after successfully fetching the result payload, ensuring that Langflow only receives consumable documents. If fetching the result fails, the poll returns a FAILED outcome with an appropriate error message. [[1]](diffhunk://#diff-4f1804e65cbd6e028ade63f12e0ae73de3499fbd1af8419c95ae1e10b2f7d6aeR57-R61) [[2]](diffhunk://#diff-4f1804e65cbd6e028ade63f12e0ae73de3499fbd1af8419c95ae1e10b2f7d6aeR89-R107)

**Testing improvements:**

* Added and updated unit tests in `tests/unit/test_docling_polling_service.py`:
  - Tests now verify that the result is fetched after a SUCCESS status and that the fetch is awaited exactly once. [[1]](diffhunk://#diff-d0ab958b95442fdf50a7c555a2afcebcd29c522bd1a2e10e9f73541edd0736c4R45-R53) [[2]](diffhunk://#diff-d0ab958b95442fdf50a7c555a2afcebcd29c522bd1a2e10e9f73541edd0736c4R66-R91)
  - Added a new test to ensure that if result fetching fails after a SUCCESS status, the polling service returns a FAILED outcome and includes the error detail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Polling now treats tasks as successful only after the final result is fetchable; fetch failures mark the task as FAILED with error details.

* **New Features**
  * Upload/sync: duplicate-file detection with a confirmation dialog that lists duplicate names and an option to overwrite (replace duplicates).
  * Sync API supports an explicit "replace duplicates" option.

* **UI**
  * New incident reporter icon; improved task error/notification UI and theme tokens for failure/task-status colors.
  * OAuth login now uses a consent prompt.

* **Tests**
  * Added tests for polling, filename dedupe, and task UI behaviors.

* **Documentation**
  * MCP SDK docs rewritten to announce retirement and the new built-in /mcp endpoint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1660?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


sdks/mcp/pyproject.toml
```diff
@@ -1,41 +1,27 @@
...
 ]
-dependencies = [
-    "mcp>=1.0.0",
-    "httpx>=0.28.0",  
-    "openrag-sdk==0.3.0rc1",
-]
-
-[project.scripts]
-openrag-mcp = "openrag_mcp:main"
+dependencies = []
```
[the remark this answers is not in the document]
ricofurtado: Came from Rebase

frontend/components/icons/incident-reporter-icon.tsx
```diff
@@ -0,0 +1,24 @@
...
+export function IncidentReporterIcon({
+  className,
+  ...props
+}: React.SVGProps<SVGSVGElement>) {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 32 32"
+      fill="currentColor"
+      aria-hidden
```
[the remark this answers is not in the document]
ricofurtado: Came from Rebase

frontend/components/duplicate-handling-dialog.tsx
```diff
@@ -61,6 +77,19 @@ export const DuplicateHandlingDialog: React.FC<
           </DialogDescription>
         </DialogHeader>
 
+        {namesProvided && (
+          <ul className="text-sm text-muted-foreground list-disc pl-5 space-y-0.5">
+            {visibleNames.map((name) => (
+              <li key={name} className="break-all">
+                {name}
```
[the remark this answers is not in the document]
ricofurtado: Came with the Rebase

frontend/app/upload/[provider]/page.tsx
```diff
@@ -467,6 +475,89 @@ export default function UploadProviderPage() {
...
+          submitSync(connector, nonDuplicateFiles, false);
+        } else {
+          toast.info(
+            `All ${duplicateNames.length} selected file(s) already exist. Nothing was synced.`,
+          );
+        }
+      }
+      setPendingSync(null);
+    }
+    setDuplicateDialogOpen(open);
```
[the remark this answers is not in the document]
Comment (ricofurtado) [reply frontend/app/upload/[provider]/page.tsx:558]: Came with the rebase

src/services/docling_polling_service.py
```diff
@@ -81,8 +88,54 @@ async def poll_until_ready(
             elapsed = time.monotonic() - start
 
             if snapshot.state == DoclingTaskState.SUCCESS:
+                result_fetch_errors = 0
```
mpawlow, line 91: ## (1a) [Normal] Inner result-fetch retry loop ignores `max_seconds` deadline

### Problem

- The `while True` retry loop introduced for `fetch_task_result` does not check the outer polling deadline
- Once `DoclingTaskState.SUCCESS` is reached, the loop can run indefinitely until the `result_fetch_retry_budget` is exhausted
- With `result_fetch_retry_budget=3` and `max_interval=30.0s` (the outer backoff-grown `interval`), this adds up to 4 × 30s = **120 extra seconds** beyond `max_seconds` before returning `FAILED`

### Potential Solution

Check `deadline` at the top of each retry iteration:

```python
while True:
    if time.monotonic() >= deadline:
        return DoclingPollResult(
            outcome=PollOutcome.TIMEOUT,
            detail="Docling result fetch timed out",
            last_snapshot=snapshot,
            elapsed_seconds=time.monotonic() - start,
        )
    try:
        await self.docling_service.fetch_task_result(task_id)
        break
    except DoclingTransientError as e:
        ...
```

src/services/docling_polling_service.py
```diff
@@ -81,8 +88,54 @@ async def poll_until_ready(
...
+                                last_snapshot=snapshot,
+                                elapsed_seconds=elapsed,
+                            )
+                        logger.debug(
+                            "Transient error fetching Docling result, retrying",
+                            task_id=task_id,
+                            attempt=result_fetch_errors,
+                            error=str(e),
+                        )
+                        await asyncio.sleep(interval)
```
mpawlow: ## (1b) [Normal] Transient-error retry sleep reuses backoff-grown outer poll interval


### Problem

- In the result-fetch retry block, `await asyncio.sleep(interval)` reuses the outer loop's `interval` variable, which grows via exponential backoff up to `max_interval` (default 30s)
- If polling has already backed off to 30s intervals, each transient retry during result fetch will sleep 30s — an excessive and user-visible delay after the task has already succeeded in Docling

### Potential Solution

Use a short, fixed retry delay for result fetches (e.g., 2s), separate from the outer poll interval:

```python
RESULT_FETCH_RETRY_DELAY = 2.0
...
await asyncio.sleep(RESULT_FETCH_RETRY_DELAY)
```

src/services/docling_polling_service.py
```diff
@@ -81,8 +88,54 @@ async def poll_until_ready(
             elapsed = time.monotonic() - start
 
             if snapshot.state == DoclingTaskState.SUCCESS:
+                result_fetch_errors = 0
+                while True:
+                    try:
```
mpawlow, line 114: ## (1c) [Normal] `fetch_task_result` response is discarded — double API call and race window

### Problem

- `await self.docling_service.fetch_task_result(task_id)` is called solely for side-effect validation — the returned payload is not captured or returned via `DoclingPollResult`
- Callers (e.g., Langflow) must call `fetch_task_result` again to retrieve the document content, meaning **every successful conversion triggers two fetches**
- Between the validation call and the caller's second fetch, the Docling result could expire or be evicted (race window), resulting in a caller-side failure even though polling reported `PollOutcome.SUCCESS`

### Background Information

- `DoclingPollResult` currently carries `outcome`, `last_snapshot`, and `elapsed_seconds` only
- Adding the fetched payload to `DoclingPollResult` would eliminate the double call and close the race window while also making the intent ("result is available") explicit in the return value

### Code References

- `src/services/docling_polling_service.py` — the `DoclingPollResult` dataclass (lines 33–37)

### Potential Solution

Extend `DoclingPollResult` to carry the fetched result and return it:

```python
@dataclass
class DoclingPollResult:
    outcome: PollOutcome
    detail: str | None = None
    last_snapshot: DoclingStatusSnapshot | None = None
    elapsed_seconds: float = 0.0
    result: dict[str, Any] | None = None  # populated on SUCCESS
```

Then in the success path:

```python
result_payload = await self.docling_service.fetch_task_result(task_id)
...
return DoclingPollResult(
    outcome=PollOutcome.SUCCESS,
    last_snapshot=snapshot,
    elapsed_seconds=elapsed,
    result=result_payload,
)
```

sdks/mcp/pyproject.toml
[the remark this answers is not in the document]
mpawlow: ## (1d) [Major] Retired MCP package ships with empty `dependencies` — broken PyPI install

### Problem

- `sdks/mcp/pyproject.toml` now sets `dependencies = []` and keeps the `[build-system]` block intact
- `openrag-mcp==0.3.0` will be a **publishable but broken package** on PyPI — users who `pip install openrag-mcp==0.3.0` or upgrade from an earlier version will get a wheel that imports nothing and cannot do anything
- The `[project.scripts]` entry (`openrag-mcp = "openrag_mcp:main"`) was removed, so the CLI entrypoint is gone, but the package is still installable

### Background Information

- Note: While it is technically pre-existing in origin, it is landing on `main` via this PR and should be resolved before merge
- Recommended options: (1) keep a shim that raises `RuntimeError` with migration instructions, (2) yank the old release and ensure 0.3.0 is never published, or (3) remove the `[build-system]` block entirely to prevent building a wheel

### Potential Solution

Add a deprecation shim so the package fails loudly with a helpful message instead of silently:

```python
# sdks/mcp/src/openrag_mcp/__init__.py
raise ImportError(
    "openrag-mcp is retired. Connect directly to your OpenRAG /mcp endpoint. "
    "See https://docs.openr.ag/mcp for migration instructions."
)
```

### Alternative Solutions

- Remove `[build-system]` entirely so the package cannot be built/published as a wheel

src/services/docling_polling_service.py
```diff
@@ -81,8 +88,54 @@ async def poll_until_ready(
...
+                        await self.docling_service.fetch_task_result(task_id)
+                        break
+                    except DoclingTransientError as e:
+                        result_fetch_errors += 1
+                        if result_fetch_errors > result_fetch_retry_budget:
```
mpawlow, line 119: ## (1e) [Minor] Off-by-one in `result_fetch_retry_budget` comparison

### Problem

- The condition `if result_fetch_errors > result_fetch_retry_budget` (with default `budget=3`) allows **4 attempts** before failing, not 3
- The error message `"failed after {result_fetch_errors} transient errors"` will report `4`, which is inconsistent with the documented budget of `3`

### Potential Solution

Change `>` to `>=` to match the semantics of the named parameter:

```python
if result_fetch_errors >= result_fetch_retry_budget:
```

src/dependencies.py
mpawlow, line 629: ## (1f) [Minor] `import base64` placed inside function body

### Problem

- `import base64` was added inside `get_api_key_user_async` rather than at the module's top-level
- This is non-idiomatic Python (PEP 8) and inconsistent with the existing `from config.settings import IBM_AUTH_ENABLED` pattern in the same function, which is a lazy import for a specific reason (circular dependency avoidance)
- `base64` is a stdlib module with no circular-import risk; it belongs at the top of the file

### Potential Solution

Move to file-level imports at the top of `src/dependencies.py`.

frontend/components/duplicate-handling-dialog.tsx
mpawlow, line 83: ## (1h) [Minor] `key={name}` in duplicate names list risks React duplicate-key warning

### Problem

- `visibleNames.map((name) => (<li key={name} ...>))` uses the filename string as the React key
- If the same filename appears more than once in `duplicateNames` (which is unlikely but possible if the caller provides duplicates), React will emit a duplicate-key warning and may misrender the list

### Potential Solution

Use the array index (or a prefixed composite key) to guarantee uniqueness:

```tsx
{visibleNames.map((name, idx) => (
  <li key={`${idx}-${name}`} className="break-all">
    {name}
  </li>
))}
```

tests/unit/test_docling_polling_service.py
```diff
@@ -62,13 +70,32 @@ async def test_loops_through_processing_then_success(
...
     result = await polling_service.poll_until_ready(
         task_id="t1", poll_interval=1.0, max_seconds=60.0
     )
 
     assert result.outcome == PollOutcome.SUCCESS
     assert mock_docling_service.check_task_status.call_count == 4
     assert no_sleep.call_count == 3
+    mock_docling_service.fetch_task_result.assert_awaited_once_with("t1")
+
+
```
mpawlow, line 84: ## (1i) [Minor] No test for transient error budget exhaustion in result fetch

### Problem

- `tests/unit/test_docling_polling_service.py` adds coverage for the `DoclingServeError` (non-transient) path but has no test verifying that `DoclingTransientError` retries exceeding `result_fetch_retry_budget` also return `PollOutcome.FAILED`
- The off-by-one in (1e) above would only be caught by a parametrized test that asserts exactly how many retries occur

### Potential Solution

Add a test similar to:

```python
@pytest.mark.asyncio
async def test_result_fetch_exceeds_transient_budget(polling_service, mock_docling_service, no_sleep):
    mock_docling_service.check_task_status.return_value = _snap(DoclingTaskState.SUCCESS)
    mock_docling_service.fetch_task_result.side_effect = DoclingTransientError("network error")

    result = await polling_service.poll_until_ready(
        task_id="t1", poll_interval=1.0, max_seconds=60.0, result_fetch_retry_budget=3
    )

    assert result.outcome == PollOutcome.FAILED
    assert mock_docling_service.fetch_task_result.await_count == 3  # exactly budget, not budget+1
```

mpawlow (changes requested):
@ricofurtado 

## Code Review 1

- See PR comments: (1a) to (1h)

mpawlow (changes requested):
@ricofurtado 

## Code Review 2

- See PR comments: (2d) and (2g)

---

### (2d) [Normal] Connector sync silently treats all files as non-duplicate when duplicate-check API fails

#### Problem

- In `handleSync`, when `duplicateCheck` raises an exception for a file, the catch block logs to `console.error` and returns `{ file, isDuplicate: false }`
- If the `/api/documents/check-filename` endpoint is unavailable, returns a non-200 status, or times out, **every** selected file is treated as a non-duplicate and is submitted for sync unconditionally
- This can cause unintended duplicate ingestion: the user pressed "Sync" expecting duplicates to be surfaced, but the dialog never appears and all files are re-ingested silently

#### Background Information

- The behavior is "fail-open" by design (proceed when unsure), but there is no user-visible signal that the duplicate check was skipped
- Other upload flows (e.g., local file upload) surface errors to the user via `toast.error`; the connector path has no equivalent notification when the check service is unavailable

#### Code References

- `frontend/app/upload/[provider]/page.tsx:496-502` (catch block inside `handleSync`)

#### Potential Solution

Show a user-visible warning toast when duplicate-check requests fail, so the user is aware the check was skipped:

```typescript
} catch (err) {
  console.error(`[Connector Sync] Duplicate check failed for ${file.name}:`, err);
  toast.warning("Duplicate check unavailable", {
    description: "Could not verify duplicates. Files will be synced without deduplication.",
  });
  return { file, isDuplicate: false };
}
```

#### Alternative Solutions

- Fail the entire sync attempt on any duplicate-check error and prompt the user to retry, which prevents silent re-ingestion

---

### (2g) [Normal] `extractReadableLine` colon-split is fragile for URL-containing error strings

#### Problem

- `extractReadableLine` in `task-error-display.ts` splits on `":"` to find the last colon-delimited clause:
  ```typescript
  const colonParts = beforeCausedBy.split(":");
  const lastClause = colonParts[colonParts.length - 1]?.trim();
  ```
- When error strings contain URLs (e.g., `"Docling fetch failed: https://docling:8080/result not found"`), the split on `":"` fragments the URL and returns a path segment (e.g., `"8080/result not found"`) as the "readable" last clause
- The minimum-length guard (`>= 10`) allows these fragments to be selected, producing unhelpful single-line summaries in the task panel

#### Background Information

- The condition `lastClause.length >= 10 && lastClause.length <= FILE_ERROR_MAX_LINE_LENGTH` is intended to filter out very short or very long fragments, but URL port/path fragments frequently satisfy both constraints

#### Code References

- `frontend/lib/task-error-display.ts:55-67`

#### Potential Solution

Split only on `": "` (colon + space) instead of `":"`, which avoids fragmenting `https://host:port` patterns while still splitting on human-readable `"Component: error message"` pairs:

```typescript
const colonParts = beforeCausedBy.split(": ");
```

PR #1674, opened 2026-05-25
fix: add replace_duplicates option for Google Drive sync

ricofurtado:
# fix: Google Drive sync replaces existing files instead of failing

## Problem

When syncing Google Drive via the UI sync button (both the connector-specific `/connectors/google_drive/sync` and the global `/connectors/sync-all` endpoints), files that were already indexed would fail with:

> `File with name 'XXXX' already exists`

This happened because `ConnectorFileProcessor` and `LangflowConnectorFileProcessor` both gate re-ingestion behind a `replace_duplicates` flag that defaults to `False`. The sync code paths were never passing `replace_duplicates=True`, so every re-sync of a previously indexed file was treated as a duplicate error instead of an update.

The `sync_specific_files` path (used for manual file selection) already had `replace_duplicates` wired correctly — the gap was exclusively in the background/button-triggered sync paths.

## Changes

- **`src/api/connectors.py`** — Pass `replace_duplicates=True` for `google_drive` in both the `connector_sync` and `sync_all_connectors` endpoints, covering the `sync_specific_files` call (primary path when document IDs are indexed) and the `sync_connector_files` fallback (filename-based path for older Langflow-ingested files).
- **`src/connectors/service.py`** — Add `replace_duplicates: bool = False` parameter to `sync_connector_files` and thread it through to `ConnectorFileProcessor`.
- **`src/connectors/langflow_connector_service.py`** — Add `filename_filter` and `replace_duplicates` parameters to `sync_connector_files` (both were missing, causing a latent `TypeError` on the fallback path) and thread `replace_duplicates` through to `LangflowConnectorFileProcessor`.

## Scope

Only Google Drive sync behavior changes. All other connectors continue to receive `replace_duplicates=False` (the existing default). Manual file uploads via the UI are unaffected — those processors are instantiated separately and their `replace_duplicates` flag remains user-controlled.

## Test plan

- [ ] Sync a Google Drive connection that already has indexed files — tasks should complete with status `indexed` or `unchanged`, no `already exists` errors
- [ ] Confirm a modified Google Drive file is re-indexed with updated content after sync
- [ ] Confirm an unmodified file returns `unchanged` (hash deduplication still applies)
- [ ] Verify other connector types (OneDrive, SharePoint, S3, IBM COS) are unaffected
- [ ] Run `pytest tests/unit/ -k connector`

## Summary by CodeRabbit

* **New Features**
  * Google Drive, OneDrive, and SharePoint syncs now replace duplicate files by default for cleaner results.
  * Connector syncs expose an explicit "replace duplicates" option to control duplicate handling.
  * Connector syncs can be limited by filename filters so only selected files are queued and processed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1674?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/api/connectors.py
+++ b/src/api/connectors.py
@@ -21,7 +21,7 @@ logger = get_logger(__name__)
 
 def _connector_sync_should_replace(connector_type: str) -> bool:
     """Return True for connector types where sync should replace existing indexed files."""
-    return connector_type == "google_drive"
+    return connector_type in ["google_drive", "sharepoint", "onedrive"]
 
 
 async def get_synced_file_ids_for_connector(
--- a/src/api/connectors.py
+++ b/src/api/connectors.py
@@ -508,6 +508,7 @@ async def connector_sync(
                     user.user_id,
                     existing_file_ids,
                     jwt_token=jwt_token,
+                    replace_duplicates=(connector_type == "google_drive"),
                 )
             else:
                 # Fallback: use filename filtering (for Langflow-ingested files without document_id)
@@ -522,6 +523,7 @@ async def connector_sync(
                     max_files=None,
                     jwt_token=jwt_token,
                     filename_filter=set(existing_filenames),
+                    replace_duplicates=(connector_type == "google_drive"),
                 )
         task_ids = [task_id]
         await TelemetryClient.send_event(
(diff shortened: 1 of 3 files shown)


lucaseduoli (approved):
LGTM

PR #1623, opened 2026-05-18
feat: disable langflow ingestion via UI

lucaseduoli:
This pull request introduces a new feature allowing users to disable document ingestion via Langflow and instead use a traditional OpenRAG ingestion pipeline. The change is implemented as a new `disable_ingest_with_langflow` setting, which is configurable through both the frontend settings UI and backend configuration/environment variables. The backend logic, API endpoints, and processors have been updated to respect this setting, ensuring that document ingestion routes and processing behavior adapt accordingly.

**Frontend: Ingestion Settings UI and API**

- Added a new toggle in the ingestion settings section (`IngestSettingsSection`) to allow users to enable or disable Langflow ingestion. This setting is now part of the settings API requests and state management. [[1]](diffhunk://#diff-98d32c3d9b9d4405a524181d14e28a44dbd05b21a7db2d2fa656ec72fa08ddb1R49-R50) [[2]](diffhunk://#diff-98d32c3d9b9d4405a524181d14e28a44dbd05b21a7db2d2fa656ec72fa08ddb1R151-R166) [[3]](diffhunk://#diff-98d32c3d9b9d4405a524181d14e28a44dbd05b21a7db2d2fa656ec72fa08ddb1R209) [[4]](diffhunk://#diff-98d32c3d9b9d4405a524181d14e28a44dbd05b21a7db2d2fa656ec72fa08ddb1R253) [[5]](diffhunk://#diff-98d32c3d9b9d4405a524181d14e28a44dbd05b21a7db2d2fa656ec72fa08ddb1R445-R463) [[6]](diffhunk://#diff-87697f49f6ea81ac0de26e2f69087918fb4636e76893e68684211ecf187c9f9eR21) [[7]](diffhunk://#diff-057cd44a1e25ed69f8b92eaf2dc7737c781478f2cda1424b30292d65ebddaaa1R21)

**Backend: Configuration and Models**

[...]

PR #1709, opened 2026-05-29
feat: retry broke after #1623 started deleting staged temp files

Wallgau:
upload retry broke after #1623 started deleting staged temp files
immediately after processing. The enhanced task API still marks those
failures as RETRYABLE, but retry then fails with source_file_missing
because the temp path no longer exists on disk.
Clean up upload temps per file only when they are not retryable:
- keep temps for failed local uploads classified as RETRYABLE
- delete temps for successes and non-retryable failures
- on cancel, force-delete all temps (no retry expected)
- if unlink fails, keep the path in temp_file_paths so cleanup can be
  retried later (safe_unlink swallows errors, so verify with exists)
Add unit tests for retryable retention, mixed success/failure cleanup,
failed unlink tracking, and ensuring retryable temps are never unlinked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retryable local upload failures now retain staged temp files for retry; non-retryable temps are removed. Cancelling tasks, evicting old tasks, or shutting down now force-remove staged temps to prevent leftovers. Temp path normalization now warns on mismatches to avoid broken retry retention.

* **Tests**
  * Added unit tests covering retry vs non-retry temp retention, unlink behavior, and forced cleanup on eviction/shutdown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/langflow-ai/openrag/pull/1709?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/services/task_service.py
+++ b/src/services/task_service.py
@@ -594,20 +594,9 @@ class TaskService:
                 upload_task.status = TaskStatus.COMPLETED
                 upload_task.updated_at = time.time()
 
-            # Clean up temp files after all processing is complete
-            if hasattr(upload_task, "temp_file_paths") and upload_task.temp_file_paths:
-                from utils.file_utils import safe_unlink
-
-                for temp_path in upload_task.temp_file_paths:
-                    try:
-                        safe_unlink(temp_path)
-                        logger.debug("Cleaned up temp file", temp_path=temp_path)
-                    except Exception as cleanup_error:
-                        logger.warning(
-                            "Failed to clean up temp file after processing",
-                            temp_path=temp_path,
-                            error=str(cleanup_error),
-                        )
+            # Clean up upload temps that are not retryable; keep local RETRYABLE
+            # failures so retry can re-read the staged source file.
+            self._cleanup_upload_temp_files(upload_task)
 
             status: str = "FAILED"
 
@@ -1368,7 +1357,51 @@ class TaskService:
                     file_task.error = "Task cancelled by user"
                     file_task.updated_at = time.time()
 
+        self._cleanup_upload_temp_files(upload_task, force=True)
+
         return True
+
+    def _is_retryable_local_upload_temp(
+        self, upload_task: UploadTask, temp_path: str
+    ) -> bool:
+        """True when a staged temp belongs to a failed local RETRYABLE upload."""
+        file_task = upload_task.file_tasks.get(temp_path)
+        if file_task is None or file_task.status != TaskStatus.FAILED:
+            return False
+        if not os.path.isabs(temp_path):
+            return False
+        metadata = self._infer_failure_metadata(file_task)
+        return bool(metadata and metadata.get("actionable_by") == "RETRYABLE")
+
+    def _cleanup_upload_temp_files(
+        self, upload_task: UploadTask, *, force: bool = False
+    ) -> None:
+        """Remove staged upload temp files that are not retryable.
+
+        Keeps temps for failed local uploads classified as RETRYABLE so retry
+        can reuse the original source path. Use *force* to delete all temps
+        (e.g. user cancelled the task).
+        """
+        if not getattr(upload_task, "temp_file_paths", None):
+            return
+
+        from utils.file_utils import safe_unlink
+
+        retained: list[str] = []
+        for temp_path in upload_task.temp_file_paths:
+            if not force and self._is_retryable_local_upload_temp(upload_task, temp_path):
+                retained.append(temp_path)
+                continue
+            safe_unlink(temp_path)
+            if os.path.exists(temp_path):
+                retained.append(temp_path)
+                logger.warning(
+                    "Failed to clean up temp file after processing",
+                    temp_path=temp_path,
+                )
+            else:
+                logger.debug("Cleaned up temp file", temp_path=temp_path)
+        upload_task.temp_file_paths = retained
 
     async def shutdown(self):
         """Cleanup process pool and cancel all background tasks
--- a/src/services/task_service.py
+++ b/src/services/task_service.py
@@ -388,7 +388,17 @@ class TaskService:
         if temp_file_paths:
             if upload_task.temp_file_paths is None:
                 upload_task.temp_file_paths = []
-            upload_task.temp_file_paths.extend(temp_file_paths)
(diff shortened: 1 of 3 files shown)


src/services/task_service.py
```diff
@@ -1368,16 +1360,57 @@ async def cancel_task(self, user_id: str, task_id: str) -> bool:
                     file_task.error = "Task cancelled by user"
                     file_task.updated_at = time.time()
 
+        self._cleanup_upload_temp_files(upload_task, force=True)
+
         return True
 
+    def _is_retryable_local_upload_temp(self, upload_task: UploadTask, temp_path: str) -> bool:
```
mpawlow: ### (1a) [Normal] `_is_retryable_local_upload_temp` silently fails when `temp_path` is not a `file_tasks` key

#### Problem

- `_is_retryable_local_upload_temp` does `upload_task.file_tasks.get(temp_path)` to find the corresponding `FileTask`
- `file_tasks` is keyed by `str(item)` where `item` comes from the `items` list passed to `create_custom_task`
- `temp_file_paths` is a separate list that is expected to overlap with `items`, but this is not enforced
- When `temp_path` is not a `file_tasks` key, `file_task` is `None` and the method returns `False`
- This silently classifies the temp as non-retryable, causing it to be deleted even if the underlying failure was retryable

#### Background Information

- For `create_upload_task` (line 251–256): `temp_file_paths=file_paths` and `items=file_paths` — they are identical, so the lookup always succeeds ✓
- For `create_langflow_upload_task` (line 300–301): `temp_file_paths=temp_file_paths if temp_file_paths is not None else file_paths` — a caller could pass distinct temp paths that differ from the processing keys in `file_tasks`
- If a new upload type ever stages to a temp location that differs from the item key, the silent `None` return would defeat the entire RETRYABLE retention goal with no error, warning, or test failure

#### Code References

- `src/services/task_service.py:359–365` — `file_tasks` dict keying in `create_custom_task`
- `src/services/task_service.py:300–301` — `temp_file_paths` parameter in `create_langflow_upload_task`

#### Potential Solution

Add a defensive guard and/or explicit documentation of the invariant directly in `_is_retryable_local_upload_temp`:

```python
def _is_retryable_local_upload_temp(self, upload_task: UploadTask, temp_path: str) -> bool:
    """True when a staged temp belongs to a failed local RETRYABLE upload.

    Precondition: temp_path must equal the corresponding file_tasks key (i.e.
    the original item path). Callers that use distinct temp paths should not
    call this method or extend it to accept a FileTask directly.
    """
    if not os.path.isabs(temp_path):
        return False
    file_task = upload_task.file_tasks.get(temp_path)
    if file_task is None or file_task.status != TaskStatus.FAILED:
        return False
    metadata = self._infer_failure_metadata(file_task)
    return bool(metadata and metadata.get("actionable_by") == "RETRYABLE")
```

Or alternatively, add a `logger.debug` when `file_task is None` so silent misses are observable in logs.

#### Alternative Solutions

- Accept `file_task: FileTask` directly instead of `temp_path: str`, eliminating the dict lookup entirely. This would require callers to pass the file task explicitly, making the coupling explicit and compile-time visible.

tests/unit/test_task_service_retry_failed_files.py
```diff
@@ -245,3 +245,108 @@ async def test_retry_connector_id_path_does_not_require_local_file(task_service)
...
+    ft_success = FileTask(file_path=other_path, filename="ok.pdf")
+    ft_success.status = TaskStatus.COMPLETED
+    task = UploadTask(
+        task_id="task-1",
+        total_files=2,
+        file_tasks={retryable_path: ft_retryable, other_path: ft_success},
+        temp_file_paths=[retryable_path, other_path],
+    )
+
+    with patch("utils.file_utils.safe_unlink") as mock_unlink:
```
mpawlow: ### (1b) [Minor] New tests are implicitly environment-dependent — `os.path.exists` is not mocked

#### Problem

- Several new tests mock `safe_unlink` (preventing actual deletion) but do NOT mock `os.path.exists`
- After `safe_unlink` (a no-op mock), `_cleanup_upload_temp_files` calls `os.path.exists(temp_path)` for real on the filesystem
- If a hardcoded path like `/var/folders/tmp/corrupt.tmp` or `/var/folders/tmp/success.tmp` happens to exist on the test machine, `os.path.exists` returns `True` and the path is retained
- This would cause `assert task.temp_file_paths == []` to fail, making the test non-deterministic

#### Code References

- `tests/unit/test_task_service_retry_failed_files.py:291–295` — `test_cleanup_upload_temp_files_keeps_retryable_local_failure` (mocks `safe_unlink`, no `os.path.exists` mock for `other_path`)
- `tests/unit/test_task_service_retry_failed_files.py:310–314` — `test_cleanup_upload_temp_files_removes_non_retryable_failure` (mocks `safe_unlink`, no `os.path.exists` mock)
- `tests/unit/test_task_service_retry_failed_files.py:348–352` — `test_cleanup_upload_temp_files_does_not_unlink_retryable_local_failure` (no `safe_unlink` or `os.path.exists` mock needed for retryable path, but `other_path` not present in this test)

#### Potential Solution

Add `patch("os.path.exists", return_value=False)` to the tests that assert `temp_file_paths == []`, matching the pattern already used in `test_cleanup_upload_temp_files_retains_path_when_unlink_fails`:

```python
def test_cleanup_upload_temp_files_removes_non_retryable_failure(task_service):
    temp_path = "/var/folders/tmp/corrupt.tmp"
    ...
    with (
        patch("utils.file_utils.safe_unlink"),
        patch("os.path.exists", return_value=False),  # file successfully deleted
    ):
        task_service._cleanup_upload_temp_files(task)

    assert task.temp_file_paths == []
```

#### Alternative Solutions

- Use `pytest`'s `tmp_path` fixture to create real temp files, then verify they were actually deleted. This avoids mocking `os.path.exists` entirely and tests real filesystem behavior.

mpawlow (approved):
@Wallgau 

## Code Review 1 

- ✅  Approved / LGTM  🚀 
- See: PR comments: 
   - (1a) Normal severity
   - (1b) Minor severity

lucaseduoli (approved):
Working perfectly. I think Mike's feedbacks are good

PR #1717, opened 2026-06-01
fix: Update ingestion flow configuration to support group acl

edwinjosechittilappilly:
Fix ingestion flow

### DIFF
--- a/flows/ingestion_flow.json
+++ b/flows/ingestion_flow.json
@@ -184,238 +184,6 @@
         "className": "",
         "data": {
           "sourceHandle": {
-            "dataType": "TextInput",
-            "id": "TextInput-CTKlr",
-            "name": "text",
-            "output_types": [
-              "Message"
-            ]
-          },
-          "targetHandle": {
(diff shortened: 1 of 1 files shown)


edwinjosechittilappilly:
@coderabbitai create a follow up Issue to fix the URL ingestion flow and to plan how to use URL ingestion via mcp using the ingest token service.
assign it to @edwinjosechittilappilly  and @phact.
Also add a draft plan and suggestions.

edwinjosechittilappilly:
@coderabbitai can you create an issue on Openrag needs to use the universal JWT make opensearch call and Docling
for openserach on Saas we already use the JWT, similarly we need to use JWT for docling in Authorization header, I believe it is in place wrt @lucaseduoli  previous PR, update the docling serve component to accept JWT headers.

edwinjosechittilappilly:
@coderabbitai create an issue: enable OpenRAG and Langflow dynamically switch different Docling URLs.
By making those URLs as global variable passthrough for ingestion pipelines in headers.

lucaseduoli (approved):

PR #1725, opened 2026-06-01
fix: Address LGPL License concerns

mfortman11:
Address LGPL License concerns around sharp (a package brought in by next.js but we don't use)

## Summary by CodeRabbit

* **Chores**
  * Disabled built-in image optimization to simplify the build environment.


edwinjosechittilappilly (approved):
lgtm

PR #1726, opened 2026-06-01
fix: make sure feedback state doesn't remain accross chats

mfortman11:
To test:

Create chat and get a response.
React to the response using the thumbs up or down
Go look at another chat and ensure that a feedback response doesn't carry over to the other chat

## Bug Fixes
* Improved message rendering reliability and consistency when switching between different conversations in the chat interface.


Wallgau (approved):
LGTM

PR #1736, opened 2026-06-02
feat: Make providers admin-only and enforce RBAC

edwinjosechittilappilly:
Revoke non-admin provider permissions and enforce admin-only provider access across backend and frontend. Adds an Alembic migration (0006_revoke_provider_override_nonadmin) to delete the previously-seeded providers:override:self joins for developer and user roles and updates the DB seed to stop granting provider perms to those roles. Backend: gates provider-related endpoints and v1 APIs behind providers:read / providers:write (and API-key equivalents), redacts the providers section in settings for callers without providers:read, and requires providers:write when updating provider fields. Frontend: avoids polling provider health for non-admins when RBAC is enforced, adds an OnboardingBlocked UI for non-admins when onboarding is required, prevents non-admin access to the Langflow tab (config:write), and waits for permission resolution to avoid flashing the admin wizard.

## Summary by CodeRabbit

* **New Features**
  * Non-admin users see an onboarding-blocked screen (with refresh) until setup completes; a loader appears while permissions resolve.

* **Security**
  * Provider configuration, provider-health polling, model listing, and settings endpoints now enforce provider-specific permissions; settings responses redact provider details for unauthorized callers.
  
* **Chores**
  * Database migration and seed updates revoke non-admin provider override access.


mfortman11:
Screenshot 2026-06-04 at 12 43 31 PM
 blocked state update

mfortman11 (approved):

PR #1760, opened 2026-06-04
fix: static sec findings / operator token auto-mounting and cpu/mem res limits

rodageve:
**Description:**
- Cherry picked fix for - static sec findings / operator token auto-mounting and cpu/mem res limits


edwinjosechittilappilly (approved):
LGTM

PR #1781, opened 2026-06-05
fix: target agent component when updating chat flow system prompt

### LINKED ISSUE #1514: [Bug]: Updating Agent Instructions in settings doesn't have any effect
### OpenRAG Version

0.4.1

### Deployment Method

uv add (installed in project)

### Operating System

macOS 26.3.1

### Python Version

3.13.13

### Affected Area

Settings (configuration, model providers)

### Bug Description

Updating the agent instructions in the OpenRAG settings does not change the agent behaviour in Langflow.

From inspecting the requests to and from the back end, it seems that the updated agent instructions are stored somewhere. But they are not applied to the Langflow flow when the agent is sent a message.

### Steps to Reproduce

1. Open the OpenRAG settings.
2. Add something obvious to the agent instructions, like "You only talk like a pirate"
3. Open a new chat and send a message to the agent
4. Observe that the agent does not talk like a pirate

### Expected Behavior

Updating the agent instructions should have an impact on the behaviour of the agent.

### Actual Behavior

No impact.

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

Vchen7629:
## Summary
Fixes https://github.com/langflow-ai/openrag/issues/1514. update_chat_flow_system_prompt  was targeting the LLM provider component (e.g. "Ollama", "WatsonX") instead of the Agent component where the system_prompt field lives. As a result, saving new agent instructions in OpenRAG settings never updates langflow

## Changes:
- Updated update_chat_flow_system_prompt to use AGENT_COMPONENT_DISPLAY_NAME as the target component
- Added unit tests verifying the value is written to the Agent component in the flow

## Summary by CodeRabbit

* **Improvements**
  * Simplified system prompt updates for Langflow flows.

* **Tests**
  * Added unit tests to verify system prompt functionality in Langflow flows.

### DIFF
--- a/src/services/flows_service.py
+++ b/src/services/flows_service.py
@@ -673,16 +673,14 @@ class FlowsService:
         await self._update_flow_field(LANGFLOW_CHAT_FLOW_ID, "model_name", model_name,
                                 node_display_name=target_llm_id)
 
-    async def update_chat_flow_system_prompt(self, system_prompt: str, provider: str):
+    async def update_chat_flow_system_prompt(self, system_prompt: str):
         """Helper function to update the system prompt in the chat flow"""
         if not LANGFLOW_CHAT_FLOW_ID:
             raise ValueError("LANGFLOW_CHAT_FLOW_ID is not configured")
 
-        # Determine target component IDs based on provider
-        target_agent_id = self._get_provider_component_ids(provider)[1]
-
-        await self._update_flow_field(LANGFLOW_CHAT_FLOW_ID, "system_prompt", system_prompt,
-                                node_display_name=target_agent_id)
+        await self._update_flow_field(
+            LANGFLOW_CHAT_FLOW_ID, "system_prompt", system_prompt, node_display_name=AGENT_COMPONENT_DISPLAY_NAME
+        )
 
     async def update_flow_docling_preset(self, preset: str, preset_config: dict):
         """Helper function to update docling preset in the ingest flow"""
--- a/src/api/settings/langflow_sync.py
+++ b/src/api/settings/langflow_sync.py
@@ -1,0 +1,307 @@
+"""Langflow synchronization helpers for settings/onboarding flows.
+
+Pushes the latest user config (provider creds, model selection, system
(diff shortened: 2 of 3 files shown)


src/services/flows_service.py
```diff
@@ -705,16 +705,13 @@ async def update_chat_flow_model(self, model_name: str, provider: str):
             LANGFLOW_CHAT_FLOW_ID, "model_name", model_name, node_display_name=target_llm_id
         )
 
-    async def update_chat_flow_system_prompt(self, system_prompt: str, provider: str):
+    async def update_chat_flow_system_prompt(self, system_prompt: str):
```
Wallgau, line 676: Nice catch on targeting the Agent node. I'd suggest keeping the provider parameter and only fix the lookup.
_get_provider_component_ids(provider)[1] points at the LLM component, not Agent, that’s the bug. 
You can use AGENT_COMPONENT_DISPLAY_NAME fixes it without changing the method signature.

Wallgau (changes requested):

src/services/flows_service.py
Vchen7629: Thanks for the suggestion, just wondering if there is a planned use for the provider parameter going forward?  The parameter is unused in the method body with my changes so keeping it might be misleading for future readers.

Wallgau: Fair point, I was leaning toward keeping provider to keep the diff small and avoid touching the call site/tests, since the real bug is the Agent lookup, not the signature. But you're right: it is not used so the target it’s dead weight. Removing it is cleaner.
Please just double-check nothing else calls update_chat_flow_system_prompt(..., provider) before dropping it Thanks!

Wallgau (approved):
Great work!

src/services/flows_service.py
Vchen7629: yeah only the _update_langflow_system_prompt method in langflow_sync.py calls it so it should be fine

Vchen7629:
Closed since I created a new pr #1918 to move my changes from my fork branch to main repo branch

PR #1800, opened 2026-06-08
fix: honor langflow command and args from CR spec (#1788)

rodageve:
**Description:**
- Cherry pick fix from main

The langflow deployment was ignoring the command and args fields from the OpenRAG CR spec, always using hardcoded values instead. This prevented users from customizing the langflow startup command.

Changes:
- Add commandOrDefault() and argsOrDefault() helper functions
- Update langflowDeployment() to use spec.Command and spec.Args with fallback to default values
- Aligns with existing pattern used by docling-serve and docling-worker

Fixes issue where langflow spec with custom command/args was not applied to the deployment.


edwinjosechittilappilly (approved):
LGTM

zzzming (approved):

PR #1801, opened 2026-06-08
fix: remove clickhouse connect from langflow image

lucaseduoli:
This pull request makes a minor change to the `Dockerfile.langflow` to address package management. The main update is the removal of the `clickhouse-connect` package after installation.

- Dependency management:
  * Uninstalls the `clickhouse-connect` package after all dependencies are installed to ensure it is not present in the final Docker image.


edwinjosechittilappilly (approved):
pre empt approval. Merge only after test in CPD

PR #1802, opened 2026-06-08
fix: remove clickhouse connect from Langflow image

lucaseduoli:
This pull request makes a minor change to the `Dockerfile.langflow` to address a dependency issue. Specifically, it ensures that the `clickhouse-connect` package is uninstalled after installing the required dependencies, likely to avoid conflicts or unnecessary bloat in the final image.

* Uninstalls the `clickhouse-connect` package after the main dependencies are installed to prevent potential issues with this package in the runtime environment.

## Release Notes

* **Chores**
  * Updated the container’s Python dependency setup during image build.
  * Added `uv` to improve dependency management in the build process.
  * Removed `clickhouse-connect` after installation to ensure the final image has an updated package set.


edwinjosechittilappilly (approved):
LGTM. 

Hey is it a CVE fix? will it prevent langflow to load?

PR #1814, opened 2026-06-09
fix: preserve HTTPException status in update_docling_preset (#1586)

### LINKED ISSUE #1586: bug: HTTPException(400) swallowed as 500 in update_docling_preset
### Description

In `src/api/settings/endpoints.py`, the `update_docling_preset` endpoint raises an `HTTPException(status_code=400)` for invalid preset values, but the broad `except Exception` block at the end of the function catches it and re-raises it as an `HTTPException(status_code=500)`. This means clients receive a 500 Internal Server Error instead of the intended 400 Bad Request.

### Affected Code

**File:** `src/api/settings/endpoints.py` (~lines 1500–1509)

```python
    except Exception as e:
        logger.error("Failed to update docling settings", error=str(e))
        raise HTTPException(
            status_code=500, detail=f"Failed to update docling settings: {str(e)}"
        ) from e
```

### Proposed Fix

Re-raise `HTTPException` immediately before the generic handler so the intended 400 status codes are preserved:

```diff
-    except Exception as e:
+    except HTTPException:
+        raise
+    except Exception as e:
         logger.error("Failed to update docling settings", error=str(e))
         raise HTTPException(
-            status_code=500, detail=f"Failed to update docling settings: {str(e)}"
+            status_code=500, detail="Failed to update docling settings"
         ) from e
```

### 🤖 Prompt for AI Agents

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `src/api/settings/endpoints.py` around lines 1500 - 1504, The broad except
block is converting intended HTTPException(400) into a 500; update the error
handling in the endpoint in src/api/settings/endpoints.py so that if the caught
exception is an HTTPException it is immediately re-raised (e.g. check
isinstance(e, HTTPException) and raise), otherwise log the error with
logger.error and raise a new HTTPException(status_code=500, ...) from e; keep
references to logger and HTTPException in the same handler so the original 400
responses for invalid docling presets are preserved.
```

### References

- **PR:** langflow-ai/openrag#1570
- **Review comment:** https://github.com/langflow-ai/openrag/pull/1570#discussion_r3219472252
- **Reported by:** @edwinjosechittilappilly

hunterxtang:
Fixes #1586

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for the docling-preset endpoint: client-side issues (such as invalid preset values) now preserve the original HTTP status codes and messages instead of being converted into a generic server error.
* **Tests**
  * Added a unit test to confirm invalid preset requests return a 400 response with the expected error detail.


Vchen7629 (approved):
LGTM

src/api/settings/endpoints.py
mpawlow: ### (1a) [Normal] Raw exception message leaked in 500 response `detail` string

#### Problem

- Embedding `str(e)` in API response bodies can expose internal implementation details (file paths, library internals, config values, stack-frame fragments) to any caller — a recognized security anti-pattern (OWASP A05:2021 — Security Misconfiguration; CWE-209: Information Exposure Through an Error Message).
- The error is already fully captured server-side via `logger.error("Failed to update docling settings", error=str(e))`, so no diagnostic value is lost by removing `str(e)` from the detail.
- Note: This was explicitly called out in the linked issue's own proposed fix (#1586):
  ```diff
  -            status_code=500, detail=f"Failed to update docling settings: {str(e)}"
  +            status_code=500, detail="Failed to update docling settings"
  ```
  The PR applied the `except HTTPException: raise` part of that fix but did not carry forward this secondary hardening.

#### Code References

- `src/api/settings/endpoints.py:1630` — primary (this PR's scope)

#### Potential Solution

Strip the raw exception from the client-facing detail string:

```python
except Exception as e:
    logger.error("Failed to update docling settings", error=str(e))
    raise HTTPException(
        status_code=500, detail="Failed to update docling settings"
    ) from e
```

src/api/settings/endpoints.py
```diff
@@ -1621,7 +1621,9 @@ async def update_docling_preset(
             settings=settings_toggles,
             preset_config=preset_config,
         )
-
+    except HTTPException:
+        # Preserve intended HTTP status codes (e.g. 400 for an invalid preset)
+        raise
```
mpawlow, line 1602: ### (1b) [Minor] No regression test added to guard the fixed behavior

#### Problem

- The PR fixes a bug where an invalid preset silently produced a 500 instead of the intended 400, but no test was added to assert the corrected behavior.
- Without a test, any future refactor that reorders or widens the `except` chain could silently reproduce the original bug with no CI signal.
- A unit test calling `update_docling_preset` directly with an invalid preset body and asserting `status_code=400` would require no infra and would be straightforward to add.

#### Code References

- `src/api/settings/endpoints.py:1590–1594` — the `raise HTTPException(status_code=400, ...)` path the fix protects
- `tests/unit/test_docling_service.py` — closest existing test file; an additional test belongs here or in a new `test_settings_endpoints.py`

#### Potential Solution

```python
import pytest
from fastapi import HTTPException
from unittest.mock import AsyncMock

@pytest.mark.asyncio
async def test_update_docling_preset_invalid_preset_returns_400():
    from api.settings.endpoints import update_docling_preset
    from api.settings.models import DoclingPresetBody

    body = DoclingPresetBody(preset="nonexistent_preset")

    with pytest.raises(HTTPException) as exc_info:
        await update_docling_preset(body=body, session_manager=AsyncMock(), user=AsyncMock())

    assert exc_info.value.status_code == 400
```

mpawlow (changes requested):
Good job @hunterxtang 

## Code Review 1

- See PR comments: (1a), (1b) for relative easy tweaks / fixes

hunterxtang:
1a) Removed `str(e)` from the client-facing `detail` so the 500 no longer leaks internal exception text; kept `str(e)` in the `logger.error` call for server-side diagnostics, and preserved the `from e` chaining. The detail is now a static `"Failed to update docling settings"`.

1b) Added `test_update_docling_preset_invalid_preset_returns_400` in `tests/unit/api/test_settings_endpoints.py`. It sends an invalid preset and asserts the call raises `HTTPException` with `status_code == 400`. To confirm it actually guards the fix, I verified it fails (returns 500) when the `except HTTPException: raise` block is removed, then restored the block. Any future refactor that reorders or widens the except chain will trip CI, which was the concern here.

mpawlow:
Thanks @hunterxtang 

- Changes look good  👍 
- Running remaining workflows...
   - Waiting to see if E2E Tests / Integration Tests pass...

mpawlow:
> Error: OPENAI_API_KEY is not set

- Integration tests are failing
- I think this is a workflow error related to this being a forked Pull Request
- Following up with the rest of the team on this...

mpawlow:
> Error: 6-12T17:26:14,233][ERROR][o.o.s.a.BackendRegistry  ] [504d3485a6c9] OpenSearch Security not initialized. (you may need to run securityadmin)

> Error: OPENAI_API_KEY is not set

- ❌  Suspicion: Forked PR causing the E2E tests and integration tests to fail

CC @Wallgau @lucaseduoli

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM 🚀 
- ⚠️ Need to resolve E2E / Integration test failures in the workflow before merge (potential infrastructure problems)

mpawlow:
🚧  Re-running the GitHub workflow to see if the E2E / Integration test failures are resolved

mpawlow:
Still failing... 
- This has got to be the cause --> https://github.com/langflow-ai/openrag/pull/1814#issuecomment-4693882360

@Wallgau @lucaseduoli 
- `Question`: Should we just forcibly merge this forked PR in the interim OR attempt to resolve the underlying problem of forked PR access to the OpenAI credentials within the GitHub Action context?

Wallgau:
> Still failing...
> 
> * This has got to be the cause --> [fix: preserve HTTPException status in update_docling_preset (#1586) #1814 (comment)](https://github.com/langflow-ai/openrag/pull/1814#issuecomment-4693882360)
> 
> @Wallgau @lucaseduoli
> 
> * `Question`: Should we just forcibly merge this forked PR in the interim OR attempt to resolve the underlying problem of forked PR access to the OpenAI credentials within the GitHub Action context?

we should probably create a new branch from the release one and apply changes there, @lucaseduoli ?

mpawlow:
> Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.

- See: https://github.com/langflow-ai/openrag/settings/secrets/actions
- There is supposed to be a setting to override this in `Settings > Actions > General > Fork pull request workflows`
   - Then enable a **Send secrets to workflows from pull requests** option
- However, I don't see it. It is likely defined at the langflow GitHub org level (which I don't have admin access to)
    - https://github.com/organizations/langflow-ai/settings

PR #1815, opened 2026-06-09
fix: report actual synced connection count in connector_sync (#1547)

### LINKED ISSUE #1547: bug: sync_all_connectors reports inflated connections_synced count
## Summary

`sync_all_connectors` in `src/api/connectors.py` returns `connections_synced: len(active_connections)` even though only a single `task_id` is actually dispatched and returned per call. Callers that use `connections_synced` to verify sync completeness will receive an inflated count that does not reflect the number of connectors actually synced.

## Location

`src/api/connectors.py` — around line 311 (pre-existing on `main`)

## Expected behaviour

`connections_synced` should reflect the number of connectors for which a sync task was successfully started (i.e. `len(task_ids)` or the actual dispatched count).

## Actual behaviour

The field is set to `len(active_connections)`, which is the total number of active connections regardless of how many were actually processed in the current call.

## Impact

Misleads callers/operators into believing more connectors were synced than actually were. Could mask partial-sync failures in observability dashboards or automated checks.

## References

- Identified during review of PR #1538 (https://github.com/langflow-ai/openrag/pull/1538) by @edwinjosechittilappilly
- Pre-existing issue on `main`, not introduced by that PR

hunterxtang:
Fixes #1547
connector_sync reported connections_synced: len(active_connections), but it selects only the first working connection and dispatches a single task, so the count (and the message) were inflated whenever multiple connections were active. Changed to len(task_ids) and fixed the matching message to singular.
Added test_connector_sync_reports_one_connection_with_multiple_active: with 3 active connections it asserts connections_synced == 1. Verified it fails on main (reports 3) and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Connector sync responses now correctly report the number of connections started based on returned task IDs.
  * Response messaging uses singular wording when a single connection is started for clearer, grammatically correct feedback.

* **Tests**
  * Added unit tests validating HTTP status, task ID reporting, and correct singular/plural messaging in connector sync responses.


richardz403 (approved):
Ran the unit test test_connector_sync_response and the changes seem to be in order.

PR #1818, opened 2026-06-09
fix: use run-mode-aware admin client for init_index instead of the end-user JWT client

edwinjosechittilappilly:
In SaaS with IBM auth, _ensure_index_exists built the OpenSearch client from the end-user's JWT. That identity can search/write documents but lacks index-admin privileges on managed OpenSearch, so the first admin call in init_index (HEAD /<index> via indices.exists) failed with TransportError(500, '') and connector sync returned 500.

Add Clients.create_index_admin_opensearch_client, mirroring the onboarding client selection: saas -> platform service token (user-JWT fallback for legacy deployments without OPENRAG_SERVICE_TOKEN), on_prem/oss -> OpenSearch basic auth. Use it in _ensure_index_exists (fixes connector sync, traditional uploads, and router ingest in one place) and fold the onboarding block in settings/endpoints.py onto the shared helper.

## Summary by CodeRabbit

* **Chores**
  * Improved OpenSearch client initialization and authentication handling across deployment modes to streamline index setup operations and enhance credential management consistency.

* **Tests**
  * Added comprehensive unit tests for client selection logic across different deployment configurations.

### DIFF
--- a/src/api/documents.py
+++ b/src/api/documents.py
@@ -178,14 +178,14 @@ async def delete_chunks_by_document_ids(
 
 async def _ensure_index_exists(jwt_token: str = None):
     """Create the OpenSearch index if it doesn't exist yet."""
-    from config.settings import IBM_AUTH_ENABLED
     from config.settings import clients as app_clients
     from main import init_index
 
-    opensearch_client = None
-    if IBM_AUTH_ENABLED and jwt_token:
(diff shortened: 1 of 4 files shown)


ricofurtado (approved):
Liked the centralized OpenSearch admin client selection logic

PR #1823, opened 2026-06-10
refactor: effect cleanup for timeouts

mfortman11:
`setTimeout` /listeners in `useEffect` with no cleanup → memory leaks and late callbacks. `chat/page.tsx` (×2), `auth/callback/page.tsx`, onboarding, `agent-settings-section.tsx`.

## Release Notes

* **Bug Fixes**
  * Improved OAuth authentication redirect handling to prevent delayed navigation after the page lifecycle ends.
  * Refined chat input focus timing so focus won’t occur after conversation state changes or during unmount.
  * Stabilized onboarding completion behavior by clearing pending completion callbacks when effects re-run or are cancelled.
  * Fixed settings timing issues by ensuring any delayed UI actions are properly cleaned up when the component re-renders or is left.


frontend/app/onboarding/_components/onboarding-upload.tsx
```diff
@@ -161,7 +162,7 @@ const OnboardingUpload = ({ onComplete }: OnboardingUploadProps) => {
             refetchNudges();
 
             // Wait a bit before completing (after filter is created)
-            setTimeout(() => {
+            completeTimeoutId = setTimeout(() => {
```
Wallgau: Same as auth callback, suggest cancelled flag + guard before/inside setTimeout

frontend/app/auth/callback/page.tsx
```diff
@@ -158,7 +160,7 @@ function AuthCallbackContent() {
             localStorage.removeItem("auth_redirect_to");
 
             // Redirect to the original page or home
-            setTimeout(() => {
+            redirectTimeoutId = setTimeout(() => {
```
Wallgau: The cleanup runs before this timer is created, because the timer is set after await fetch and await refreshAuth(). If the user leaves the page early, cleanup does nothing and the redirect can still happen later.
I'd suggest to use a cancelled flag: set it to true in cleanup, and check it before starting the timer and inside the timer callback.

frontend/app/onboarding/_components/onboarding-upload.tsx
```diff
@@ -171,11 +172,13 @@ const OnboardingUpload = ({ onComplete }: OnboardingUploadProps) => {
         refetchNudges();
```
Wallgau, line 175: We should only call onComplete() when isCreatingFilter is false. Use else if (!isCreatingFilter) instead of else, and let .finally() call onComplete() after filter creation finishes.
} else if (!isCreatingFilter) {
 refetchNudges();

Wallgau (approved):
LGTM, good cleanup for the sync timeout cases. Just few notes

PR #1824, opened 2026-06-10
fix: a11y fixes

mfortman11:
aria-controls on comboboxes

  When you put role="combobox" on a button, screen readers announce it as a combo box — a text-field-plus-dropdown pattern. The ARIA spec requires aria-controls to point to the ID of the popup list so assistive tech
  can link the two: "this button controls that list." Without it, a screen reader user knows a combobox exists but can't navigate to its options.

  The fix: generate a stable unique ID with useId(), put it as the id on the listbox element (CommandList / CommandGroup), and reference it via aria-controls={listboxId} on the button.
  
prefers-reduced-motion
  
  Some users have vestibular disorders where animations cause nausea or dizziness. They set "reduce motion" in their OS settings, which the browser exposes as prefers-reduced-motion: reduce. WCAG 2.3.3 requires sites
  to respect this.

  The project uses Framer Motion extensively but never checked this preference — every animation played regardless. The fix is a global CSS rule that forces all animations and transitions to essentially 0ms duration
  when the user has requested reduced motion. It's a single catch-all that covers both CSS animations and Framer Motion (which uses inline styles that inherit transition-duration behavior from the cascade when
  !important is used).

  ---
  The ESLint warning on allOptions at line 71 is pre-existing and unrelated to our changes — it's about the useEffect at line 90 that was already there.

## Summary by CodeRabbit

* **Accessibility**
  * Added support for users' reduced motion system preferences to minimize animations and transitions across the application, improving overall usability for those with motion sensitivities or vestibular disorders
  * Enhanced accessibility infrastructure with improved ARIA attributes on form controls and selection components, providing better support and compatibility for screen readers and other assistive technologies


Wallgau (approved):
LGTM

PR #1827, opened 2026-06-10
refactor: Don't return full tanstack query result

mfortman11:
When you write `const queryResult = useQuery(...)`, TanStack Query hands you the full result object — but it has no way to know which fields you actually use. So it conservatively re-renders your component whenever any field changes (`fetchStatus`, `dataUpdatedAt`, internal counters, etc.), even if you only care about `data` and `isLoading`.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined multiple data-fetching hooks in the query layer to return results directly, removing intermediate result handling.
  * Updated the nudges query to return a focused set of commonly used state values instead of the full query object.
  * Kept existing query behavior, caching, and configuration the same, with no expected user-facing impact.


Wallgau (approved):
Niice

PR #1842, opened 2026-06-11
fix: put correct file type in langflow-less ingestion

lucaseduoli:
This pull request improves how temporary files are handled during file uploads by preserving the original file extensions, which helps with format detection. It also adds unit tests to ensure this behavior for both traditional and langflow upload ingestion tasks.

**File Upload Improvements:**
- Temporary files are now created with the original file extension (if available) instead of always using `.tmp`, improving compatibility with downstream format detection [[1]](diffhunk://#diff-585a2e9d897ffbd47df61ed9efb432c1c7193e4d88cb147cd5e84e5b4ca3a9faL108-R113) [[2]](diffhunk://#diff-585a2e9d897ffbd47df61ed9efb432c1c7193e4d88cb147cd5e84e5b4ca3a9faL209-R217).

**Testing Enhancements:**
- Added and updated unit tests in `test_disable_ingest_setting.py` to verify that the file extension is preserved for both traditional and langflow upload ingestion tasks, and to clean up temporary files after tests.

**Dependency Updates:**
- Added `os` import to `router.py` to support file extension extraction.

## Summary by CodeRabbit

* **Bug Fixes**
  * File uploads now preserve original file extensions during processing, with automatic fallback for files that lack an extension.

* **Tests**
  * Added/updated unit tests to verify preserved extensions and fallback behavior for both traditional and Langflow-style uploads.

### DIFF
--- a/src/api/router.py
+++ b/src/api/router.py
@@ -1,6 +1,7 @@
 """Router endpoints that automatically route based on configuration settings."""
 
 import json
+import os
 import tempfile
 
 from fastapi import Depends, File, Form, UploadFile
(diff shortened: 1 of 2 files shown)


src/api/router.py
```diff
@@ -105,8 +106,11 @@ async def _traditional_upload_ingest_task(
             for upload_file in upload_files:
                 content = await upload_file.read()
                 original_filenames.append(upload_file.filename)
-                # Generate unique temp file to avoid collisions
-                temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=".tmp")
+                # Generate unique temp file with the original extension to assist docling/format detection
+                suffix = os.path.splitext(upload_file.filename)[1] if upload_file.filename else ""
+                if not suffix:
+                    suffix = ".tmp"
```
edwinjosechittilappilly, line 119: @lucaseduoli  .tmp suffix will fail by default

edwinjosechittilappilly (approved):
Code LGTM

PR #1868, opened 2026-06-12
fix: post-ingestion verification must not fail files on OpenSearch 0utages

edwinjosechittilappilly:
Integration test fix

#1851 added a "verify the document actually landed in OpenSearch" check after Langflow ingestion. check_document_exists returns False when OpenSearch stays unreachable after retries — the right call for its dedupe callers (reprocess rather than skip), but the verification callers read that same False as "chunks never landed" and marked the file FAILED. A transient 401 (OpenSearch loads the backend JWKS lazily, so the first user-JWT queries after a cold start fail) then failed every in-flight file — this is what broke test_onboarding_sample_docs.py::test_onboarding_ingests_sample_docs_and_creates_openrag_docs_filter across all PRs since #1851 merged.

   - check_document_exists grows an on_error mode: dedupe callers keep "assume_missing"; the two verification call sites use "assume_exists" so an infrastructure error is inconclusive, not fatal.
   - Verification now prefers the platform writer client (_verification_client): "did the chunks land in the index" is a system integrity check, not a user-visibility check, so it should not depend on the user JWT/JWKS trust chain at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ingestion verification to avoid marking files as failed when search service is temporarily unreachable; verification now errs on the side of assuming presence in flaky connectivity scenarios.
  * Prefer the platform search service when available to make post-ingestion checks more resilient.

* **Tests**
  * Added tests covering verification behavior across connection scenarios and the service-selection fallback.

### DIFF
--- a/src/models/processors.py
+++ b/src/models/processors.py
@@ -2,7 +2,7 @@ import asyncio
 import mimetypes
 import os
 import time
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 from config.settings import clients, get_embedding_model, get_index_name, get_openrag_config
 from utils.document_processing import (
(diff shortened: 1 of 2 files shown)


edwinjosechittilappilly:
@coderabbitai please review this PR.

Wallgau (approved):
LGTM

PR #1883, opened 2026-06-15
fix: check backend health inside container

### LINKED ISSUE #1880: [Bug]: make health reports backend as Not responding when backend container is healthy
### OpenRAG Version

0.5.0

### Deployment Method

Podman

### Operating System

macOS Tahoe 26.5.1

### Python Version

3.14.5

### Affected Area

TUI (Terminal User Interface)

### Bug Description

`make health` says the backend is `Not responding`, even though the backend container is actually healthy.

In my Podman setup, the backend port is not published to the host, so this check fails:

```bash
curl http://127.0.0.1:8000/health
```

But checking the backend from inside the container works:

```bash
podman exec openrag-backend curl http://localhost:8000/health
```

and returns:

```json
{"status":"ok"}
```

So `make health` is giving a false negative for the backend. It should either check the backend inside the container or handle setups where the backend port is not published to the host.

### Steps to Reproduce

1. Start OpenRAG using Podman.
2. Run:
make health
3. Notice that the backend is reported as not responding:
Backend:    Not responding
4. Check the backend container directly:
podman exec openrag-backend curl http://localhost:8000/health
5. Notice that the backend is actually healthy:

```json
{"status":"ok"}
```
6. Check the backend container ports:
podman ps

The backend shows internal ports like `8000/tcp`, but not a host mapping like `0.0.0.0:8000->8000/tcp`.

This means `make health` is checking the backend from the host, but the backend is only reachable inside the container.

### Expected Behavior

Make health should return Backend: Healthy

### Actual Behavior

Make Health returns Backend: Not responding

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

NishadA05:
Fixes #1880 

## Summary

Updates `make health` so the backend check first tries the existing host health endpoint, then falls back to checking the `/health` endpoint inside the running `openrag-backend` container using Podman or Docker.

This fixes a false negative where `make health` reports:

```text
Backend:    Not responding

even though the backend container is healthy and returns:

{"status":"ok"}
Testing

Tested with backend running:

make health

Result:

Backend:    Healthy

Tested with backend stopped:

podman stop openrag-backend
make health

Result:

Backend:    Not responding

Restarted backend and confirmed health check returns healthy again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1003,7 +1003,15 @@ health: ## Check health of all services
 	@printf "$(CYAN)Frontend:$(NC)   "
 	@if curl -s -k --fail http://127.0.0.1:$${FRONTEND_PORT:-3000}/ >/dev/null 2>&1; then printf "$(GREEN)Healthy$(NC)\n"; else printf "$(RED)Not responding$(NC)\n"; fi
 	@printf "$(CYAN)Backend:$(NC)    "
-	@if curl -s -k --fail http://127.0.0.1:8000/health >/dev/null 2>&1; then printf "$(GREEN)Healthy$(NC)\n"; else printf "$(RED)Not responding$(NC)\n"; fi
+	@if curl -s -k --fail http://127.0.0.1:8000/health >/dev/null 2>&1; then \
+		printf "$(GREEN)Healthy$(NC)\n"; \
+	elif command -v podman >/dev/null 2>&1 && podman ps --format "{{.Names}}" | grep -q "^openrag-backend$$" && podman exec openrag-backend curl -s -k --fail http://127.0.0.1:8000/health >/dev/null 2>&1; then \
+		printf "$(GREEN)Healthy$(NC)\n"; \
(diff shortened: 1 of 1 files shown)


hunterxtang (approved):
LGTM

Verified locally:
- The hardcoded `openrag-backend` matches `container_name: openrag-backend` in `docker-compose.yml`. Since `container_name` is set explicitly, neither runtime applies a project prefix, so the anchored `grep -q "^openrag-backend$"` matches exactly. The TUI compose is a symlink to the root file, so the Podman/TUI path uses the same name.
- `curl` is present in `Dockerfile.backend`, so the `exec` branch genuinely succeeds (consistent with the manual `podman exec` in the issue).
- Shell logic is POSIX-valid (`sh -n` clean) and works under Make's default `/bin/sh`; tabs/continuations are correct.
- Applies cleanly to the current `release-0.5.1` tip.
- `--fail` correctly treats non-2xx as unhealthy, and the `ps | grep` guard avoids a confusing error when the container isn't running. Trying Podman before Docker is right for the reported macOS/Podman setup.

yogichipalkatti (commented):
Looks good to me. 

Restarted the backend locally and confirmed the health check correctly returns "Healthy" again.
- Eliminates false negative health reports for the backend service
- Improves reliability of the make health diagnostic command
- Helps operators accurately assess system status

PR #1884, opened 2026-06-15
fix: improve pagination and scroll to the file row that is currently processing the ingestion

Wallgau:
https://github.com/user-attachments/assets/2706ad9b-db9f-4390-a08e-3dbae615d789

Summary
When ingesting files on the Knowledge page, the grid now automatically jumps to the correct pagination page so users can see processing status without manually hunting through pages.

Overwrite (file already indexed): navigates to the page containing that file’s existing row
New file: navigates to the last page where the processing overlay is appended
Cloud uploads (Google Drive, OneDrive, SharePoint): focus targets persist across the redirect from /upload/[provider] → /knowledge
Problem
The Knowledge table uses AG Grid with client-side pagination (25 rows/page). After uploading or overwriting a file, the grid stayed on the current page — e.g. staying on page 1 while an overwritten file on page 2 entered processing. Local uploads and cloud connector uploads both had this gap; cloud uploads additionally lost focus events because navigation unmounted the Knowledge page before the event could fire.

Test plan

 Overwrite on another page: On page 1, overwrite a file whose row is on page 2 → grid jumps to page 2, row shows processing

 New local upload: On page 1, upload a new file → grid jumps to last page when processing row appears

 Google Drive upload: Ingest from /upload/google_drive, redirected to Knowledge → grid jumps to correct page

 Google Drive overwrite: Pick an existing file, confirm overwrite → grid jumps to that file’s page

 Already on correct page: Overwrite while viewing the file’s page → stays on same page, status updates

 No duplicate/missing rows: Paginate through all pages after mixed uploads; no ghost or duplicate rows

## Summary by CodeRabbit

* **New Features**
  * Knowledge grid now automatically navigates to and focuses newly added or processing ingest rows during ingestion, including restore after pagination/row updates; focus is triggered from both upload and sync actions.
* **Improvements**
  * Task overlays now match files using multiple filename/source URL variations with priority-based status resolution (processing over failed) for more reliable row mapping.
  * Ingest-focus state is persisted for the session and applied on page load.
* **Other**
  * Added additional AG Grid modules to improve row navigation.
  * Backend connector sync now propagates connector type into file processing.


lucaseduoli (approved):
Tested and approved

PR #1885, opened 2026-06-15
fix: fixed issue #1587

richardz403:
## Description
Fixes issue [#1587](https://github.com/langflow-ai/openrag/issues/1587), where resetting a flow causes the LLM selection to revert to a stale or default model. 

In `src/api/settings/langflow_sync.py`, the fallback branch of `_update_langflow_model_values` (called during `reapply_all_settings()`) was only iterating and refreshing configured embedding providers. It completely bypassed updating the LLM providers, meaning the chat flow never received a `force_llm_update=True` signal.

## Changes
* Modified the fallback block in `src/api/settings/langflow_sync.py` inside `reapply_all_settings()`.
* If everything is blank, it automatically falls back to your system's default agent settings (config.agent) and forces an update (force_llm_update=True) to the chat flow using that default LLM provider and model.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Langflow model synchronization by adding a reliable fallback to the currently configured LLM provider/model when explicit embedding/LLM settings aren’t provided, improving reliability.
* **Infrastructure**
  * Updated Docker Compose to expose the backend service on host port `8000` for easier local access.

### DIFF
--- a/frontend/app/api/queries/useGetSettingsQuery.ts
+++ b/frontend/app/api/queries/useGetSettingsQuery.ts
@@ -76,6 +76,7 @@ export interface Settings {
   };
   localhost_url?: string;
   ingest_via_chat?: boolean;
+  show_provider_ingest_settings?: boolean;
   segment_write_key?: string;
   environment?: string;
 }
--- a/frontend/components/cloud-picker/unified-cloud-picker.tsx
+++ b/frontend/components/cloud-picker/unified-cloud-picker.tsx
@@ -1,6 +1,8 @@
 "use client";
 
 import { useEffect, useState } from "react";
+import { useGetSettingsQuery } from "@/app/api/queries/useGetSettingsQuery";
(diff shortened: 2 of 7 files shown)


NishadA05 (commented):
I checked out this PR locally and reviewed the diff. The change is narrow and limited to `src/api/settings/langflow_sync.py`. It updates the fallback path in `_update_langflow_model_values()` so that when no explicit model/provider overrides are passed, the configured agent LLM provider/model is forced back into the Langflow chat flow with `force_llm_update=True`.

That seems to address the main issue where the fallback path refreshed embedding providers but did not force an LLM update after a flow reset/reapply.

One note: the issue prompt mentioned iterating all configured LLM providers similar to the embedding provider loop. This PR instead updates only the active/default `config.agent` LLM provider/model. That seems sufficient if the goal is to restore the active chat model after reset, but it is narrower than the originally suggested approach.

PR #1886, opened 2026-06-15
refactor: no derived state

mfortman11:
Every fix follows the same React-recommended technique: adjust state during render instead of in a useEffect. Compare the previous value of the upstream data; if it changed, call setState synchronously during render. React batches these updates and re-invokes the component before committing to the DOM, so the stale frame is never painted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved onboarding reliability when switching API-key sources, including provider auto-selection and default model/embedding selection.
  * Fixed model selector popover behavior when the default-open state changes.
  * Improved chat onboarding sync so step and layout reflect settings changes immediately.
  * Refined knowledge search and filters so overridden queries reset and stay aligned with the active filter.
  * Updated navigation conversation auto-selection behavior and expanded function-call results automatically during streaming.
  * Ensured S3 bucket and selected-bucket defaults apply correctly when settings load.


Wallgau (approved):
Great work!

PR #1892, opened 2026-06-16
fix:  overwite duplicate files for connectors

Wallgau:
Summary
Fixes connector file overwrite and duplicate detection when the connector returns a raw filename that differs from the MIME-cleaned name OpenRAG actually indexes (e.g. Google Docs "My Report" → "My Report.pdf"). Aligns backend dedupe/delete/ingest/metadata on file_task.filename, and improves frontend knowledge-table overlay matching so in-flight sync rows merge correctly with indexed rows.

Problem
Connector ingestion cleans filenames before indexing (clean_connector_filename). Duplicate detection, overwrite deletes, and metadata updates were still keyed on document.filename from the connector API. When those names diverged, overwrite could skip re-ingestion or write the wrong filename back onto chunks after ingest.

The knowledge table had a related UI issue: task overlays matched on a single identity key, so rows could fail to merge when filenames differed by extension or alias (.txt/.md, spaces vs underscores). URL-only rows could also generate malformed alias keys when filename transforms were applied to full URLs.

Changes
Backend (src/models/processors.py, src/connectors/service.py)
Use file_task.filename consistently for rename cleanup, duplicate checks, overwrite deletes, Langflow upload, standard ingest, and chunk suffix selection.
Pass connector_type from ConnectorService into ConnectorFileProcessor explicitly.
Add indexed_filename to _update_connector_metadata so post-ingest metadata writes the cleaned name, not the raw connector name.
Remove hardcoded extension validation — unsupported formats are rejected by Docling at ingest time instead of maintaining a static allowlist that would drift from Docling's supported formats.
Frontend (frontend/lib/knowledge-table-state.ts)
Add filename alias matching mirroring backend get_filename_aliases (.txt/.md swap, space/slash variants).
Index and lookup task overlays by alias keys with priority: processing > failed > active.
When merging backend rows with overlays: keep backend filename/source_url paired, surface processing/failed from the task, prefer backend connector_type when set.
Skip filename transforms on HTTP URLs to avoid mangled alias keys; use exact URL match for URL-sourced rows.
Tests
Connector processor: cleaned filename used for dedupe, standard ingest, Langflow upload, and metadata update.
Metadata update: indexed_filename overrides raw document.filename.
Removed test asserting early .exe rejection (validation deferred to Docling).

## Release Notes

* **New Features**
  * Enhanced file matching system that resolves filename variations and URL formats during knowledge operations
  * Improved connector synchronization with more robust filename handling and conflict resolution in metadata updates

* **Tests**
  * Added validation for metadata filename handling in connector operations
  * Added coverage for filename cleaning and deduplication during connector ingestion

### DIFF
--- a/src/models/processors.py
+++ b/src/models/processors.py
@@ -734,51 +734,6 @@ class ConnectorFileProcessor(TaskProcessor):
 
             connector_type = self.connector_type or connection.connector_type
 
-            # Validate file extension early if filename is available
-            VALID_EXTENSIONS = {
-                "adoc",
-                "asciidoc",
-                "asc",
-                "bmp",
-                "csv",
-                "dotx",
-                "dotm",
-                "docm",
-                "docx",
-                "htm",
-                "html",
-                "jpeg",
-                "jpg",
-                "md",
-                "pdf",
-                "png",
-                "potx",
-                "ppsx",
-                "pptm",
-                "potm",
-                "ppsm",
-                "pptx",
-                "tiff",
-                "txt",
-                "xls",
-                "xlsx",
-                "xhtml",
-                "webp",
-            }
-            # Only pre-validate when we have a real filename. When the filename
-            # falls back to the connector file_id (e.g. a deletion event re-added
-            # by sync_specific_files, where no name is known), skip this check so
-            # the deletion reaches the 404 -> chunk-cleanup path below. Files that
-            # still exist are re-validated after download (see below).
-            if file_task.filename and file_task.filename != file_id:
-                ext = file_task.filename.split(".")[-1].lower() if "." in file_task.filename else ""
-                if ext not in VALID_EXTENSIONS:
-                    file_task.status = TaskStatus.FAILED
-                    file_task.error = f"The file '{file_task.filename}' has an incompatible type."
-                    file_task.updated_at = time.time()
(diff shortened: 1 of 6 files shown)


src/models/processors.py
```diff
@@ -730,50 +732,7 @@ async def process_item(self, upload_task: UploadTask, item: str, file_task: File
...
-                "potm",
-                "ppsm",
-                "pptx",
-                "tiff",
-                "txt",
-                "xls",
-                "xlsx",
-                "xhtml",
-                "webp",
-            }
```
mfortman11, line 763: why remove these? It prevents the fast failures

mfortman11 (changes requested):

src/models/processors.py
Wallgau: Because it was Hardcoded, The VALID_EXTENSIONS set in ConnectorFileProcessor had to be kept in sync with Docling manually. There was already a similar list in flows/components/docling_remote.py, and they could drift apart as Docling adds or changes supported formats.

Wallgau: But you might be right, I think we have a slightly different list from docling so let's keep it

mfortman11 (approved):

PR #1912, opened 2026-06-17
fix: add warning logs for 10k aggregation limit in sync operations

yogichipalkatti:

Affected functions:
- get_synced_file_ids_for_connector(): Warns when connector_file_id,
- get_synced_id_to_filename_map(): Warns when document_id aggregation
  reaches the limit

unique values per field

### DIFF
--- a/src/api/connectors.py
+++ b/src/api/connectors.py
@@ -68,7 +68,12 @@ async def get_synced_file_ids_for_connector(
             result.get("aggregations", {}).get("unique_connector_file_ids", {}).get("buckets", [])
         )
         connector_file_ids = [b["key"] for b in connector_file_id_buckets if b["key"]]
-
+        if len(connector_file_id_buckets) == 10000:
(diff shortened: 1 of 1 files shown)


richardz403 (commented):
Looks good to me.

The additional code correctly alerts and warns when connector file IDs, document IDs, or filename aggregations hit 10k items.

Ran locally to check for other conflicting issues, none were found.

Vchen7629 (approved):
LGTM, verified that all the tests pass when running locally

NishadA05 (commented):
This adds useful warning logs when the OpenSearch aggregations hit the 10,000 bucket limit, which makes possible truncation visible instead of silently returning incomplete sync data.

One note: this does not fully remove the 10k limit, but it is a helpful improvement for now.

Wallgau (approved):
Great work, please create an issue for the enhanced solution

PR #1913, opened 2026-06-17
chore: add node check and manager count check for opensearch health

edwinjosechittilappilly:
chore: add node check and manager count check for opensearch health

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -48,6 +48,10 @@ OPENSEARCH_NODE_COUNT_CHECK_ENABLED = os.getenv(
 # Expected cluster size, used only when the node-count check is enabled.
 OPENSEARCH_EXPECTED_NODE_COUNT = get_env_int("OPENSEARCH_EXPECTED_NODE_COUNT", 6)
 OPENSEARCH_EXPECTED_DATA_NODE_COUNT = get_env_int("OPENSEARCH_EXPECTED_DATA_NODE_COUNT", 2)
+# Minimum reachable cluster-manager (master) nodes, gated by the same flag.
+OPENSEARCH_EXPECTED_CLUSTER_MANAGER_COUNT = get_env_int(
+    "OPENSEARCH_EXPECTED_CLUSTER_MANAGER_COUNT", 2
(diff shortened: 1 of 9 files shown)


src/config/settings.py
```diff
@@ -39,6 +39,23 @@
...
+OPENSEARCH_NODE_COUNT_CHECK_ENABLED = os.getenv(
+    "OPENSEARCH_NODE_COUNT_CHECK", "true"
+).strip().lower() in ("true", "1", "yes")
+
+# Expected cluster size, used only when the node-count check is enabled.
+OPENSEARCH_EXPECTED_DATA_NODE_COUNT = get_env_int("OPENSEARCH_EXPECTED_DATA_NODE_COUNT", 2)
+# Minimum reachable cluster-manager (master) nodes, gated by the same flag.
+OPENSEARCH_EXPECTED_CLUSTER_MANAGER_COUNT = get_env_int(
+    "OPENSEARCH_EXPECTED_CLUSTER_MANAGER_COUNT", 2
+)
```
jimdickinson, line 53: Why not have 3 here?

src/config/settings.py
```diff
@@ -39,6 +39,23 @@
...
+# Minimum reachable coordinating-only nodes, gated by the same flag.
+OPENSEARCH_EXPECTED_COORDINATING_NODE_COUNT = get_env_int(
+    "OPENSEARCH_EXPECTED_COORDINATING_NODE_COUNT", 2
+)
+
```
jimdickinson, line 62: Not needed IMO

edwinjosechittilappilly: Should i keep or remove it?

src/config/settings.py
edwinjosechittilappilly: Done Made addl node count by default to be 3

edwinjosechittilappilly:
Changed DEfaults node check to False for docker compose and e2e tests to unblock CI and for OSS compatibility

rodageve (approved):

zzzming (approved):

PR #1918, opened 2026-06-18
fix: target agent component when updating chat flow system prompt

Vchen7629:
## Summary
Fixes https://github.com/langflow-ai/openrag/issues/1514. update_chat_flow_system_prompt  was targeting the LLM provider component (e.g. "Ollama", "WatsonX") instead of the Agent component where the system_prompt field lives. As a result, saving new agent instructions in OpenRAG settings never updates langflow

## Changes:
- Updated update_chat_flow_system_prompt to use AGENT_COMPONENT_DISPLAY_NAME as the target component
- Added unit tests verifying the value is written to the Agent component in the flow


Vchen7629:
This is a the same changes as PR #1781, just reopened to change the branch from my fork to the main repo to avoid e2e/integration test issues

Wallgau (approved):
LTM, great work!

PR #1929, opened 2026-06-18
fix: Document deletion now properly updates the stale knowledge filter source count

### LINKED ISSUE #1254: [Bug]: Knowledge filter shows 1 source even when filter contains no documents
### OpenRAG Version

0.4.0dev8

### Deployment Method

uv add (installed in project)

### Operating System

macOS Tahoe 26.3

### Python Version

3.13

### Affected Area

Knowledge Filters (partitions, document filtering)

### Bug Description

When all documents in an existing filter are deleted, the filter still indicates the presence of sources.

### Steps to Reproduce

1. Go to knowledge.
2. Create a filter with one source.
3. After saving this filter, delete this source doc from the knowledge base.
4. Check the information related to the filter.

### Expected Behavior

The filter must show 0 sources.

### Actual Behavior

Shows 1 source in the filter information.

### Relevant Logs

```shell

```

### Screenshots

Image

### Additional Context

_No response_

### Checklist

Vchen7629:
## Summary
Fixes #1254 . Knowledge filter containing specific documents used to show stale source counts even though the document was deleted and would show inflated counts if more sources are added even after the page was refreshed. This is because the source count was computed from query_data.filters.data_sources, a saved snapshot of filenames that's never pruned when a document is deleted.

## Demo (Before)
[screen-capture (5).webm](https://github.com/user-attachments/assets/990e1b42-e3b2-4d04-bc5f-3a05b90e0ebe)

## Demo (After)
[screen-capture (3).webm](https://github.com/user-attachments/assets/57e7f76b-8e36-4eff-ae26-ee103784cb6a)

## Changes
- `search_knowledge_filters` now computes `active_source_count` per filter by checking which of its `data_sources` filenames still have indexed documents (one batched OpenSearch query via the admin client, so the count is consistent for every viewer of a shared filter)
- Added `build_existing_filenames_agg_body` helper in `opensearch_queries.py`
- `knowledge-filter-list.tsx` badge now uses `active_source_count`, falling back to the old `.length` if absent
- `knowledge-filter-panel.tsx` source dropdown now filters its selected sources against the currently-active sources before display, so its count matches too
- Added unit tests for `search_knowledge_filters`

## Summary by CodeRabbit

* **New Features**
  * Filters now display an accurate **“active sources”** count from currently indexed documents, with correct pluralization.

* **Bug Fixes**
  * The “sources” selector preserves wildcard selection and now hides any invalid/unknown selections.
  * If a filter’s metadata is malformed, the count is omitted for that filter while other filters still render correctly.

* **Tests**
  * Added unit tests for zero-count behavior, malformed-metadata tolerance, and deduplicating shared filenames to avoid repeated admin lookups.


frontend/components/knowledge-filter-panel.tsx
```diff
@@ -372,7 +372,13 @@ export function KnowledgeFilterPanel() {
             <div className="space-y-2">
               <MultiSelect
                 options={sourceOptions}
-                value={selectedFilters.data_sources}
+                value={
+                  selectedFilters.data_sources[0] === "*"
+                    ? selectedFilters.data_sources
+                    : selectedFilters.data_sources.filter((source) =>
+                        sourceOptions.some((option) => option.value === source),
```
Wallgau: For each selected source, sourceOptions.some() linearly scans all options, so this is O(D × S) (D = selected sources, S = available sources), and it re-runs on every render of the panel (typing in name/description/query, slider drags, etc...). Since sourceOptions comes from the wildcard search and can be large, consider building a Set once for O(1) lookups, dropping this to O(D + S):

const availableSourceValues = new Set(sourceOptions.map((o) => o.value));
value={
  selectedFilters.data_sources[0] === "*"
    ? selectedFilters.data_sources
    : selectedFilters.data_sources.filter((source) =>
        availableSourceValues.has(source),
      )
}

Wallgau (dismissed):
nice work, just quick change to improve performance would be great

frontend/components/knowledge-filter-panel.tsx
Vchen7629: Good point, I will fix this

Wallgau (approved):
Great work!

PR #1933, opened 2026-06-18
feat: improve ingest tracking events

mfortman11:
Improves event tracking details around start and completion of ingest

## Summary by CodeRabbit

* **Chores**
  * Improved ingestion analytics across document upload, shared bucket ingestion, and connector sync by emitting “process started” and failure events with richer context (source/origin, connector type, totals, and embedding model when available).
  * Updated task creation/registration so ingestion tasks are consistently tagged with a `source` (such as file/folder/path/connector), improving the accuracy of completion and error tracking.

### DIFF
--- a/frontend/components/knowledge-dropdown.tsx
+++ b/frontend/components/knowledge-dropdown.tsx
@@ -35,7 +35,7 @@ import { useAuth } from "@/contexts/auth-context";
 import { useIsCloudBrand } from "@/contexts/brand-context";
 import { useTask } from "@/contexts/task-context";
 import { usePermissions } from "@/hooks/use-permissions";
-import { trackStartProcess } from "@/lib/analytics";
+import { trackProcessFailure, trackStartProcess } from "@/lib/analytics";
 import {
   getConnectorDescriptor,
   getConnectorDescriptors,
@@ -346,6 +346,13 @@ export function KnowledgeDropdown() {
       await uploadFileUtil(file, replace);
       refetchTasks();
     } catch (error) {
+      trackProcessFailure({
+        processType: "Ingestion",
+        process: "Document Upload",
+        category: "Knowledge",
+        source: "file",
+        resultValue: error instanceof Error ? error.message : "Unknown error",
+      });
       // Dispatch event that chat context can listen to
       // This avoids circular dependency issues
       if (typeof window !== "undefined") {
@@ -387,8 +394,15 @@ export function KnowledgeDropdown() {
     for (const batch of batches) {
       try {
         const result = await uploadFiles(batch, replace);
-        addTask(result.taskId);
+        addTask(result.taskId, { source: "folder" });
       } catch (error) {
+        trackProcessFailure({
+          processType: "Ingestion",
+          process: "Document Upload",
+          category: "Knowledge",
+          source: "folder",
+          resultValue: error instanceof Error ? error.message : "Unknown error",
+        });
         console.error("[Folder Upload] Batch upload failed:", error);
         toast.error("Batch upload failed", {
           description: error instanceof Error ? error.message : "Unknown error",
@@ -618,7 +632,7 @@ export function KnowledgeDropdown() {
           throw new Error("No task ID received from server");
         }
 
-        addTask(taskId);
+        addTask(taskId, { source: "path" });
         setFolderPath("");
         // Refetch tasks to show the new task
         refetchTasks();
@@ -635,6 +649,13 @@ export function KnowledgeDropdown() {
         }
       }
     } catch (error) {
+      trackProcessFailure({
+        processType: "Ingestion",
+        process: "Document Upload",
+        category: "Knowledge",
+        source: "path",
+        resultValue: error instanceof Error ? error.message : "Unknown error",
+      });
       console.error("Folder upload error:", error);
     } finally {
       setFolderLoading(false);
--- a/frontend/components/knowledge-dropdown.tsx
+++ b/frontend/components/knowledge-dropdown.tsx
@@ -35,6 +35,7 @@ import { useAuth } from "@/contexts/auth-context";
 import { useIsCloudBrand } from "@/contexts/brand-context";
 import { useTask } from "@/contexts/task-context";
 import { usePermissions } from "@/hooks/use-permissions";
+import { trackStartProcess } from "@/lib/analytics";
 import {
   getConnectorDescriptor,
   getConnectorDescriptors,
@@ -333,6 +334,13 @@ export function KnowledgeDropdown() {
 
   const uploadFile = async (file: File, replace: boolean) => {
     setFileUploading(true);
+    trackStartProcess({
+      processType: "Ingestion",
+      process: "Document Upload",
+      category: "Knowledge",
+      source: "file",
+      total_files: 1,
+    });
 
     try {
       await uploadFileUtil(file, replace);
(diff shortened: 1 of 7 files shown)


frontend/lib/analytics.ts
```diff
@@ -70,6 +70,16 @@ export const trackButton = <T = Record<string, unknown>>({
 }: T & ButtonEventParams): void =>
   track("Button Clicked", { action, ...rest } as Record<string, unknown>);
 
+interface StartProcessParams {
```
mpawlow, line 73: ### (1a) [Minor] `source` field is untyped — bypasses compile-time validation

#### Problem

- `StartProcessParams` defines `processType`, `process`, and `category` but omits `source`
- Every call site passes `source` as an extra generic prop (typed as `T = Record<string, unknown>`), which bypasses TypeScript compile-time checking
- Valid values (`"file"`, `"folder"`, `"path"`, `"connector"`) are undocumented at the type level — a future caller could pass any arbitrary string with no warning

#### Code References

- `frontend/app/upload/[provider]/page.tsx` line 100 (`source: "connector"`)
- `frontend/components/connectors/shared-bucket-view.tsx` line 85 (`source: "connector"`)
- `frontend/components/knowledge-dropdown.tsx` lines 341, 373, 600 (`source: "file"`, `"folder"`, `"path"`)

#### Potential Solution

- Add `source` as an optional typed union to `StartProcessParams`:
  ```typescript
  interface StartProcessParams {
    processType: string;
    process?: string;
    category?: string;
    source?: "file" | "folder" | "path" | "connector";
  }
  ```
- This follows the same pattern as `EndProcessParams` and makes valid values discoverable/enforced by the compiler

frontend/components/knowledge-dropdown.tsx
```diff
@@ -333,6 +334,13 @@ export function KnowledgeDropdown() {
 
   const uploadFile = async (file: File, replace: boolean) => {
     setFileUploading(true);
+    trackStartProcess({
```
mpawlow, line 337: ### (1b) [Minor] "Started Process" event fires even when `uploadFile` HTTP request fails — orphaned start event

#### Problem

- In `knowledge-dropdown.tsx:uploadFile`, `trackStartProcess` is called unconditionally before the `try` block
- If `uploadFileUtil` throws (e.g., network failure, 4xx/5xx response), the `catch` block shows an error toast but no task is created server-side
- Without a task, `task-context.tsx` never fires `trackProcessSuccess` or `trackProcessFailure`
- Result: a "Started Process" analytics event has no corresponding "Ended Process" event — these orphaned events inflate apparent funnel drop-off and skew conversion metrics

#### Code References

- `frontend/lib/analytics.ts` lines 79–81 (`trackStartProcess`)
- `frontend/contexts/task-context.tsx` lines 438–463 (completion event paths)

#### Potential Solution

- Move `trackStartProcess` into the `try` block, after a successful response from `uploadFileUtil`:
  ```typescript
  const uploadFile = async (file: File, replace: boolean) => {
    setFileUploading(true);
    try {
      await uploadFileUtil(file, replace);
      trackStartProcess({
        processType: "Ingestion",
        process: "Document Upload",
        category: "Knowledge",
        source: "file",
        total_files: 1,
      });
      refetchTasks();
    } catch (error) { ... }
  ```
- This ensures a "Started Process" event only fires when a task has been accepted by the server

frontend/components/knowledge-dropdown.tsx
```diff
@@ -359,6 +367,14 @@ export function KnowledgeDropdown() {
     filesToUpload: File[],
     replace: boolean,
   ) => {
+    trackStartProcess({
```
mpawlow, line 377: ### (1c) [Normal] Batch folder uploads produce a 1:N start-to-end event ratio

#### Problem

- `uploadFolderBatches` fires a single `trackStartProcess` event once at entry
- It then creates **one task per batch** via `uploadFiles` in a loop (line 389: `addTask(result.taskId)`)
- Each task completion independently fires `trackProcessSuccess` or `trackProcessFailure` via `task-context.tsx`
- For a 100-file folder upload with `uploadBatchSize = 10`, this produces 1 start event and 10 end events — a 1:10 mismatch

#### Background Information

- Funnel analysis tools (Segment included) assume a 1:1 start/end relationship per user action
- A 1:N ratio causes the funnel to show an inflated completion rate (multiple end events per start)
- It also makes aggregate metrics like `total_files` inconsistent: the start event reports `total_files: 100`, but each end event reports `total_files: 10`

#### Code References

- `frontend/contexts/task-context.tsx` lines 421–463 (per-task completion tracking)

#### Potential Solution

- Track a single logical start and single logical end by waiting for all batches to resolve, then emitting one summary end event from `uploadFolderBatches` rather than relying on per-task completion events
- Alternatively, move the start event inside the batch loop so each batch has its own paired start/end:
  ```typescript
  for (const batch of batches) {
    try {
      const result = await uploadFiles(batch, replace);
      trackStartProcess({ ..., total_files: batch.length });
      addTask(result.taskId);
    } catch (error) { ... }
  }
  ```

frontend/contexts/task-context.tsx
```diff
@@ -437,6 +444,8 @@ export function TaskProvider({ children }: { children: React.ReactNode }) {
               total_files: currentTask.total_files,
               failed_files: failedFiles,
               duration_seconds: currentTask.duration_seconds,
+              embedding_model: embeddingModel,
```
mpawlow, line 456: ### (1d) [Minor] `source` dimension missing from completion events — breaks source-level funnel analysis

#### Problem

- All five `trackStartProcess` call sites include a `source` field (`"file"`, `"folder"`, `"path"`, `"connector"`)
- Neither `trackProcessSuccess` nor `trackProcessFailure` in `task-context.tsx` includes a `source` field
- This means it is impossible to segment completion rates by upload source (e.g., "what % of connector ingestions succeed vs file uploads?")

#### Code References

- `frontend/contexts/task-context.tsx` lines 438–463 (`trackProcessFailure` / `trackProcessSuccess` calls)
- `frontend/components/knowledge-dropdown.tsx` lines 337, 370, 596 (start events with `source`)

#### Potential Solution

- Derive or propagate `source` alongside `connector_type` and include it in completion payloads:
  ```typescript
  const source = connectorType === "local"
    ? "file"   // or use a separate ref to store the source per task_id
    : "connector";
  
  trackProcessSuccess({
    ...,
    connector_type: connectorType,
    source,
  });
  ```
- For finer granularity (`"file"` vs `"folder"` vs `"path"`), store the source in a ref alongside `taskConnectorTypesRef`, similar to how connector type is stored

mpawlow (approved):
## Code Review 1

- ✅  Approved / LGTM  🚀 
- See PR comments (1a), (1b), (1d) for **Minor** feedback for potential consideration
- See PR comment: (1c) to verify intended behavior

frontend/lib/analytics.ts
mfortman11: Because this is for tracking only type safety is pretty low priority

frontend/components/knowledge-dropdown.tsx
mfortman11: I think we keep this for now unless there are issues w/ the amplitude dashboards.

frontend/components/knowledge-dropdown.tsx
mfortman11: fixed by adding a failed event in the catch blocks

PR #1935, opened 2026-06-22
fix: Set Model after Provider is deleted

mfortman11:
How to replicate and test

1. Add a second provider and select models from that providers
2. Delete that provider
3. Previously: The model input was empty
4. Now: A default model is selected

## Release Notes

* **New Features**
  * Automatic fallback selection for LLM and embedding models when no model is set, using provider-appropriate defaults.

* **Updates**
  * Default LLM model updated from “gpt-4o-mini” to “gpt-5.4-mini”.

* **Bug Fixes**
  * Improved behavior when removing or switching providers: models now reset to appropriate provider defaults instead of being cleared incorrectly.

* **Tests**
  * Added unit tests covering provider removal and default model selection for both LLMs and embeddings.


mpawlow:
## Functional Review 1

- ✅  LGTM  🚀

frontend/app/settings/_components/agent-settings-section.tsx
```diff
@@ -133,6 +146,18 @@ export function AgentSettingsSection() {
     },
   });
 
+  const allLlmOptions = useMemo(
+    () => groupedLlmModels.flatMap((g) => g.options),
+    [groupedLlmModels],
+  );
+
+  useEffect(() => {
```
mpawlow, line 169: ### (1a) [Minor] Fallback `useEffect` can fire duplicate save mutations during incremental model loading

#### Problem

- The new effects run `handleModelChange` / `handleEmbeddingModelChange` whenever `!llm_model && options.length > 0`.
- `allLlmOptions` / `allEmbeddingOptions` are memoized off the per-provider query results. Those queries resolve at *different times* (OpenAI, Anthropic, Ollama, watsonx), so the memo identity changes several times during initial load.
- Each identity change re-runs the effect. Until the in-flight `updateSettingsMutation` completes and `settings` is refetched, `settings.agent?.llm_model` is still empty, so the guard passes again and a second/third mutation fires.
- Net effect: multiple redundant `PATCH /settings` calls (and multiple "Settings updated successfully" toasts) for a single intended default-selection.

#### Background Information

- CodeRabbit flagged this; its stated premise ("`groupedLlmModels` isn't memoized") is now outdated since the PR memoizes both lists — but the residual race remains because the memo inputs themselves change as queries resolve.

#### Code References

- `frontend/app/settings/_components/ingest-settings-section.tsx:127-139` (approx.)

#### Potential Solution

- Add a one-shot `useRef` guard so the auto-select fires at most once per empty-model state:

```ts
const autoSelectedLlm = useRef(false);
useEffect(() => {
  if (autoSelectedLlm.current) return;
  if (!settings.agent?.llm_model && allLlmOptions.length > 0) {
    autoSelectedLlm.current = true;
    const fallback = allLlmOptions.find((o) => o.default) || allLlmOptions[0];
    handleModelChange(fallback.value, fallback.provider);
  }
}, [settings.agent?.llm_model, allLlmOptions]);
```

#### Alternative Solutions

- Gate the effect on `!isLoadingAnyLlmModels` so it only runs once all provider queries have settled (single, stable `allLlmOptions`).
- Track `updateSettingsMutation.isPending` and skip while a mutation is in flight.

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
- See PR comment (1a) for potential / optional consideration

frontend/app/settings/_components/agent-settings-section.tsx
mfortman11: Good catch fixed

PR #1946, opened 2026-06-23
fix: adds safety guard to ag grid

mfortman11:
**Problem**: Calling `api.forEachNode()` / `api.getSelectedRows()` before AG Grid's row model is fully initialized causes a `RowApiModule not registered` error. This is a timing-dependent race condition on mount — more likely on slower machines or with React strict mode.

**Fix**: Added a `gridReadyRef` flag set by AG Grid's `onGridReady` callback, and a `getGridApi()` helper that returns null until the grid is ready. All 5 grid API call sites now go through this helper, so they safely no-op if called before initialization.

## Bug Fixes
* Improved reliability of selection and bulk-delete operations on the Knowledge page by guarding grid actions until the grid is fully initialized (and safely handling teardown).
* Enhanced the chat “waiting too long” timeout behavior by correcting the timer trigger logic to respond reliably to streaming state.
* Improved AG Grid runtime support by registering the required Row API module so related grid functionality is available at runtime.

### DIFF
--- a/frontend/app/knowledge/page.tsx
+++ b/frontend/app/knowledge/page.tsx
@@ -453,6 +453,12 @@ function SearchPage() {
 
   const gridRows = fileResults;
   const gridRef = useRef<AgGridReact>(null);
+  const gridReadyRef = useRef(false);
+
+  const getGridApi = useCallback(() => {
+    if (!gridReadyRef.current) return null;
+    return gridRef.current?.api ?? null;
+  }, []);
 
   // Re-run only when row identity/status changes, not on every list poll reference.
   const gridRowsSelectionKey = useMemo(
@@ -466,7 +472,7 @@ function SearchPage() {
   );
 
   useEffect(() => {
-    const api = gridRef.current?.api;
+    const api = getGridApi();
     if (!api) {
       return;
     }
@@ -475,7 +481,7 @@ function SearchPage() {
     setSelectedRows((current) =>
       sameFileSelection(current, nextSelected) ? current : nextSelected,
     );
-  }, [gridRowsSelectionKey, isDeletableKnowledgeRow]);
+  }, [gridRowsSelectionKey, isDeletableKnowledgeRow, getGridApi]);
 
   const columnDefs: ColDef<File>[] = [
     {
@@ -745,17 +751,18 @@ function SearchPage() {
   };
 
   const onSelectionChanged = useCallback(() => {
-    if (!gridRef.current) {
+    const api = getGridApi();
+    if (!api) {
       return;
     }
     const nextSelected = syncGridSelectionToDeletableRows(
-      gridRef.current.api,
+      api,
       isDeletableKnowledgeRow,
     );
     setSelectedRows((current) =>
       sameFileSelection(current, nextSelected) ? current : nextSelected,
     );
-  }, [isDeletableKnowledgeRow]);
+  }, [isDeletableKnowledgeRow, getGridApi]);
 
   const handleBulkDelete = async () => {
     const rowsToDelete = selectedRows.filter(isDeletableKnowledgeRow);
@@ -826,9 +833,7 @@ function SearchPage() {
       setShowBulkDeleteDialog(false);
 
       // Clear selection in the grid
-      if (gridRef.current) {
-        gridRef.current.api.deselectAll();
-      }
+      getGridApi()?.deselectAll();
     } catch (error) {
       toast.error(
         error instanceof Error
@@ -886,7 +891,7 @@ function SearchPage() {
                 onDelete={() => setShowBulkDeleteDialog(true)}
                 onCancel={() => {
                   setSelectedRows([]);
-                  gridRef.current?.api.deselectAll();
+                  getGridApi()?.deselectAll();
                 }}
               />
             </div>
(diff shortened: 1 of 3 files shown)


mpawlow:
## Functional Review 1

@mfortman11 

- ❌ Problem is still occurring for me
- Not sure if I need to do something to explicitly reset UI state

```
AG Grid: error #200 "Unable to use api.forEachNode as RowApiModule is not registered.  Check if you have registered the module:\n\nimport { ModuleRegistry } from 'ag-grid-community'; \nimport { RowApiModule } from 'ag-grid-community'; \n\nModuleRegistry.registerModules([ RowApiModule ]); \n\nFor more info see: https://www.ag-grid.com/react-data-grid/modules/" "\nSee https://www.ag-grid.com/react-data-grid/errors/200?_version_=34.2.0&gridId=1&gridScoped=false&rowModelType=clientSide&isUmd=false&moduleName=RowApi&reasonOrId=api.forEachNode"
```

image

frontend/app/knowledge/page.tsx
```diff
@@ -1011,6 +1016,9 @@ function SearchPage() {
             }
             isRowSelectable={(params) => isDeletableKnowledgeRow(params.data)}
             domLayout="normal"
+            onGridReady={() => {
```
mpawlow: ### (1a) [Minor] Duplicate `onGridReady` inline handlers across both grid instances

#### Problem

- Both the cloud-brand and non-cloud `AgGridReact` instances define **identical** inline `onGridReady` arrow functions
- Inline handlers in JSX are recreated on every render; for `onGridReady` this is low-cost (AG Grid calls it once, not on re-renders), but the duplication still violates the DRY principle and creates two maintenance points if the handler ever needs to grow.

#### Code References

- `frontend/app/knowledge/page.tsx` line 1019–1021 (cloud instance)
- `frontend/app/knowledge/page.tsx` line 1056–1058 (non-cloud instance)

#### Potential Solution

- Extract a single named `useCallback` above the JSX return and use it on both instances:
  ```tsx
  const handleGridReady = useCallback(() => {
    gridReadyRef.current = true;
  }, []);
  ```
  Then in both `<AgGridReact>` elements:
  ```tsx
  onGridReady={handleGridReady}
  ```

frontend/app/knowledge/page.tsx
```diff
@@ -453,6 +453,12 @@ function SearchPage() {
 
   const gridRows = fileResults;
   const gridRef = useRef<AgGridReact>(null);
+  const gridReadyRef = useRef(false);
```
mpawlow, line 456: ### (1b) [Minor] `gridReadyRef` is never reset when the grid re-initializes

#### Problem

- There is no `onGridPreDestroyed` (or equivalent teardown) callback to reset `gridReadyRef.current = false`.
- If AG Grid ever destroys and re-creates within the same component lifecycle (e.g., a key-prop change, or a future refactor that conditionally mounts/unmounts the grid while the parent component stays alive), `gridReadyRef.current` will remain `true` during the reinitialization gap.

#### Background Information

- In practice this is **functionally safe** today because `getGridApi()` has a second layer of defense — `gridRef.current?.api ?? null` — that returns `null` whenever the AG Grid ref doesn't have an active API yet (including mid-reinit):
  ```ts
  const getGridApi = useCallback(() => {
    if (!gridReadyRef.current) return null;   // fast-path early exit
    return gridRef.current?.api ?? null;       // safety net even if ref is stale
  }, []);
  ```
  So no incorrect behavior exists in the current code. However, the `gridReadyRef` check is semantically misleading when it is `true` but the grid isn't actually ready — it makes the guard's invariant harder to reason about for future maintainers.
- Since `isCloudBrand` is a static brand configuration (not toggled at runtime), the scenario of switching between the two grids at runtime is not currently triggered.

#### Code References

- `frontend/app/knowledge/page.tsx` lines 1019–1021, 1056–1058 (`onGridReady` handlers)

#### Potential Solution

- Add `onGridPreDestroyed` alongside `onGridReady` on both grid instances to reset the flag:
  ```tsx
  onGridPreDestroyed={() => {
    gridReadyRef.current = false;
  }}
  ```
  This makes the guard's invariant accurate throughout the grid's full lifecycle.

mpawlow (changes requested):
## Code Review 1

- ✅  Approved / LGTM  🚀 
- See PR comments: (1a), (1b) for optional consideration (both MINOR nits)
- See **Functional Review** for requested changes (might be a user error on my part)

mpawlow (approved):
## Functional Review 1

- ✅  Approved / LGTM  🚀

mpawlow:
### Orthogonal Error?

- I believe this is unrelated, but encountered this new error when trying to chat after ingesting a document

```
Console Error

Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

image

image

mfortman11:
> ### Orthogonal Error?
> * I believe this is unrelated, but encountered this new error when trying to chat after ingesting a document
> 
> ```
> Console Error
> 
> Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
> ```
> 
> image image

Added a small fix for this

```
By coercing streamingMessage to a boolean (!!streamingMessage), the effect only re-runs when it transitions between null and non-null — not on every streaming chunk. Previously it was re-firing hundreds of times during a single response because each chunk created a new object reference.
```

frontend/app/knowledge/page.tsx
mfortman11: Addressed

frontend/app/knowledge/page.tsx
mfortman11: Addressed

PR #1954, opened 2026-06-23
feat: shared documents ingestion mode for IBM COS connectors

ricofurtado:
This pull request introduces a new "shared" ingestion mode for IBM COS connectors, allowing documents to be indexed without an owner so that all users in an OpenRAG instance can access them. The change is implemented across the frontend, backend, and SDKs, with UI updates to expose the option where appropriate, backend validation and propagation of the shared flag, and adjustments to document processing logic to omit owner fields when shared mode is enabled.

## Summary by CodeRabbit

* **New Features**
  * Added support for indexing and syncing shared documents in IBM COS, including an option to make documents available to all users.
  * Updated delete actions to recognize anonymous/shared document permissions.

* **Bug Fixes**
  * Improved document deletion behavior for ownerless documents and mixed owned/shared content.
  * Fixed header handling for non-ASCII values to avoid upload and ingestion failures.
  * Made connector and ingestion flows more consistent when working with shared content.


src/connectors/service.py
```diff
@@ -348,6 +348,7 @@ async def sync_connector_files(
         filename_filter: set = None,
         ingest_settings: dict[str, Any] | None = None,
         replace_duplicates: bool = False,
+        shared: bool = False,
```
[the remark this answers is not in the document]
ricofurtado: fixed

src/api/connectors.py
phact, line 893: I think we want to pass `shared` in this path too?

src/models/processors.py
phact, line 268: What happens here when we have shared ownerless chunks?

src/api/connectors.py
phact, line 650: Hmm does this mean anyone can replace / delete ownerless documents, even without the new https://github.com/langflow-ai/openrag/pull/1954/changes#diff-d782366fa4dfe87f7b831db0d11c0b4fee7e880d25176998a41c9243a829bffdR24 permission? cc: @edwinjosechittilappilly

src/api/v1/documents.py
phact, line 212: Should also set the status_code for the jsonresponse not just the success object.

src/api/connectors.py
ricofurtado: Yes, good catch! Both calls to sync_specific_files are in the resync of  already indexed files, the else path that runs when no files are explicitly selected and no full ingest was requested. It resyncs whatever files are already in Os.

ricofurtado: I can see it... but the gap is really narrow: 
explicit file selection with `shared=True + replace_duplicates=True`. Whether it needs a guard depends on whether we want `knowledge:delete:anonymous` to be the single gate for all deletions of ownerless documents regardless of code path.

When it's reachable:

1- explicit file selection  `replace_duplicates=body.replace_duplicates ` caller controls it, so `shared=True + replace_duplicates=True` in the request body deletes ownerless docs with no `delete:anoymous` check.

2 - re-sync paths ( replace_duplicates=_connector_sync_should_replace(connector_type), which returns False for ibm_cos so these paths are not exposed.

sync_all / bucket_filter (lns 798, 788): replace_duplicates not passed, defaults to False in the service, also not exposed.

src/models/processors.py
ricofurtado: bug... deleted or replaced shared files leave orphaned ownerless chunks permanently visible to all users. will fix it. I'm almost sure I had fixed it sometime ago LOL

src/api/connectors.py
ricofurtado: Fixed!

src/models/processors.py
ricofurtado: Fixed.

src/api/connectors.py
ricofurtado: Fixed.

src/api/v1/documents.py
ricofurtado: Yep, it is omiting status code
Now it picks the most severe error code when results are mixed.

phact (approved):

PR #1958, opened 2026-06-24
fix: forward allowed_principals through connector Langflow ingestion

edwinjosechittilappilly:
Cherry-pick of #1900 onto `main`.

Documents shared via connector sync were invisible to connector-alias users in DLS-enabled deployments. The `ConnectorFileProcessor` extracted `allowed_users` and `allowed_groups` from the document ACL but silently dropped `allowed_principals` and `allowed_principal_labels` — so the Langflow ingestion call never received them, and indexed documents ended up with no principals, making them unreachable via DLS access checks.

**Fix:** Extract `allowed_principals` / `allowed_principal_labels` from `document.acl` and pass them through to the `process_item` call, matching the existing pattern for `allowed_users` / `allowed_groups`.

## Changes

- `src/models/processors.py` — extract and forward `allowed_principals` and `allowed_principal_labels` inside `ConnectorFileProcessor`

## Summary by CodeRabbit

* **New Features**
  * Expanded file ingestion permissions to carry principal-level access details, improving permission fidelity during uploads.
  * Files now preserve additional ACL information when available, alongside existing user and group permissions.

### DIFF
--- a/src/models/processors.py
+++ b/src/models/processors.py
@@ -951,10 +951,14 @@ class ConnectorFileProcessor(TaskProcessor):
                     # Extract ACL information
                     allowed_users: list[str] = []
                     allowed_groups: list[str] = []
+                    allowed_principals: list[str] = []
+                    allowed_principal_labels: list[dict[str, Any]] = []
                     if document.acl:
                         try:
                             allowed_users = document.acl.allowed_users or []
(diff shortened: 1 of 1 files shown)


mpawlow (approved):
## Code Review 1

- ✅  Approved / LGTM  🚀

PR #1964, opened 2026-06-24
feat: configurable OpenSearch shards/replicas + startup replica reconciliation

edwinjosechittilappilly:
## Summary

Cherry-pick of #1951 onto `main`.

All OpenRAG indices were hard-coded to 1 shard / 0 replicas, which is fine for single-node dev but wrong for production multi-node clusters. This change makes shard and replica counts configurable via env vars and reconciles existing indices at startup so they match the configured value without requiring a full reindex.

**Changes:**
- `src/config/settings.py` — new env vars: `OPENRAG_OPENSEARCH_NUMBER_OF_SHARDS` (default `2`), `OPENRAG_OPENSEARCH_NUMBER_OF_REPLICAS` (default `2`), `OPENRAG_ENSURE_INDEX_REPLICAS_ON_STARTUP` (default `true`), `OPENSEARCH_WAIT_MAX_RETRIES` (default `100`)
- `src/utils/opensearch_init.py` — new `_ensure_index_replicas` helper and `ensure_openrag_index_replicas` that reconciles all known OpenRAG indices at startup; replaces duplicated inline replica-update blocks across `init_index`
- `src/utils/embeddings.py` — `create_index_body` uses configured shard/replica counts instead of hard-coded values
- `src/app/lifespan.py` — startup block runs replica reconciliation (best-effort) alongside optional security bootstrap using the same authenticated client; missing service token is fatal only for the security bootstrap path
- `docker-compose.yml`, `.env.example` — defaults to 1 shard / 0 replicas / reconciliation off for single-node local dev
- `docs/docs/reference/configuration.mdx` — documents the two new shard/replica env vars
- `tests/unit/test_ensure_index_replicas.py` — unit tests: corrects mismatched replicas, noop when already matching, skips missing indices
- `tests/unit/test_embedding_fields.py` — adds test that `create_index_body` honours configured shard/replica values

## Summary by CodeRabbit

* **New Features**
  * Added configurable index shard and replica settings for OpenSearch, including optional startup reconciliation of replica counts.
  * Existing OpenSearch indices can now be updated at startup to match the configured replica count.

* **Bug Fixes**
  * Improved startup handling so OpenSearch maintenance tasks can run safely alongside security setup.
  * Added support for more consistent index creation across new and existing indexes.

* **Documentation**
  * Updated configuration docs and example environment values for the new OpenSearch settings.


mpawlow (approved):
## Code Review 1

- ✅  Approved / LGTM  🚀

PR #1973, opened 2026-06-25
fix: block image ingestion when OCR is disabled

NishadA05:
## Summary

Adds a check to stop image files from being ingested when OCR is disabled.

Images like `png`, `jpg`, `jpeg`, `bmp`, `tiff`, `tif`, and `webp` need OCR to extract text. If OCR is turned off, these files now fail early with a clear error message:

`The file '<filename>' is an image file and cannot be ingested because OCR is disabled.`

This PR adds:

* an `IMAGE_EXTENSIONS` list
* an `is_image_file()` helper
* early checks in local file uploads and connector file uploads

For local uploads, the check runs before the file is hashed or processed.

For connector uploads, the check runs when the filename is first available, and again after the connector file is loaded in case the final filename changes.

When the check fails, the task is marked as `FAILED`, the error is saved, and processing stops.

### DIFF
--- a/src/models/processors.py
+++ b/src/models/processors.py
@@ -27,6 +27,15 @@ from .tasks import FileTask, TaskStatus, UploadTask
 
 logger = get_logger(__name__)
 
+# Image file extensions that require OCR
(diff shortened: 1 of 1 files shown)


Wallgau (approved):
Great job! Thank you Nishad!

PR #1978, opened 2026-06-26
fix: require explicit opt-in for anonymous mode

failsafesecurity:
## Summary
- require explicit `OPENRAG_ALLOW_NO_AUTH=true` before anonymous mode activates
- document that leaving Google OAuth credentials blank no longer silently enables unauthenticated access
- add regression coverage for the no-auth opt-in gate and missing-cookie auth behavior

## Security context
OpenRAG currently treats missing Google OAuth credentials as an implicit no-auth mode when IBM auth is disabled. In default OSS configuration, that can turn unauthenticated requests into authenticated `AnonymousUser` requests and expose protected application surfaces. This patch makes anonymous mode an explicit local-development opt-in instead of a silent fallback.

## Validation
- `python3 -m py_compile src/config/settings.py tests/unit/config/test_no_auth_mode.py tests/unit/dependencies/test_auth_jwt_attachment.py`
- full pytest was not available in the audit workspace because repo dependencies were not installed there

## Notes
- no public submission has been created yet; this branch is prepared for maintainer review only

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened anonymous access behavior so blank Google OAuth credentials no longer automatically allow no-auth mode.
  * Added a safer local-development option to explicitly enable anonymous access when needed.
  * Improved login error handling so requests without authentication now return a clear 401 response.

* **Tests**
  * Added coverage for the updated no-auth and unauthenticated request behavior.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -136,6 +136,9 @@ OPENRAG_ENSURE_INDEX_REPLICAS_ON_STARTUP=false
 # instead of Google OAuth. The raw IBM JWT is also passed directly to OpenSearch.
 # When enabled, GOOGLE_OAUTH_CLIENT_ID/SECRET are not required for login.
 IBM_AUTH_ENABLED=false
+# Local-development escape hatch only. When false (default), leaving Google OAuth
+# credentials blank does NOT enable anonymous access.
+OPENRAG_ALLOW_NO_AUTH=false
 # URL to fetch IBM's JWT public key (used to validate ibm-openrag-session tokens).
 IBM_JWT_PUBLIC_KEY_URL=
 # Cookie name set by Traefik after successful IBM AMS authentication (default: ibm-openrag-session).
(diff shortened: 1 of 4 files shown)


failsafesecurity:
👋 Gentle ping — this PR has been open for ~15 days. The fix requires explicit opt-in for anonymous mode to prevent silent authenticated access. Ready to address any feedback.

failsafesecurity:
Friendly follow-up on this security PR. It has been open for a while without human reviewer feedback. Happy to rebase, trim the patch, or adjust the approach if that helps. Thanks for taking a look.

failsafesecurity:
Friendly follow-up on this security fix PR. Happy to rebase, split the change, or adjust the approach if there are compatibility or scope concerns. If you'd prefer to continue privately via security@ or GitHub private reporting, I can resubmit there.

PR #1982, opened 2026-06-29
fix: universal file names

edwinjosechittilappilly:
fix: universal file names

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of non-ASCII values in request headers to prevent ingestion/request errors.
  * Stopped sending the filename global variable for both file- and URL-based ingestion runs.
  * Updated the OpenRAG ingestion pipeline layout and wiring, including revised routing for document table processing and multimodal embedding model selection.
* **Tests**
  * Added unit coverage for ASCII-safe header encoding and safe header construction.
  * Updated ingestion callback tests to assert expected OpenSearch input configuration.
  * Tightened onboarding sample-doc ingestion gating to require an Anthropic API key.
* **Chores**
  * Renamed a CI toggle key for the OpenSearch node count check.


edwinjosechittilappilly:
Screenshot 2026-06-29 at 12 32 16 PM

src/services/langflow_file_service.py
```diff
@@ -426,7 +426,7 @@ async def run_ingestion_flow(
             "X-Langflow-Global-Var-OWNER_NAME": str(owner_name),
             "X-Langflow-Global-Var-OWNER_EMAIL": str(owner_email),
             "X-Langflow-Global-Var-CONNECTOR_TYPE": str(connector_type),
-            "X-Langflow-Global-Var-FILENAME": filename,
+            # "X-Langflow-Global-Var-FILENAME": filename,
```
mfortman11: Should we cleanup as opposed to commented out?

src/services/langflow_file_service.py
```diff
@@ -607,7 +607,7 @@ async def run_url_ingestion_flow(
             "X-Langflow-Global-Var-ALLOWED_GROUPS": json.dumps([]),
             "X-Langflow-Global-Var-ALLOWED_PRINCIPALS": json.dumps([]),
             "X-Langflow-Global-Var-DOCLING_TASK_ID": "",
-            "X-Langflow-Global-Var-FILENAME": str(docs_url),
+            # "X-Langflow-Global-Var-FILENAME": str(docs_url),
```
mfortman11: Same here

flows/ingestion_flow.json
mfortman11, line 1: Do we need flow changes?

mfortman11 (approved):
LGTM works well just a couple checks/nits

PR #2009, opened 2026-07-02
fix: add warning logs for 10k aggregation limit

yogichipalkatti:
Temporarily fixes Issue #1548 

Affected functions:

get_synced_file_ids_for_connector(): Warns when connector_file_id,
get_synced_id_to_filename_map(): Warns when document_id aggregation
reaches the limit

unique values per field

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connector sync discovery logging when results hit the maximum aggregation bucket limit, which can cause incomplete file-id and filename mappings.
  * Warnings now include the connector type and the returned bucket count to make it clearer when and where truncation occurs (including connector file IDs, document IDs, and filename mapping results).

### DIFF
--- a/src/api/connectors.py
+++ b/src/api/connectors.py
@@ -29,6 +29,12 @@ from utils.telemetry import Category, MessageId, TelemetryClient
 
 logger = get_logger(__name__)
 
+# Maximum number of buckets returned by a single OpenSearch terms aggregation.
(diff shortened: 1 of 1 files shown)


lucaseduoli (changes requested):
Could you please extract that 10k limit in a constant?

PR #2010, opened 2026-07-02
fix: apply new settings on retry

mfortman11:
https://github.ibm.com/lakehouse/tracker/issues/76645

Do an upload which fails (like jpg) which asks for you to turn on OCR

Before if you applied the setting change then retried the new settings weren't picked up

## Summary by CodeRabbit

* **Bug Fixes**
  * OCR and image description settings are now refreshed at processing time, so ingestion uses the latest configuration.
  * File processing now applies the current OCR and picture-description options consistently across upload and ingestion flows.
  * Improved consistency between standard and Langflow-based processing paths when handling ingestion settings.


Wallgau (changes requested):
SUPPORTED_FILE_TYPES in knowledge-dropdown.tsx controls which files users can pick for local upload (Add Knowledge, folders, chat attach, onboarding). If an image type isn’t on that list, users can’t upload it from their machine, even if the backend supports images when OCR is on.
So if this PR is meant to let users upload images locally, we’ll need to update that list too (probably only when OCR is enabled in settings). Otherwise users will still be blocked in the file picker.

Wallgau (approved):

PR #2026, opened 2026-07-06
fix: improve unstable context provider

mfortman11:
**The isuue**

`<BrandContext.Provider value={{ brand, setBrand }}>`

`{ brand, setBrand }` creates a new object every render. Even if `brand` and `setBrand` haven't changed, `{} !== {}` — React uses reference equality to compare context values. A new reference means React treats the value as changed and re-renders every consumer.

`BrandProvider` sits near the root of the component tree, so it re-renders whenever anything above it does (layout changes, auth state updates, route transitions, etc.). Each of those re-renders created a new value object, which triggered re-renders in all 38 consuming files — and their entire subtrees.

**The fix**

Memoize `setBrand` with `useCallback` and the value object with `useMemo`.

The `useMemo` fix makes React reuse the same object reference until `brand` or `setBrand` actually changes. Same reference = no context change = consumers skip re-rendering.

## Summary by CodeRabbit

* **Refactor**
  * Improved brand selection handling by stabilizing internal callback and context value references.
  * Kept existing behavior the same for brand updates, including saved preferences and current access restrictions.


Wallgau (approved):
useContext(X) to replace with use(X) will be addressed in another PR for depreciated apis

PR #2051, opened 2026-07-08
feat: add preview-mode ingest backend with live index proof (OSS + SaaS)

Wallgau:
Add preview-mode ingest backend with live index proof (OSS + SaaS)
Summary
Introduces the backend plumbing for preview-mode ingest so the frontend can show live ingest progress and prove that chunks actually landed in OpenSearch without persisting Docling layout JSON on the server. This is the backend half of the visual parser feature; the layout skeleton remains frontend-only, and the backend only serves task phase + index proof.

Preview mode is opt-in per request and gated by run mode: it's enabled in OSS and SaaS, and disabled on_prem.

What's included
New opt-in preview flag on both ingest entrypoints:
POST upload router (src/api/router.py) accepts preview form field.
Connector sync (ConnectorSyncBody.preview) for user-initiated ingests (explicit file selection, not automated re-sync).
The flag is honored only when is_ingest_preview_enabled() returns true; otherwise it's silently ignored and logged at debug level.
preview_mode threaded end-to-end through UploadTask, FileTask, TaskService (all create_*_task paths), connector ConnectorService, and both LangflowFileProcessor / ConnectorFileProcessor.
New index-proof endpoint — GET /ingest/preview/{task_id}/index-proof (src/api/ingest_preview.py), registered in internal.py. Queries OpenSearch live for indexed chunk metadata (phase, per-chunk text previews, page, char counts, embedding model/dimensions). Supports a file query param to select a specific file within a multi-file preview task. Returns 404 when preview is unavailable, the task is missing, or the task is not a preview ingest.
IngestPreviewService (src/services/ingest_preview_service.py) — a stateless helper (no in-memory Docling cache) wired into the DI container and dependencies.py. Handles not-ready / OpenSearch-unavailable / search-failure cases gracefully.
document_id tracking on FileTask — set in both the Langflow processor (from file hash) and the connector processor (from the document) before ingest, so the index-proof lookup can resolve chunks.
Task serialization change — preview tasks keep completed files in /tasks/enhanced (normally only in-flight/failed files are surfaced) so the multi-file preview carousel can enumerate all files.
is_ingest_preview_enabled() helper added to run_mode_utils.py.
Minor hardening in docling_service.py: guard against document being None in the docling-serve response.
Intentionally excluded
The spike's /docling endpoint and preview-only Docling options (embedded page images / persisted layout JSON) are not included. The backend serves task phase + index proof only.

Tests:
tests/unit/api/test_ingest_preview.py — endpoint guards (run mode, missing task, non-preview task).
tests/unit/services/test_ingest_preview_service.py — index-proof service (ready/not-ready, error paths, chunk shaping).
tests/unit/test_router_preview_upload.py — router preview flag threading.
tests/unit/test_task_service.py — enhanced task serialization for preview vs normal tasks.
tests/unit/config/test_disable_ingest_setting.py — updated for the new path.
Notes: 
Preview is off by default and requires the caller to send preview=true; only takes effect in OSS/SaaS.
The index-proof query is bounded (size: 200) and only pulls a small _source allowlist, so it's a cheap read.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in ingest preview for upload and connector sync requests (feature-flagged), including propagation of preview mode into ingest tasks.
  * Introduced an ingest “index proof” endpoint that returns chunk details and readiness for a preview upload.
  * Preview-mode tasks now retain completed files in task listings.
* **Bug Fixes**
  * Improved preview gating and error handling when preview is unavailable or when the requested task/file can’t be found.
  * Updated preview identifiers so the preview index proof uses the correct document id.


src/services/ingest_preview_service.py
```diff
@@ -0,0 +1,185 @@
...
+                "phase": phase,
+                "chunk_count": 0,
+                "chunks": [],
+                "chunks_returned": 0,
+                "chunks_truncated": False,
+                "document_id": document_id,
+                "error": "search_failed",
+            }
+
+        hits_section = response.get("hits", {})
```
mpawlow, line 148: ### (1a) [Minor] Missing test coverage for real production failure paths in `IngestPreviewService`

#### Problem

- `IngestPreviewService.get_index_proof` has two error branches — `opensearch_client is None` → `"opensearch_unavailable"`, and `opensearch_client.search()` raising → `"search_failed"` — that are exercised by neither `tests/unit/services/test_ingest_preview_service.py` nor any other test in the diff. 

#### Background Information

- These are exactly the failure modes most likely to occur in a live deployment (OpenSearch down, network blip, mapping error), and the endpoint's behavior on those paths (404 vs partial payload vs 5xx) is currently unverified by CI.

#### Code References

- `tests/unit/services/test_ingest_preview_service.py` (whole file); logic under test: `src/services/ingest_preview_service.py:100-146`.

#### Potential Solution

- Add the two tests CodeRabbit already proposed (mock `opensearch_client=None` and `opensearch_client.search.side_effect = RuntimeError(...)`), asserting `ready is False` and the corresponding `error` key.

src/services/docling_service.py
mpawlow: ### (1b) [Normal] `_poll_result` still lacks the null-`document` guard applied to its sibling method

#### Problem

- `DoclingService.fetch_task_result` (`src/services/docling_service.py:383-385`) was hardened in this PR to handle `payload.get("document")` being explicitly `None` (`document = payload.get("document") or {}`). The analogous polling method, `_poll_result` (`src/services/docling_service.py:421`), still does `result_json.get("document", {}).get("json_content")`, which raises an unhandled `AttributeError` — not the intended `DoclingServeError` — if `result_json["document"]` is explicitly `None`.

#### Background Information

- This is CodeRabbit's original nitpick from the first review pass

#### Potential Solution

```diff
-                doc_content = result_json.get("document", {}).get("json_content")
+                document = result_json.get("document") or {}
+                doc_content = document.get("json_content")
```

- This mirrors the fix already applied a few lines away in `fetch_task_result`.

src/models/processors.py
```diff
@@ -817,6 +817,7 @@ def __init__(
         ingest_settings: dict[str, Any] | None = None,
         replace_duplicates: bool = False,
         connector_type: str | None = None,
+        preview_mode: bool = False,
```
mpawlow: ### (1c) [Minor] `preview_mode` is stored on `ConnectorFileProcessor`/`LangflowFileProcessor` but never read

#### Problem

- Both processor classes gained a `preview_mode: bool = False` constructor parameter and `self.preview_mode = preview_mode` assignment (`src/models/processors.py:838`, `:1351`), but the attribute is never read anywhere in the codebase (`grep -rn "\.preview_mode\b" src/` outside the assignment sites and `UploadTask.preview_mode` returns nothing). All actual preview behavior is driven by `UploadTask.preview_mode` and `FileTask.document_id`, not by this per-processor flag.

#### Code References

- `src/models/processors.py:820`, `:838`, `:1336`, `:1351`

#### Potential Solution

- Either remove the unused attribute (thread `preview_mode` only where it's actually consumed — `TaskService`/`UploadTask`), or, if a future processor-level behavior change is genuinely planned, note that explicitly in a comment so it doesn't read as leftover/incomplete wiring.

src/api/connectors.py
```diff
@@ -893,6 +896,10 @@ async def connector_sync(
             selected_files = [f.get("id") for f in selected_files_raw if f.get("id")]
             file_infos = selected_files_raw
 
+    # Preview mode is opt-in from the connector upload UI (OSS/SaaS). It applies
```
mpawlow, line 931: ### (1d) [Minor] No unit test coverage for the connector-sync preview flag

#### Problem

- `ConnectorSyncBody.preview` and its threading into `preview_mode = body.preview and is_ingest_preview_enabled()` (`src/api/connectors.py:896-899`) has zero direct test coverage. 
- The upload-router equivalent (`src/api/router.py`) got two dedicated tests (`tests/unit/test_router_preview_upload.py`); the connector path — which additionally interacts with duplicate detection and `shared`/ACL flags before reaching the preview gate — did not.

### Code References

- `src/api/connectors.py:896-899`, `:1027` (`preview_mode=preview_mode` passed to `sync_specific_files`)

#### Potential Solution

- Add a test mirroring `test_router_preview_upload.py`'s pattern: call `connector_sync` with `body.preview=True` and assert `preview_mode=True` reaches `connector_service.sync_specific_files`.

src/services/ingest_preview_service.py
```diff
@@ -0,0 +1,185 @@
...
+    return sorted(
+        hits,
+        key=lambda hit: (
+            (hit.get("_source") or {}).get("page") or 0,
+            _chunk_sequence(hit.get("_id")),
+        ),
+    )
+
+
+def _extract_hit_total(hits_section: dict[str, Any], fallback: int) -> int:
```
mpawlow, line 38: ### (1e) [Minor] `_extract_hit_total` raises `TypeError` if OpenSearch returns `total.value: null`

#### Problem

```python
def _extract_hit_total(hits_section: dict[str, Any], fallback: int) -> int:
    total = hits_section.get("total")
    if isinstance(total, dict):
        return int(total.get("value", fallback))
```

- If `total` is `{"value": None}` (key present, value `None`), `.get("value", fallback)` returns `None` — the default is only used when the key is *missing* — and `int(None)` raises `TypeError`. 

#### Background Information

- This is CodeRabbit's nitpick from the second review pass

#### Potential Solution

```diff
-        return int(total.get("value", fallback))
+        value = total.get("value")
+        return int(value) if value is not None else fallback
```

src/utils/run_mode_utils.py
```diff
@@ -24,6 +25,11 @@ def is_run_mode_on_prem() -> bool:
     return get_run_mode() == RUN_MODE_ON_PREM
 
 
+def is_ingest_preview_enabled() -> bool:
```
mpawlow: ### (1g) [Normal] `is_ingest_preview_enabled` semantically doesn't belong in this utility

#### Problem

- The `src/utils/run_mode_utils.py` is specifically scoped to providing convenient helper functions for determining the run mode of OpenRAG
- The Ingestion Preview feature flag doesn't really fit here

#### Potential Solution

- Recommend moving this feature flag check to it's own stand-alone utility with a dedicated scope to the Ingestion Preview feature

mpawlow (approved):
## Code Review 1

- ✅  Approved / LGTM  🚀 
- See non-blocking PR comments for consideration
   - (1b) (1g)
- See Minor PR comments for optional consideration
  - (1a) (1e) (1c) (1d)

edwinjosechittilappilly:
@coderabbitai review the PR and suggest to put the feature in a feature flag disabled by default so it could be safely merged untill the feature is ready.

src/services/docling_service.py
```diff
@@ -380,10 +380,10 @@ async def fetch_task_result(
         if payload.get("status") == "failure" or payload.get("errors"):
             raise DoclingServeError(f"Docling processing failed: {_format_docling_error(payload)}")
 
-        doc_content = payload.get("document", {}).get("json_content")
-        if doc_content is None:
+        document = payload.get("document") or {}
+        if document.get("json_content") is None:
             raise DoclingServeError("docling-serve response missing document.json_content")
-        return doc_content
+        return document["json_content"]
```
edwinjosechittilappilly, line 386: Do we need to change the .get()

ricofurtado (approved):

src/services/docling_service.py
Wallgau: Yes, payload.get("document", {}) only substitutes {} when the key is absent. If docling-serve returns {"document": null} (key present, value None), the default does not apply, so you get None.get("json_content") → AttributeError, which escapes as an unhandled 500 rather than a clean DoclingServeError.
payload.get("document") or {} collapses both the missing-key and the explicit-null cases to {}, so you always get the intended DoclingServeError.

src/utils/ingest_preview_flag.py
```diff
@@ -0,0 +1,24 @@
...
+    set ``OPENRAG_INGEST_PREVIEW_ENABLED=true`` to turn it on. The run-mode
+    check is AND-ed with the flag: preview is only ever available in OSS and
+    SaaS run modes (never on_prem), and even there only when the flag is set.
+    """
+    if not _ingest_preview_flag_enabled():
+        return False
+    return is_run_mode_oss() or is_run_mode_saas()
+
+
+def _ingest_preview_flag_enabled() -> bool:
```
mpawlow: ### (2a) [Minor] `is_ingest_preview_enabled` reads `os.environ` directly, outside `config/settings.py`

#### Problem

- `src/utils/ingest_preview_flag.py:23` calls `os.getenv("OPENRAG_INGEST_PREVIEW_ENABLED", "false")` directly. 
- Architectural pattern: *"`config/settings.py` consumes env into a typed `OpenRAGConfig` object via `config_manager`. Treat this module as the only place that should read `os.environ` directly; everything else imports the parsed values."* 
- CodeRabbit flagged it today (2026-07-14) as a Major "path instructions" violation.

#### Background Information

- This is real, but `os.getenv` outside `config/settings.py` is already a widespread, pre-existing pattern in this codebase (`rbac_service.py`, `user_service.py`, `encryption.py`, `db/engine.py`, `logging_config.py`, connector modules, etc. — 60+ call sites). 
- Note: The convention is aspirational/partially-enforced rather than actually CI-gated, so this is not a regression unique to this PR

#### Code References

- `src/config/settings.py` (existing `OpenRAGConfig` pattern, e.g. how `OPENRAG_RBAC_ENFORCE` or similar booleans are typically exposed)

#### Potential Solution

- Add `ingest_preview_enabled: bool` to `OpenRAGConfig` in `config/settings.py`, parsed once at config-load time, and have `is_ingest_preview_enabled()` read `get_openrag_config().ingest_preview_enabled` (AND-ed with the existing run-mode check) instead of calling `os.getenv` itself.

#### Alternative Solutions

- Leave as-is and accept the existing codebase-wide inconsistency (lowest effort, matches majority precedent elsewhere in `utils/`).

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM  🚀 
- See Minor PR comment (2a) for optional consideration
- All 7 findings from **Code Review 1** (1a–1g) were addressed 
- Verified via diff + re-run tests (55/55 passing, ruff/mypy clean).

edwinjosechittilappilly:
@coderabbitai please check if the feature is completely under the feature flag is_ingest_preview_flag_enabled ensure its in settings.

edwinjosechittilappilly:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> echo "=== run_mode_utils.py ==="
> cat -n src/utils/run_mode_utils.py
> 
> echo "=== settings module search ==="
> fd -e py settings
> echo "---"
> rg -n "class Settings|OPENRAG_RUN_MODE|BaseSettings|pydantic" -g '*.py' src | grep -i settings | head -30
> 
> echo "=== search for OPENRAG_INGEST_PREVIEW_ENABLED usages ==="
> rg -n "OPENRAG_INGEST_PREVIEW_ENABLED|is_ingest_preview_enabled|_ingest_preview_flag_enabled" -g '*.py'
> 
> echo "=== check all call sites of is_ingest_preview_enabled to confirm gating ==="
> rg -n "is_ingest_preview_enabled\(" -g '*.py' -A2 -B2
> 
> echo "=== check preview_mode usages that might bypass the flag ==="
> rg -n "preview_mode|preview:" -g '*.py' src | grep -v tests
> ```
> 
> Length of output: 3469
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> echo "=== git branch/log ==="
> git branch -a
> git log --oneline -5
> 
> echo "=== full run_mode_utils.py line count ==="
> wc -l src/utils/run_mode_utils.py
> 
> echo "=== search whole repo case-insensitive ==="
> rg -ni "ingest_preview_enabled|ingest_preview_flag" -g '*.py'
> 
> echo "=== search .env.example ==="
> cat .env.example | grep -n -i preview -A3 -B3
> 
> echo "=== check settings.py for preview mentions ==="
> rg -ni "preview" src/config/settings.py
> 
> echo "=== check where preview_mode is set without gating check ==="
> rg -n "preview_mode" src --include=*.py -B3 -A1 | head -150
> ```
> 
> Length of output: 1429
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> echo "=== check if ingest_preview files exist ==="
> fd ingest_preview
> echo "=== fetch PR branch explicitly ==="
> git fetch origin pull/2051/head:pr-2051 2>&1 | tail -20
> git log --oneline -10 pr-2051 2>&1
> echo "=== checkout pr branch ==="
> git checkout pr-2051 2>&1 | tail -5
> echo "=== now check run_mode_utils.py ==="
> cat -n src/utils/run_mode_utils.py
> ```
> 
> Length of output: 3010
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> echo "=== ingest_preview_flag.py ==="
> cat -n src/utils/ingest_preview_flag.py
> 
> echo "=== usages ==="
> rg -n "ingest_preview_flag|is_ingest_preview_enabled" -g '*.py'
> 
> echo "=== .env.example preview section ==="
> grep -n -i "preview" -A5 -B5 .env.example
> 
> echo "=== settings.py preview mentions ==="
> rg -ni "preview" src/config/settings.py
> 
> echo "=== check ingest_preview.py full ==="
> cat -n src/api/ingest_preview.py
> 
> echo "=== check router.py preview flow ==="
> rg -n "preview" src/api/router.py -B2 -A2
> 
> echo "=== check connectors.py preview flow ==="
> rg -n "preview" src/api/connectors.py -B2 -A2
> 
> echo "=== check v1 documents ingest preview test to see endpoint ==="
> rg -n "preview" tests/unit/api/test_v1_documents_ingest_preview.py | head -30
> ```
> 
> Length of output: 2218
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> echo "=== usages of is_ingest_preview_flag_enabled / is_ingest_preview_enabled ==="
> rg -n "ingest_preview_flag_enabled|is_ingest_preview_enabled" -g '*.py' || true
> ```
> 
> Length of output: 391
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> pwd
> git status | head -5
> echo "---"
> rg -n "is_ingest_preview_enabled" --type py || true
> echo "---direct file check---"
> grep -rn "is_ingest_preview_enabled" src tests || true
> ```
> 
> Length of output: 460
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> find / -name "ingest_preview_flag.py" 2>/dev/null
> echo "---"
> cat -A src/utils/ingest_preview_flag.py 2>/dev/null | head -5
> echo "---byte count---"
> wc -c src/utils/ingest_preview_flag.py 2>/dev/null
> echo "---python grep---"
> python3 -c "
> with open('src/utils/ingest_preview_flag.py') as f:
>     content = f.read()
> print(repr(content[:200]))
> print('is_ingest_preview_enabled' in content)
> "
> ```
> 
> Length of output: 1074
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> python3 -c "
> import subprocess, os
> matches = []
> for root, dirs, files in os.walk('.'):
>     if '.git' in root:
>         continue
>     for fn in files:
>         if fn.endswith('.py'):
>             path = os.path.join(root, fn)
>             try:
>                 with open(path, errors='ignore') as f:
>                     content = f.read()
>             except Exception:
>                 continue
>             if 'is_ingest_preview_enabled' in content or 'is_ingest_preview_flag_enabled' in content:
>                 for i, line in enumerate(content.splitlines(), 1):
>                     if 'is_ingest_preview_enabled' in line or 'is_ingest_preview_flag_enabled' in line:
>                         matches.append(f'{path}:{i}: {line.strip()}')
> for m in matches:
>     print(m)
> "
> ```
> 
> Length of output: 2999
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> sed -n '260,300p' src/config/settings.py
> 
> echo "=== .env.example preview ==="
> python3 -c "
> with open('.env.example') as f:
>     content = f.read()
> import re
> idx = content.lower().find('preview')
> print(content[max(0,idx-200):idx+400])
> "
> 
> echo "=== v1 documents ingest preview file ==="
> find . -iname "*v1*document*" -o -iname "*v1_ingest*" 2>/dev/null | grep -v node_modules
> 
> echo "=== check test_v1_documents_ingest_preview.py content ==="
> sed -n '1,60p' tests/unit/api/test_v1_documents_ingest_preview.py
> ```
> 
> Length of output: 4523
> 
> `@edwinjosechittilappilly` Verified — the feature is fully gated behind `is_ingest_preview_enabled()`, and it does live in settings.
> 
> * `src/config/settings.py` (Line 276) — `is_ingest_preview_flag_enabled()` reads `OPENRAG_INGEST_PREVIEW_ENABLED` (default `"false"`), owning the raw env parsing consistently with the other feature-flag accessors in that module (e.g. `is_azure_blob_enabled`).
> * `src/utils/ingest_preview_flag.py` — `is_ingest_preview_enabled()` wraps the settings flag and AND-s it with `is_run_mode_oss() or is_run_mode_saas()`, so it's never available in `on_prem` even if someone flips the flag.
> * All three entrypoints consistently gate through this same function:
>   
>   * `src/api/router.py` Line 48: `preview_mode = requested_preview and is_ingest_preview_enabled()`
>   * `src/api/connectors.py` Line 933: `preview_mode = body.preview and is_ingest_preview_enabled()`
>   * `src/api/ingest_preview.py` Line 27: `_require_preview_task` returns 404 when `not is_ingest_preview_enabled()`
> * `.env.example` documents it as commented-out/disabled by default (`# OPENRAG_INGEST_PREVIEW_ENABLED=false`).
> * Unit tests confirm off-by-default behavior: `tests/unit/utils/test_ingest_preview_flag.py`, `tests/unit/test_router_preview_upload.py`, `tests/unit/api/test_connector_sync_response.py`, `tests/unit/api/test_ingest_preview.py`.
> * There's also a regression fix (commit `4f779157`, "fix: forward explicit preview flag from v1 ingest endpoint") with a dedicated test (`tests/unit/api/test_v1_documents_ingest_preview.py`) ensuring the v1 endpoint doesn't accidentally bypass the flag via a bug in form-default forwarding.
> 
> I didn't find any path that sets `preview_mode=True` without first passing through `is_ingest_preview_enabled()`. `IngestPreviewService` and `TaskService` trust the already-gated `preview_mode` value, so this should be safe to merge with the feature disabled by default.
> 
> ʘ‿ʘ

@Wallgau  we would need to make it availabel for onprem also in upcoming PRs.

PR #2052, opened 2026-07-08
fix: OAuth connector callback redirect fix

mfortman11:
**The Issue**
Users get stuck on the connector redirect page

This can be tested using Google OAuth or Sharepoint

**The Bug**
The callback page's `handleCallback` async function and redirect timeout were both inside a single `useEffect` that used a cancelled closure variable. Two things killed the redirect:

1. In dev (React Strict Mode): React double-mounts → cleanup sets cancelled = true → the async `handleCallback` from the first mount continues and sets `status = "success"`, but the `if (!cancelled)` guard prevents the redirect timeout from being scheduled.
2. In pre-prod/production: `refreshAuth` was not wrapped in `useCallback`, so it got a new function reference on every `AuthProvider` render. Since `refreshAuth` was in the effect's dependency array, any `AuthProvider` re-render (from `checkAuth()` or `fetchPermissions()` running on mount) triggered the effect cleanup/re-run cycle, producing the same `cancelled = true` race condition.

The fix (two changes):

- `callback/page.tsx`: Split the redirect into a separate `useEffect` that reacts to `status === "success" && redirectTarget`. This is state-driven and works regardless of how many times the callback effect's closure gets "cancelled."
- `auth-context.tsx`: Wrapped `refreshAuth` in `useCallback` to stabilize its reference, preventing unnecessary effect re-runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth callback handling to complete the backend exchange reliably and redirect users after sign-in in both connector and IBM flows.
  * Added stronger validation, clearer UI status/error rendering, and more robust cleanup to prevent stale or duplicate callback processing.
* **Refactor**
  * Updated auth refresh to use a stable function reference while preserving existing behavior.
* **Chores**
  * Upgraded React Doctor tooling in CI for more reliable frontend health checks.

### DIFF
--- a/.github/workflows/lint-frontend.yml
+++ b/.github/workflows/lint-frontend.yml
@@ -71,10 +71,10 @@ jobs:
       - name: React Doctor score check
         working-directory: frontend
         run: |
-          SCORE=$(npx -y react-doctor@0.5.8 --score --yes)
+          SCORE=$(npx -y react-doctor@0.7.2 --score --yes)
           echo "React Doctor score: $SCORE"
           if [ "$SCORE" -lt 40 ]; then
             echo "::error::React Doctor score $SCORE is below the minimum threshold of 40"
-            npx -y react-doctor@0.5.8 --verbose --yes
+            npx -y react-doctor@0.7.2 --verbose --yes
             exit 1
           fi
--- a/.github/workflows/react-doctor.yml
+++ b/.github/workflows/react-doctor.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: millionco/react-doctor@v2
         with:
           directory: frontend
-          version: "0.5.8"
+          version: "0.7.2"
         # Advisory by default: React Doctor reports findings on every PR — a
         # sticky summary comment, inline review comments, and a commit status
         # with the health score — but never fails the check, so it won't red-X
--- a/frontend/app/auth/callback/page.tsx
+++ b/frontend/app/auth/callback/page.tsx
@@ -81,15 +81,15 @@ function AuthCallbackContent() {
   });
 
   useEffect(() => {
-    const callbackKey = `callback_processed_${code}`;
-
-    if (sessionStorage.getItem(callbackKey)) return;
-    sessionStorage.setItem(callbackKey, "true");
-
     if (validationError) {
       cleanupOAuthStorage();
       return;
     }
+
+    const callbackKey = `callback_processed_${code}`;
+
+    if (sessionStorage.getItem(callbackKey)) return;
+    sessionStorage.setItem(callbackKey, "true");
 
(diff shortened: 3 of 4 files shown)


lucaseduoli (approved):
Working nicely for me!

mfortman11:
**React Doctor fixes**

1. `no-fetch-in-effect` — fetch() inside useEffect

The problem: The original code had a raw fetch("/api/auth/callback", ...) inside a useEffect, wrapped in an async handleCallback() function. React can re-run effects when dependencies change or double-fire them in dev (Strict Mode). This was the root cause of users getting stuck on "You will be redirected shortly" — the effect's cancelled closure variable got set to true during a re-run, so the redirect timeout was never scheduled even though the fetch succeeded.

The fix: We moved the fetch into a React Query useMutation (the exchangeCode function). The useEffect now just calls exchangeCode(...) — it no longer contains fetch directly. React Query manages the request lifecycle independently from the effect lifecycle, so even if the effect re-runs or cleans up, the mutation continues and its onSuccess/onError callbacks still fire.

---
2. `query-mutation-missing-invalidation` — useMutation with no cache update

The problem: The useMutation we introduced had no cache invalidation. After exchanging the OAuth code, a new data source connection exists on the backend, but React Query's cached connectors list doesn't know about it.

The fix: Added onSettled to the mutation that calls queryClient.invalidateQueries({ queryKey: ["connectors"], exact: false }). This tells React Query "the connectors data is stale, refetch it." In practice users wouldn't see stale data because the page redirects immediately, but it's correct hygiene — if anything delays the redirect, the cache is right.

---
3. `no-chain-state-updates` — state setters inside effects causing render cascades

The problem: The useEffect was calling setPurpose(detectedPurpose) and setValidationError(...) synchronously. Each setState call triggers a re-render, and the resulting state fed into derived values (status) that were dependencies of the redirect effect — creating a chain: effect fires → sets state → re-render → second effect fires. Each link is a wasted render cycle.

The fix: We pulled both values out of the effect entirely:

- purpose became a useState with a lazy initializer — useState(() => readFromLocalStorage()). It computes once on mount and never changes. We need it "sticky" because localStorage gets cleaned up after success; if it recomputed, it would return the wrong value.
- validationError became a plain computed variable during render — const validationError = errorParam ? "..." : !code ? "..." : null. URL params don't change, so this is stable. No state needed at all.
- status and error also became computed values derived from validationError and callbackMutation.isError/isSuccess, instead of being managed with setStatus/setError calls inside the effect.

The effect now has zero setState calls of its own — it only calls exchangeCode() (which is a mutation trigger, not a state setter) and cleanupOAuthStorage() (which is a side effect on localStorage). All the render-driving state comes either from React Query's mutation state or from values computed during render.

mfortman11:
@coderabbit review

PR #2053, opened 2026-07-08
fix: japanese file names

edwinjosechittilappilly:
Root cause: COS/Azure/S3 connectors build document.id as the raw, non-ASCII object key ("bucket::報告書.pdf"), and that value was reused verbatim as document_id in the default Langflow ingestion path — the only ingestion path that puts non-ASCII text into ASCII-only HTTP headers and an unbounded-length OpenSearch identifier. Google Drive (opaque ASCII ID) and manual upload (content hash) were structurally immune; COS/Azure/S3 were the only paths exposed.

Fix applied (matches the already-proven split used by the non-Langflow ingestion path):
- document_index_writer.py — DocumentIndexContext gained a connector_file_id field, now written onto every indexed chunk when present.
- langflow_file_service.py — _resolve_document_id now hashes any supplied connector_file_id into a stable ASCII document_id, instead of using it verbatim; connector_file_id threads through run_ingestion_flow → _configure_ingest_callback → the JWT context.
- processors.py — the Langflow-path connector branch now passes connector_file_id=document.id (not document_id=document.id), and the post-ingest existence check / metadata sync now match on connector_file_id.
- connectors/service.py — stale comments corrected (dedupe/ACL-sync already matched both fields, no logic change needed there).
- Added 13 new unit tests (test_resolve_document_id_connector_file_id.py, test_document_index_writer_connector_file_id.py, test_bucket_connector_unicode_file_ids.py) — confirmed they fail on the old code and pass on the fix.
- Ran the full unit suite: same 14 pre-existing failures before and after (unrelated to this change — auth/settings/encryption tests), no new regressions.

## Summary by CodeRabbit

* **New Features**
  * Improved connector file handling so file identifiers are tracked more consistently during ingestion and indexing.
  * Added support for preserving non-ASCII bucket/object names in connector file ID round trips.

* **Bug Fixes**
  * Fixed matching and metadata updates so ingested chunks are found using the correct connector file identifier.
  * Ensured document IDs are generated consistently when a connector file ID is provided.

### DIFF
--- a/src/connectors/service.py
+++ b/src/connectors/service.py
@@ -258,9 +258,11 @@ class ConnectorService:
             raise RuntimeError("Backend OpenSearch write client is unavailable")
 
         # Update ACL if changed (hash-based skip optimization).
-        # Match both document_id and connector_file_id: non-Langflow connector
-        # chunks store the connector id in connector_file_id (document_id holds the
-        # content hash), while Langflow chunks store it in document_id.
+        # Match both document_id and connector_file_id: both the Langflow and
+        # non-Langflow ingestion paths store the raw connector id in
+        # connector_file_id (document_id holds a content/id hash), except for
+        # pre-migration chunks indexed before that split existed, which only
+        # have document_id set to the raw connector id.
         acl_result = await update_document_acl(
             document_id=document.id,
             acl=document.acl,
@@ -286,9 +288,10 @@ class ConnectorService:
             await write_client.update_by_query(
                 index=self.index_name,
                 body={
-                    # Match both fields: non-Langflow chunks carry the connector id
-                    # in connector_file_id (document_id is the content hash),
-                    # Langflow chunks carry it in document_id.
+                    # Match both fields: both ingestion paths carry the raw
+                    # connector id in connector_file_id (document_id is a
+                    # content/id hash); pre-migration chunks only have it in
+                    # document_id.
                     "query": {
                         "bool": {
                             "should": [
--- a/src/connectors/service.py
+++ b/src/connectors/service.py
@@ -302,6 +302,10 @@ class ConnectorService:
                             "should": [
                                 {"term": {"document_id": document.id}},
                                 {"term": {"connector_file_id": document.id}},
+                                # See check_document_exists (models/processors.py):
+                                # some indices predate the explicit keyword
+                                # mapping for this field.
+                                {"term": {"connector_file_id.keyword": document.id}},
                             ],
                             "minimum_should_match": 1,
                         }
--- a/src/connectors/service.py
+++ b/src/connectors/service.py
@@ -231,7 +231,6 @@ class ConnectorService:
                         owner_user_id,
                         connector_type,
                         jwt_token,
-                        id_field="connector_file_id",
                     )
 
                 return {
@@ -246,7 +245,6 @@ class ConnectorService:
         owner_user_id: str,
         connector_type: str,
         jwt_token: str = None,
-        id_field: str = "document_id",
         indexed_filename: str | None = None,
     ):
         """Update indexed chunks with connector-specific metadata"""
--- a/src/connectors/service.py
+++ b/src/connectors/service.py
@@ -135,7 +135,7 @@ class ConnectorService:
                 docling_polling_service=self.task_service.docling_polling_service
                 if self.task_service
                 else None,
-                document_id=document.id,
+                connector_file_id=document.id,
                 source_url=document.source_url,
                 allowed_users=allowed_users,
                 allowed_groups=allowed_groups,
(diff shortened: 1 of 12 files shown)


mpawlow (commented):
## Functional Review 1

- Encountered problem with ingestion using **Azure Blob Storage** connector
- See **Backend** and **Langflow** logs in subsequent comments

Image

mpawlow:
### OpenRAG Logs (Azure)

- [lf-logs.txt](https://github.com/user-attachments/files/29813433/lf-logs.txt)
- [be-logs.txt](https://github.com/user-attachments/files/29813434/be-logs.txt)

edwinjosechittilappilly:
@coderabbitai check why is connector_file_id a required and why not the earlier document_id

mpawlow:
## Functional Review 2

- ✅ Successfully ingested multiple documents with Japanese filenames from the **Azure Blob Storage** connector

image

src/models/processors.py
```diff
@@ -99,14 +100,33 @@ async def check_document_exists(
...
         retry_delay = 1.0
 
+        # Some deployments' indices predate connector_file_id's addition to the
+        # explicit mapping (config/settings.py), so OpenSearch dynamically
+        # mapped it as analyzed `text` (with a `.keyword` multi-field) instead
+        # of the intended `keyword` type. A plain term query against such a
+        # field tokenizes the query value and rarely matches the raw id, so
+        # also match its `.keyword` multi-field. document_id has always been
+        # explicitly `keyword` since index creation and never has this issue.
+        if field == "connector_file_id":
```
mpawlow, line 111: ### (1a) [Normal] `check_document_exists` fails mypy — CI-breaking type error

#### Problem

- The new branch added to `check_document_exists` assigns `query` two structurally different dict shapes depending on `field`, without an explicit type annotation. Mypy infers the variable's type from the first (`if`) branch and then rejects the second (`else`) assignment as incompatible, failing the type check.

#### Potential Solution

Add an explicit `dict[str, Any]` annotation so mypy doesn't narrow `query` to the shape of the first branch:

```python
query: dict[str, Any] = {"term": {field: file_hash}}
if field == "connector_file_id":
    query = {
        "bool": {
            "should": [
                {"term": {field: file_hash}},
                {"term": {f"{field}.keyword": file_hash}},
            ]
        }
    }
```

(`Any` is already imported at the top of this file.)

#### Alternative Solutions

- Extract the branch into a small helper function with a `-> dict[str, Any]` return annotation (e.g. `_build_exists_query(field, file_hash)`), mirroring the existing `_build_id_query` helper pattern already used in `src/utils/acl_utils.py` for the same dual-field-matching problem. Slightly more reusable if a third caller needs the same query shape later.

src/connectors/service.py
```diff
@@ -226,7 +227,11 @@ async def process_connector_document(
                 if result["status"] in ["indexed", "unchanged"]:
                     # Update all chunks with connector-specific metadata
                     await self._update_connector_metadata(
-                        document, owner_user_id, connector_type, jwt_token
+                        document,
+                        owner_user_id,
+                        connector_type,
+                        jwt_token,
+                        id_field="connector_file_id",
```
mpawlow: ### (1b) [Normal] `id_field` parameter of `_update_connector_metadata` is accepted but never used

#### Problem

`_update_connector_metadata(..., id_field: str = "document_id", ...)` takes an `id_field` parameter, and both call sites in this PR now pass `id_field="connector_file_id"`. However, the parameter is never referenced inside the function body — the ACL update (`update_document_acl(..., id_fields=("document_id", "connector_file_id"))`) and the metadata `update_by_query` both hard-code matching against **both** fields (plus `.keyword`) regardless of what `id_field` is set to.

#### Code References

`src/connectors/service.py:243-277` (parameter declared, never read) and call sites at `src/connectors/service.py:234` and `src/models/processors.py:1226`.

#### Potential Solution

Either wire `id_field` into the query construction (e.g. via the same `_build_id_query`-style helper used in `acl_utils.py`), or remove the now-vestigial parameter and its call-site arguments if always matching both fields is the intended, permanent behavior.

#### Alternative Solutions

- Delegate `_update_connector_metadata`'s query building to `utils.acl_utils._build_id_query` directly, so there is a single source of truth for the "match both id fields + `.keyword`" logic instead of three independent hard-coded copies of it (this file, `processors.py`, `acl_utils.py`).

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀
   - ✅ All unit tests pass
   - ✅ Manual functional test passes
- PR comment (1a) needs to be addressed RE: Mypy linting errors
- See PR comment (1b) for pre-merge consideration

src/connectors/service.py
edwinjosechittilappilly: Removed id_field

edwinjosechittilappilly:
Screenshot 2026-07-13 at 11 39 10 AM

PR #2058, opened 2026-07-08
feat: apply selected files to filter

mfortman11:
I always felt like if you select files in the knowledge table then click create filter those files should be auto added now they are

https://github.com/user-attachments/assets/04d7e873-39ea-41f1-9a6f-fb1a3e0a00de

## Summary by CodeRabbit

* **New Features**
  * Knowledge filters now stay in sync with your grid row selection, so selected sources always match what you’ve chosen.
  * When creating a new filter, the initial selected sources are prefilled from your current selection instead of always starting with the default “all sources” option.

* **Bug Fixes**
  * Prevents stale or mismatched source selections by clearing and updating the selected sources as the grid selection changes.


Wallgau (approved):
approved but need to address React Doctor comment

PR #2063, opened 2026-07-09
Support a custom base URL for the OpenAI embedding/LLM provider (self-hosted gateway support)

### LINKED ISSUE #2060: Support a custom base URL for the OpenAI embedding/LLM provider (self-hosted gateway support)
## Problem

OpenRAG's OpenAI provider always talks to `api.openai.com`. There's no way to point it at a self-hosted OpenAI-compatible gateway (a LiteLLM proxy, vLLM, etc.) without forking, and no supported env var for it today.

This blocks a common self-hosted deployment pattern: routing all embedding/LLM traffic through a centrally-managed gateway for unified cost tracking, virtual keys, and provider abstraction - including providers OpenRAG doesn't natively support (e.g., Cohere Embed Multilingual served via AWS Bedrock, reached through the gateway rather than directly).

## Current behavior

- `AsyncOpenAI(api_key=api_key)` is constructed with no `base_url` override in `src/config/settings.py`'s `patched_async_client` property - the client used for both query-time embedding (`search_service.py`) and non-Langflow ingest (`processors.py`).
- `OpenAIProviderConfig` (`src/api/settings/models.py`) has no `base_url` field.
- The Langflow ingestion flow's unified `EmbeddingModel` component has an `api_base` field, but `flows_service.py`'s `field_mappings["api_base"]` only maps it for `ollama` today, not `openai`.

## Proposed scope

- Add `openai_base_url` to the OpenAI provider config (`config_manager.py`'s `OpenAIConfig` / the settings models), following the same shape as `WatsonXConfig.endpoint`.
- Thread it into the `AsyncOpenAI(api_key=api_key, base_url=...)` construction in `patched_async_client` (`settings.py`).
- Add an `"openai": "OPENAI_API_BASE"` entry to `field_mappings["api_base"]` in `flows_service.py`, plus the matching Langflow global variable in `langflow_sync.py`, so the Langflow-flow ingest path can also point at the same custom endpoint.
- Document `OPENAI_BASE_URL` (or whatever the field ends up named) in `.env.example` alongside `OPENAI_API_KEY`.

## Correctness note for gateway users

OpenRAG's embedding client is patched by `agentd.patch.patch_openai_with_mcp`. A model name with no recognized non-OpenAI litellm prefix resolves to `provider == "openai"` and takes the real OpenAI SDK code path - an actual HTTP call to whatever `base_url` is configured - rather than the in-process litellm SDK path. That means any non-standard parameters a downstream model needs (e.g. Cohere's required `input_type` when the gateway is fronting Cohere-on-Bedrock) must be passed via the OpenAI SDK's `extra_body`, not as a bare kwarg. Worth a docs callout once this lands, since a directly-added litellm-native provider (e.g. a future first-class Bedrock/OCI provider) would need `input_type` passed differently (see the companion Bedrock/OCI issues).

## Acceptance criteria

- Setting the base-URL override redirects both query-time embedding and non-Langflow ingest to the custom endpoint.
- Langflow-flow ingest also respects it when set.
- Existing OpenAI-default behavior (no override configured) is unchanged.
- `watsonx`/`ollama`/`anthropic` paths unchanged.

galovics:
Fixes #2060

## Summary

Adds an `openai.base_url` override (config field, `OPENAI_BASE_URL` env var, and settings API field) so
OpenRAG's OpenAI provider can be pointed at a self-hosted OpenAI-compatible gateway (e.g. a LiteLLM proxy)
instead of `api.openai.com`.

- Threads the override into all three `AsyncOpenAI(...)` construction sites in `patched_async_client`
  (`src/config/settings.py`), including the HTTP/2 capability probe — leaving that one unpatched would
  still make a live call to `api.openai.com` on every client init even with a custom gateway configured.
- Extends the Langflow-flow ingest path too: `field_mappings["api_base"]` in `flows_service.py` previously
  only mapped `ollama` → `OLLAMA_BASE_URL`; this adds `openai` → `OPENAI_API_BASE`, and `langflow_sync.py`
  sets that Langflow global variable when the override is configured.
- Wires the settings API request/response models and persists the value through both the `update_settings`
  and `onboarding` handlers, mirroring the existing `watsonx_endpoint` pattern — deliberately *not* marking
  `openai.configured = true` from a bare `base_url`, since it's an optional override, not a credential, and
  `configured` is consumed elsewhere as "this provider is a complete, usable setup."

## Why

OpenRAG's OpenAI provider always talks to `api.openai.com` today. There's no way to route embedding/LLM
traffic through a centrally-managed self-hosted gateway without forking — a common pattern for teams running
their own LiteLLM proxy for unified cost tracking, virtual keys, or reaching providers OpenRAG doesn't
natively support (e.g. Cohere Embed Multilingual served via AWS Bedrock, reached through the gateway).

## A correctness note for anyone extending this

OpenRAG's embedding client is patched by `agentd.patch.patch_openai_with_mcp`. A model name with no
recognized non-OpenAI litellm prefix resolves to `provider == "openai"` and takes the real OpenAI SDK code
path (an actual HTTP call to whatever `base_url` is configured), not the in-process litellm SDK path. That
means any non-standard parameter a downstream model needs (e.g. Cohere's `input_type` when the gateway
fronts Cohere-on-Bedrock) must be passed via the OpenAI SDK's `extra_body`, not as a bare kwarg, on this
specific code path — verified against the real `agentd` patch source, not assumed.

## Testing

- Unit tests for the config field (default/round-trip/env-override), the `patched_async_client` construction
  (asserts the `base_url` kwarg directly across all three construction sites, with and without the override
  configured), the `flows_service.py` field-mapping change, and the settings-API write path (persists
  correctly, doesn't mark `configured`).
- A real integration test against a locally running LiteLLM proxy + a mock embeddings stub: configures the
  override, builds the real `patched_async_client`, calls `.embeddings.create(..., extra_body={"input_type":
  "search_document"})`, and confirms both a successful response and that the stub actually received
  `input_type="search_document"` — proving the full chain (config → real OpenAI SDK → real Docker LiteLLM
  proxy → stub) works, including the `extra_body` passthrough mechanism.
- Full suite: `uv run pytest tests/unit -q` → 901 passed, 13 failed. The 13 failures are pre-existing at the
  base commit (unrelated to this change — collection errors from a partially-landed refactor upstream); zero
  new failures.

## Out of scope / follow-ups

- Frontend/Settings-UI wiring for the new field is not included — this PR is backend-only.
- The current integration test relies on a test-only `litellm.register_model(...)` call to teach litellm's
  local provider table that an arbitrary gateway route alias should resolve to `openai`; production code
  does not replicate this. An arbitrary custom-gateway route name that litellm doesn't already recognize as
  OpenAI (real model name, or `openai/`-prefixed) will still raise `BadRequestError` in `agentd` before any
  HTTP call — a pre-existing `agentd`/litellm constraint, not something this PR changes. Worth a tracked
  follow-up if first-class support for arbitrary route aliases is wanted.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom OpenAI-compatible endpoints across model discovery, chat, embeddings, and provider validation.
  * Endpoints can be configured through environment variables, settings, onboarding, or individual model requests.
  * Configured endpoints now synchronize with relevant flows and global settings.
* **Bug Fixes**
  * Chat requests consistently use the configured language model.
  * Removing OpenAI settings clears saved custom endpoints.
* **Documentation**
  * Updated `.env.example` with `OPENAI_BASE_URL`.
* **Tests**
  * Added coverage for routing, persistence, validation, discovery, and proxy integration.


galovics:
Companion PRs for the other two Cohere-hosting backends: #2064 (AWS Bedrock) and #2065 (OCI GenAI). All three are independent and can be reviewed/merged in any order, though they touch some of the same files and may need a rebase if merged out of order.

vardumper:
@galovics I tried this PR by applying its patch file, but wasn't able to point OpenRAG to a custom OpenAI compatible endpoint (during onborading screens) nor in the Langflow Settings. (setting global Langflow variable OPENAI_API_KEY seems to also validate it against openai.com).
Are there more steps needed to set this up?

vardumper:
@galovics i have a basic configuration working now with a custom openai endpoint. no tests, though. could you have a look at my fork? here's an outline of what is changed there https://github.com/vardumper/openrag/blob/main/ERIK-MODS.md?

galovics:
@vardumper thanks for your verification. Frankly speaking I started working this out but wasn't able to finish. Now it's in a better shape, feel free to check it.

galovics:
Hey @edwinjosechittilappilly @mfortman11 @lucaseduoli @phact — following up on this, it's been about 4 weeks since the last ping. This is rebased onto current `main`, all CI checks are green, and I've since gone through and fixed several real bugs CodeRabbit flagged (validation gaps, a stale Langflow global-variable reference) with regression tests for each. Would really appreciate a review when you get a chance — trying to get this merged soon.

vardumper:
So I did test again. One minor error:

```
0.441 npm warn tarball tarball data for sharp@file:stubs/sharp-0.35.0.tgz (null) seems to be corrupted. Trying again.
0.455 npm error code EUSAGE
0.459 npm error
0.459 npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
0.459 npm error
0.459 npm error Missing: sharp@ from lock file
```
You likely changed package.json (adding or modifying local dependencies like sharp), but did not update package-lock.json to match it.

I re-ran npm install inside frontend, rebuilt the containers, start everything up, added the two OPENAI_BASE_URL and OPENAI_API_BASE  env vars.
During onboarding, I am stuck at the screen where I'd put my API key, I cannot activate the "take values form env" switch, plus my custom Api Key is rejected as well. "Invalid, get a proper one here...").

galovics:
Thanks for the repro, that made this easy to nail down. You were right — onboarding never had a field for a custom base URL, so your `OPENAI_BASE_URL` was never actually reaching the key check, which is why a valid gateway key came back "invalid." While testing the fix I found two more spots with the same gap: model listing only recognized real OpenAI model names (so a custom gateway's models came back empty and the required model dropdown would've stayed stuck), and the health check had the same blind spot, so it'd throw a false "Incorrect API key" error even once everything else worked.

All fixed in 79a557e — there's now a Base URL field in onboarding and in the Settings page, and I tested the whole flow end to end against a fake gateway before pushing.

Two side notes: drop `OPENAI_API_BASE` from your env, we only read `OPENAI_BASE_URL`. And the `npm ci`/sharp error is a real but unrelated bug on main — `npm install` is still the right workaround for now.

Mind pulling latest and giving it another shot?

vardumper:
gotcha. sure thing. now that looks really nice. kudos. there seems to be a validation in onboarding, that disallows my custom OAI-compatible API key. (I get the invalid api key, go to openai.com to get..... message). 
instead of just reporting... how could i contribute to your PR? fork your fork and make a PR? :D

galovics:
> gotcha. sure thing. now that looks really nice. kudos. there seems to be a validation in onboarding, that disallows my custom OAI-compatible API key. (I get the invalid api key, go to openai.com to get..... message). instead of just reporting... how could i contribute to your PR? fork your fork and make a PR? :D

added you as contributor.

vardumper:
No more problems, I could report. Tested and everything worked smoothly (I tested without prior setting any .env vars). I could enter URL/Key in onboarding.

vardumper:
Hi Arnold. I did run into a problem on a fresh stack/setup where litllm receives a plain model name eg 'gpt-oss-120b' without a provider. In src/services/llm_gateway.py — always qualify the model with its provider:
```
-litellm_model = f"{provider}/{name}" if provider != "openai" else name
+litellm_model = f"{provider}/{name}"
```
This is valid for both real OpenAI (openai/gpt-4o) and third-party gateways (openai/Qwen3-Embedding-8B).

galovics:
> Hi Arnold. I did run into a problem on a fresh stack/setup where litllm receives a plain model name eg 'gpt-oss-120b' without a provider. In src/services/llm_gateway.py — always qualify the model with its provider:
> 
> ```
> -litellm_model = f"{provider}/{name}" if provider != "openai" else name
> +litellm_model = f"{provider}/{name}"
> ```
> 
> This is valid for both real OpenAI (openai/gpt-4o) and third-party gateways (openai/Qwen3-Embedding-8B).

good catch, done.

galovics:
Bumping again — it's been about 3 weeks since the last ping. Since then @vardumper tested end-to-end on a fresh stack and confirmed the onboarding flow (custom URL + key) works cleanly, and I fixed the model-qualification bug they found in `llm_gateway.py`. Still rebased clean onto current `main` (including the recent #2321 provider-identity refactor) with CI green. Would appreciate a look whenever you have a moment — happy to address anything else that comes up.

galovics:
@edwinjosechittilappilly @mfortman11 @lucaseduoli @phact - friendly nudge, ~2 weeks since the last ping. Rebased again onto current `main` (no conflicts, CI green). It has been tested end-to-end by an external user on a fresh stack. Happy to address anything in review.

PR #2064, opened 2026-07-09
Add AWS Bedrock as an embedding provider (Cohere Embed Multilingual via Bedrock)

### LINKED ISSUE #2061: Add AWS Bedrock as an embedding provider (Cohere Embed Multilingual via Bedrock)
## Motivation

For EU-sovereign / AWS-native deployments, Cohere Embed Multilingual served via AWS Bedrock (in-region, e.g. `eu-central-1`) is a common requirement that native `api.cohere.com` (US-only) can't satisfy. OpenRAG already embeds `litellm==1.84.0`, which has first-class Bedrock-Cohere support (`litellm/llms/bedrock/embed/cohere_transformation.py`), so this is additive wiring, not new architecture - following the same pattern as the existing `watsonx`/`ollama` providers.

## Scope (mirrors the watsonx/ollama provider pattern)

| File | Change |
| --- | --- |
| `src/config/config_manager.py` | New `BedrockConfig` dataclass (`region`, optional `access_key_id`/`secret_access_key` - support IAM role/IRSA as the no-explicit-creds default), add to `ProvidersConfig`, `get_provider_config()`, the provider-list loader, and env-override loading (`AWS_REGION_NAME`, etc.) |
| `src/api/settings/models.py`, `src/api/settings/helpers.py` | Add `bedrock` to `_EMBEDDING_PROVIDER_NAMES` and the `embedding_provider` regex |
| `src/services/models_service.py` | Add `"bedrock"` to `KNOWN_PREFIXES`; add `get_bedrock_models()` (static list, see below); **add a `bedrock` branch to `update_model_registry()`** (see Correctness risk 1) |
| `src/config/settings.py` | In `patched_async_client`, add an AWS credentials block (`AWS_REGION_NAME`, optionally `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`) mirroring the existing WatsonX block |
| `src/api/models.py`, `src/api/v1/models.py` | Add `get_bedrock_models()` dispatch |
| `src/api/provider_validation.py`, `src/api/provider_health.py` | Add a Bedrock validation/health check |
| Frontend | Add Bedrock as a selectable embedding provider with region + credential fields |

## Model list

- `bedrock/cohere.embed-multilingual-v3` (plus English/light variants and `cohere.embed-v4` where available in-region)

## No Langflow-flow ingest path

There is no Bedrock embeddings component in Langflow's component library. Recommend running with `DISABLE_INGEST_WITH_LANGFLOW=true` (an existing, tested flag - see `tests/integration/core/test_non_langflow_ingestion.py`) so both ingest (`processors.py`) and query (`search_service.py`) go through the native litellm-embedded path, which already supports Bedrock. Simpler than building a new Langflow component.

## Correctness risks

1. **Model registry - query-time is fail-fast.** `search_service.py` calls `get_litellm_model_name(model_name, strict=True)`, which raises `UnknownEmbeddingProvider` unless the model is in `_model_provider_registry`, populated only by `update_model_registry()`'s hardcoded per-provider branches. Every search will fail until a `bedrock` branch is added there. The ingest-time call in `processors.py` is non-strict and fails differently - it silently falls back to the raw model name and sends a malformed request rather than raising. Needs explicit test coverage for both failure modes.
2. **`input_type`.** Cohere v3/v4 requires `input_type` (`search_document` at ingest, `search_query` at query). Once the model resolves to a real litellm provider (which `bedrock/...` will), extra kwargs on `.embeddings.create()` flow straight into `litellm.aembedding(**kwargs)` via the `agentd` patch. Fix: pass `input_type="search_document"`/`"search_query"` as a plain kwarg in `search_service.py`/`processors.py`'s embedding calls, gated on Cohere-family model detection - not currently done for any provider since none of the existing ones need it.
3. **Field-name normalization.** Verified safe: `utils/embedding_fields.py`'s `normalize_model_name()` replaces `.`, `:`, `/`, `-` uniformly, so `cohere.embed-multilingual-v3` normalizes cleanly to `cohere_embed_multilingual_v3` with no collision risk. Add a unit test for this exact model name to lock it in.

## Acceptance criteria

- `EMBEDDING_PROVIDER=bedrock` + region/creds boots; Settings UI lets a user pick Bedrock + a model.
- Ingest (non-Langflow path) produces a `chunk_embedding_cohere_embed_multilingual_v3` field of the correct dimension.
- Query embeds with `search_query`, ingest embeds with `search_document`; a known query retrieves its source chunk (integration test).
- `openai`/`watsonx`/`ollama`/`anthropic` paths unchanged.

galovics:
Fixes #2061

## Summary

Adds AWS Bedrock as a first-class embedding provider, targeting Cohere Embed Multilingual (plus English v3
and v4 variants) served via Bedrock — for EU-sovereign / AWS-native deployments where native
`api.cohere.com` (US-only) doesn't work. Routes entirely through the `litellm==1.84.0` dependency OpenRAG
already ships (`litellm/llms/bedrock/embed/cohere_transformation.py`), so this is additive wiring, not a new
SDK integration — follows the existing `watsonx`/`ollama` per-provider pattern throughout.

- New `BedrockConfig` (region + optional access key/secret — left blank by default so IAM role/IRSA works
  with zero explicit credentials) wired through `config_manager.py`'s provider-config machinery, the
  settings API, and `patched_async_client`'s credential env-var block.
- `update_model_registry()` gets a Bedrock branch that registers its models when configured — required
  because `search_service.py` resolves models via `get_litellm_model_name(model, strict=True)`, which raises
  `UnknownEmbeddingProvider` on a registry miss. Without this branch, every query-time embedding call would
  hard-fail even with everything else correctly wired.
- `input_type` (`search_document` at ingest / `search_query` at query — Cohere's embed API requires it) is
  passed as a plain kwarg to `.embeddings.create(...)` in both `search_service.py` and `processors.py`,
  gated on Cohere-model detection. Bedrock resolves through a real litellm provider prefix
  (`bedrock/cohere.embed-multilingual-v3`), so this goes through `litellm.aembedding(**kwargs)` in-process —
  the plain-kwarg form is correct here (contrast with #2060's gateway case, where the same parameter needs
  `extra_body` instead, because that path takes the real OpenAI SDK route).
- No Langflow-flow component exists for Bedrock, so ingestion runs through OpenRAG's native (non-Langflow)
  processing pipeline — deliberately kept `bedrock` **out** of `_EMBEDDING_PROVIDER_NAMES` (the
  Langflow-flow-syncable list) and introduced a separate `_ALL_EMBEDDING_PROVIDER_NAMES` for general
  provider selection, since including it in the former would raise inside `flows_service.py`'s
  Langflow-flow-reapplication loop.

## Testing

80 new tests covering: the config dataclass and env-override chain, credential-shape validation (no live
calls), the model registry registration→resolution path end-to-end (the #1 correctness risk — a dedicated
test proves `get_litellm_model_name(strict=True)` resolves a bare Cohere model name to
`bedrock/cohere.embed-multilingual-v3` after registration, and raises when not configured), `input_type`
passed as a plain kwarg at both call sites (and *not* passed for non-Cohere models), and field-name
normalization for the Bedrock model names (verified distinct from the OCI PR's dotted variants).

All mocked at the HTTP/litellm-call layer — no live calls to `*.amazonaws.com` anywhere in the suite (there
are no real AWS credentials available to exercise this against). Full suite: `uv run pytest tests/unit -q`
→ 961 passed, 13 failed (pre-existing at the base commit, unrelated to this change — identical failing test
names before and after). Zero new failures.

## A note for reviewers on Bedrock's response shape

Cross-checked litellm's Bedrock-Cohere transformation against AWS's own documented API contract while
researching this. AWS Bedrock's Cohere embed response has two shapes depending on the `embedding_types`
param: a plain list of vectors (`"response_type": "embeddings_floats"`) when omitted or single-type — the
common/default path this PR exercises — versus a dict keyed by type (`"response_type":
"embeddings_by_type"`) when 2+ types are requested. litellm's shared Cohere response parser appears to
assume the dict shape unconditionally in at least one code path, which would be a latent bug in `litellm`
itself on the default single-type response — this PR doesn't touch that code and it's outside OpenRAG's
scope to fix, but flagging it here since it's relevant to anyone validating this integration against a real
Bedrock account, and I haven't been able to verify it without live credentials.

## Out of scope / follow-ups

- Frontend/Settings-UI wiring (region/access-key inputs) is not included — backend-only.
- `src/api/settings/endpoints.py` was not wired to accept `bedrock_region`/`bedrock_access_key_id`/
  `bedrock_secret_access_key` via the settings/onboarding REST API — Bedrock can currently only be
  configured via env vars (`BEDROCK_REGION`, etc.) or a hand-edited config file, consistent with its
  env-var/IAM-role-first credential model differing from the paste-an-api-key pattern other providers use.
  Happy to wire the REST path too if maintainers would rather have UI-driven configuration from the start.
- The static model list currently covers the three base model IDs (English v3, Multilingual v3, v4); the
  geo/global cross-region inference-profile IDs (e.g. `eu.cohere.embed-v4:0`, needed to reach embed-v4 from
  `eu-central-1`) were intentionally left out to keep this PR focused on the issue's stated target
  (`cohere.embed-multilingual-v3`, which *is* directly available in-region in `eu-central-1`) — can add them
  in a follow-up if cross-region routing is wanted.

## Summary by CodeRabbit

* **New Features**
  * Added AWS Bedrock embedding support for configuration, onboarding, provider selection, and model listings.
  * Added deployment settings for Bedrock region and optional credentials, including IAM-role authentication.
  * Bedrock Cohere embeddings now use appropriate search and ingestion input types.
* **Bug Fixes**
  * Improved Bedrock health checks and validation.
  * Prevented unsupported Bedrock synchronization with Langflow.
  * Improved embedding field naming to avoid collisions.
* **Tests**
  * Added coverage for Bedrock configuration, deployment, validation, discovery, and embedding behavior.


galovics:
Companion PRs for the other two Cohere-hosting backends: #2063 (custom gateway base URL support) and #2065 (OCI GenAI). All three are independent and can be reviewed/merged in any order, though they touch some of the same files and may need a rebase if merged out of order.

galovics:
Re CodeRabbit's `os.environ` path-instruction finding on `config_manager.py`: this matches the exact pre-existing pattern used by every other provider (watsonx, openai, ollama) in this same file, so I'm leaving it as-is rather than fixing it only for Bedrock (that would be inconsistent scope creep for a single-provider PR, not a fix). Happy to take this on as a separate cross-cutting refactor PR if maintainers want it addressed — flagging that here rather than silently ignoring it. Same call made on the identical finding in #2065.

galovics:
Hey @edwinjosechittilappilly @mfortman11 @lucaseduoli @phact — bumping this, ~4 weeks since the original ping. This is rebased onto current `main`, CI is green, and I've fixed a few things CodeRabbit caught since (a stale AWS region env var on client refresh, a missing per-call region kwarg, and — this one's a real functional gap — Bedrock's Cohere Embed v3 hard-limits input to 512 tokens per chunk and the ingest chunker had no cap for it). Would appreciate a look when you have time.

galovics:
Bumping again — ~3 weeks since the last ping. This has now been re-rebased onto `main` post-#2321 (the provider-identity/embedding-space refactor), including a fix for a regression that dropped Cohere's required `input_type` param from the new retrieval path. All tests still green. Let me know if there's anything blocking a look.

galovics:
@edwinjosechittilappilly @mfortman11 @lucaseduoli @phact - friendly nudge, ~2 weeks since the last ping. Rebased again onto current `main` (no conflicts, CI green). Since the last bump: Cohere's `input_type` now also reaches Langflow's own embedding calls (previously only the backend call sites had it), and an external user confirmed the same gap class on NVIDIA NIM in #2061. Happy to address anything in review.

PR #2065, opened 2026-07-09
Add OCI Generative AI as an embedding provider (Cohere Embed Multilingual via OCI GenAI)

### LINKED ISSUE #2062: Add OCI Generative AI as an embedding provider (Cohere Embed Multilingual via OCI GenAI)
## Motivation

Oracle Cloud Infrastructure's Generative AI service hosts Cohere Embed models (including `cohere.embed-multilingual-v3.0`) as a first-party offering. litellm 1.84.0 (already pinned in this repo) has dedicated OCI GenAI embedding support (`litellm/llms/oci/embed/transformation.py`, provider prefix `oci`, registered as `LlmProviders.OCI = "oci"`), so - like Bedrock - this is additive wiring on top of an already-embedded dependency, not a new SDK integration from scratch.

## Scope (same shape as the Bedrock request, different credential set)

| File | Change |
| --- | --- |
| `src/config/config_manager.py` | New `OCIConfig` dataclass: `user`, `fingerprint`, `tenancy`, `compartment_id`, `key_file` (or inline `key`), `region`; add to `ProvidersConfig` + `get_provider_config()` + the provider-list loader + env overrides (`OCI_USER`, `OCI_FINGERPRINT`, `OCI_TENANCY`, `OCI_COMPARTMENT_ID`, `OCI_KEY_FILE`) |
| `src/api/settings/models.py`, `src/api/settings/helpers.py` | Add `oci` to `_EMBEDDING_PROVIDER_NAMES` and the `embedding_provider` regex |
| `src/services/models_service.py` | Add `"oci"` to `KNOWN_PREFIXES`; static `get_oci_models()`; **add an `oci` branch to `update_model_registry()`** (same fail-fast risk as Bedrock, see below) |
| `src/config/settings.py` | In `patched_async_client`, add the OCI credential env block |
| `src/api/models.py`, `src/api/v1/models.py`, `src/api/provider_validation.py`, `src/api/provider_health.py` | Same dispatch/health pattern as other providers |
| Frontend | Add OCI GenAI as a selectable embedding provider with its credential fields |

## Model list

- `oci/cohere.embed-multilingual-v3.0` (also `-english-v3.0`, `-light-v3.0`, `-v4.0` per litellm's supported list for this provider)

## No Langflow-flow ingest path

Same situation as Bedrock - no OCI GenAI Langflow component exists. Use `DISABLE_INGEST_WITH_LANGFLOW=true`, native litellm-embedded path only.

## Correctness risks (shared mechanism with the Bedrock request)

1. **Model registry / fail-fast query path.** Same `update_model_registry()` requirement as Bedrock; same non-strict-vs-strict split in failure behavior between `processors.py` (ingest) and `search_service.py` (query).
2. **`input_type`.** litellm's OCI transformation explicitly maps OpenAI-style `input_type` to OCI's `inputType` field (`SEARCH_DOCUMENT`/`SEARCH_QUERY`/etc.), read from `optional_params.get("input_type")`. Same fix as Bedrock: pass `input_type=` as a plain kwarg on `.embeddings.create()`, gated on Cohere-family model detection.
3. **Field-name normalization.** `cohere.embed-multilingual-v3.0` (two dots) also normalizes cleanly and distinctly from the Bedrock variant (`cohere.embed-multilingual-v3`, one dot) via the existing `normalize_model_name()`. Add a unit test for this exact model name too, to guard against any future collision between the two.

## Acceptance criteria

- `EMBEDDING_PROVIDER=oci` + OCI credentials boots; Settings UI lets a user pick OCI GenAI + a model.
- Ingest produces a `chunk_embedding_cohere_embed_multilingual_v3_0` field of the correct dimension.
- Query embeds with `SEARCH_QUERY`, ingest embeds with `SEARCH_DOCUMENT` (post `input_type` mapping); a known query retrieves its source chunk.
- Other provider paths unchanged.

galovics:
Fixes #2062

## Summary

Adds Oracle Cloud Infrastructure (OCI) Generative AI as a first-class, embedding-only provider, targeting
Cohere Embed Multilingual (`oci/cohere.embed-multilingual-v3.0`) served through OCI's GenAI service.
`litellm==1.84.0` (already an OpenRAG dependency) has dedicated OCI GenAI embedding support
(`litellm/llms/oci/embed/transformation.py`), so — like the companion Bedrock PR (#2061) — this is additive
wiring on top of an already-embedded dependency, no new SDK required.

- New `OCIConfig` (`user`/`fingerprint`/`tenancy`/`compartment_id`/`key` or `key_file`/`region`) following
  the existing per-provider dataclass pattern, wired through `config_manager.py`, the settings API
  (including full read/write/removal support — `remove_oci_config` clears all fields, with the same
  "configure another provider first" guard and embedding-fallback logic other providers have), and
  `patched_async_client`'s credential env-var block.
- `update_model_registry()` gets an OCI branch (same fail-fast risk as Bedrock — `search_service.py`'s
  `get_litellm_model_name(model, strict=True)` raises `UnknownEmbeddingProvider` on every query without it).
  A dedicated test proves registration→resolution end-to-end.
- `input_type` is passed as a plain kwarg (not `extra_body` — same reasoning as #2061, this provider
  resolves through a real litellm prefix) at both the query (`search_service.py`) and ingest
  (`processors.py`) call sites, extracted into a small shared `utils/embedding_kwargs.py` helper used by
  both. litellm's OCI transformation maps the lowercase OpenAI-style `input_type` value to OCI's uppercase
  `inputType` enum internally, so this PR passes the lowercase value and relies on litellm for that mapping
  rather than re-implementing it.
- Credential-shape validation requires exactly `user`/`fingerprint`/`tenancy`/`compartment_id` plus one of
  `key`/`key_file` — matching litellm's own `validate_environment` requirement for this provider, confirmed
  against litellm's source during research rather than guessed.
- OCI has no Langflow-flow component. Adding `oci` to the general provider-name list initially leaked into
  `langflow_sync.py`'s Langflow-flow-reapplication loop during development (which has no OCI component and
  would raise) — caught and fixed by introducing a narrower `_LANGFLOW_EMBEDDING_PROVIDER_NAMES` constant
  scoped to just the providers with a Langflow component, separate from the general embedding-provider list.

## Testing

89 new tests: the config dataclass and full settings-API read/write/removal path, credential-shape
validation (missing fields, inline PEM vs. key_file, invalid paths — no live calls), the registry
registration→resolution path end-to-end, `input_type` as a plain kwarg at both call sites intercepted at the
real production call sites (`SearchService.search_tool`, `TaskProcessor.process_document_standard`) with
assertions that non-Cohere models get none of the extra kwargs, the Langflow-allowlist separation described
above, and field-name normalization distinguishing OCI's `cohere.embed-multilingual-v3.0` (two dots) from
Bedrock's `cohere.embed-multilingual-v3` (one dot, in #2061).

All mocked at the HTTP/embeddings-client layer — no live calls to `*.oraclecloud.com` anywhere (OCI's
request signing requires a real RSA key pair in an actual tenancy, not available here). Full suite:
`uv run pytest tests/unit -q` → 973 passed, 13 failed (pre-existing at the base commit, unrelated — identical
failing test names before and after). Zero new failures.

## Out of scope / follow-ups

- Frontend/Settings-UI wiring (a form for OCI's 6 credential fields, including a PEM-upload-or-textarea
  choice) is not included — backend-only. The settings API is fully wired to support it once a form exists.
- Secret-at-rest: `config_manager.py`'s encrypt-on-save logic only targets the field literally named
  `api_key`. OCI's sensitive field is `key` (an inline PEM private key), which is not run through that same
  encryption path. Not a live issue today (env-var-only configuration is the primary path this PR
  optimizes for), but worth addressing — either by nudging the eventual UI toward `key_file` (a path, not a
  secret payload) over inline `key`, or extending the encrypt/decrypt logic to cover `key` too.
- `provider_health_cache`'s cache key wasn't extended with OCI-specific fields, so a credential change while
  `embedding_provider=oci` won't bust the short-TTL health cache mid-window. Low impact (affects only the
  health-check banner's freshness, not embedding correctness).
- Manual verification against a live OCI tenancy (real credentials, real embed call) hasn't been done and
  couldn't be within this PR's constraints — recommended before this ships to users relying on it in
  production.

## Related

Companion PRs for the other two Cohere-hosting backends discussed in the tracking issues: #2060 (custom
gateway base URL support) and #2061 (AWS Bedrock). All three are independent and can be reviewed/merged in
any order, though they touch some of the same files (`config_manager.py`, `models_service.py`,
`settings.py`) and may need a rebase if merged out of order.

## Update: `instance_principal` / `workload_identity` auth methods

Follow-up to the initial `api_key`-only version of this PR. Adds `OCIConfig.auth_method`, a
`build_oci_signer()` factory, and wires it through both embedding call sites, the model-registry gate,
validation/health-checks, and — this final increment — the settings API and docs, so all three OCI auth
methods are configurable via `OCI_AUTH_METHOD` (env var) or `oci_auth_method` (settings API / onboarding),
not just `api_key`.

- **`api_key`** (default, unchanged): OCI's API Signing Key scheme — `user`/`fingerprint`/`tenancy`/`key` or
  `key_file`, as before.
- **`instance_principal`**: authenticates as this workload's identity when running on an OCI Compute
  instance. None of `user`/`fingerprint`/`tenancy`/`key`/`key_file` apply. Prerequisites (operator-side, not
  satisfiable by this code): the workload must actually run on an OCI Compute instance, that instance must
  belong to a dynamic group, and an IAM policy must grant that dynamic group `use generative-ai-family` in
  the target compartment.
- **`workload_identity`**: authenticates as this workload's identity as an OKE pod. None of
  `user`/`fingerprint`/`tenancy`/`key`/`key_file` apply. Prerequisites: the OKE cluster must have Workload
  Identity enabled, and the pod's service account must be configured for it (standard
  `KUBERNETES_SERVICE_HOST`-based auto-detection, per [OCI's docs](https://docs.oracle.com/en-us/iaas/Content/ContEng/Tasks/contenggrantingworkloadaccesstoresources.htm)).
- Validation switches from the existing `api_key` credential-shape check to constructing the real signer
  (`oci.auth.signers.InstancePrincipalsSecurityTokenSigner()` /
  `oci.auth.signers.get_oke_workload_identity_resource_principal_signer()`) and surfacing a
  `OCISignerConstructionError` naming the concrete unmet prerequisite if that fails.
- While wiring `oci_auth_method` into the settings API, found and fixed an ordering bug in
  `update_settings()`: its embedding-provider validation block resolved `oci_auth_method` from the
  pre-existing config *before* the new request-body field could override it, so a `PATCH /settings` that
  changed `auth_method` would validate against the stale value while persisting the new one. Fixed with a
  regression test that fails without the fix.

**Testing limitation, disclosed explicitly**: neither `instance_principal` nor `workload_identity` can be
integration-tested end-to-end in this environment — there is no real OCI Compute instance and no real OKE
cluster available here. All coverage for these two auth methods is mocked at the SDK boundary
(`oci.auth.signers.InstancePrincipalsSecurityTokenSigner` /
`oci.auth.signers.get_oke_workload_identity_resource_principal_signer` are mocked; the real OCI instance
metadata service / OKE workload-identity proxymux is never called). This mirrors the existing `api_key`
testing limitation noted above (no live calls to `*.oraclecloud.com`), extended to cover signer construction
too. Manual verification against a real OCI Compute instance and a real OKE cluster with Workload Identity
enabled hasn't been done and couldn't be within this PR's constraints — recommended before either auth
method ships to users relying on it in production.

## Summary by CodeRabbit

* **New Features**
  * Added **OCI** as an embedding-only provider, with onboarding and settings support for region, auth method, and credentials (including `api_key`, `instance_principal`, `workload_identity`).
  * Enabled OCI embedding model discovery and a dedicated OCI model-listing API path.
* **Bug Fixes**
  * Improved OCI provider health handling and ensured health results refresh when OCI embedding credentials change.
  * Hardened OCI credential validation and safer OCI removal behavior with fallback to other configured providers.
* **Tests**
  * Added comprehensive unit coverage for OCI endpoints, validation, settings read/write, and embedding request credential/kwargs behavior.


galovics:
Re CodeRabbit's `os.environ` path-instruction finding on `config_manager.py`: this matches the exact pre-existing pattern used by every other provider (watsonx, openai, ollama) in this same file, so I'm leaving it as-is rather than fixing it only for OCI (inconsistent scope creep otherwise). Happy to take this on as a separate cross-cutting refactor PR if maintainers want it addressed. Same call made on the identical finding in #2064.

galovics:
Hey @edwinjosechittilappilly @mfortman11 @lucaseduoli @phact — bumping this one too, ~4 weeks out. This is rebased onto current `main` and CI is green. Happy to address anything else that comes up in review. Would appreciate a look when you get a chance.

galovics:
Bumping this one too — ~3 weeks out. Re-rebased onto `main` post-#2321, including a fix for the same `input_type` regression as #2064 plus restoring OCI's signer-based auth (instance_principal/workload_identity) that the retrieval rewrite had silently dropped. CI green, ready for review whenever you get a chance.

galovics:
@edwinjosechittilappilly @mfortman11 @lucaseduoli @phact - friendly nudge, ~2 weeks since the last ping. Rebased again onto current `main` (no conflicts, CI green). Since the last bump: Cohere's `input_type` and OCI signer auth (instance_principal/workload_identity) now also reach Langflow's own embedding calls, not just the backend call sites. Happy to address anything in review.

PR #2066, opened 2026-07-09
fix: missing ingest settings for individual files in COS

edwinjosechittilappilly:
FileBrowserDialog (the "Browse Files" individual-file picker) never read the "Make documents available to all users" toggle or ingest settings — it built its sync request with only selected_files, so shared silently defaulted to false on the backend. The whole-bucket ("folder") ingest path in SharedBucketView already did this correctly.

Changes (frontend-only, no backend changes needed — it already threads shared uniformly):
- frontend/components/file-browser-dialog.tsx: added showShared/ingestSettings props, and now includes settings and shared in the sync request body, mirroring the bucket-ingest path.
- frontend/components/connectors/shared-bucket-view.tsx: passes its existing showSharedToggle/ingestSettings state down into FileBrowserDialog.

### DIFF
--- a/frontend/components/file-browser-dialog.tsx
+++ b/frontend/components/file-browser-dialog.tsx
@@ -8,6 +8,7 @@ import {
   type RemoteFile,
   useBrowseConnectionFiles,
 } from "@/app/api/queries/useBrowseConnectionFiles";
+import type { IngestSettings } from "@/components/cloud-picker/types";
 import { formatFileSize } from "@/lib/file-format";
 import { Badge } from "./ui/badge";
 import { Button } from "./ui/button";
@@ -28,6 +29,10 @@ interface FileBrowserDialogProps {
   connectorType: string;
   connectionId: string;
   buckets?: string[];
+  /** When true, apply the "Make documents available to all users" setting from `ingestSettings`. */
+  showShared?: boolean;
+  /** Ingest settings (embedding model, chunking, OCR, shared toggle) configured on the parent screen. */
+  ingestSettings?: IngestSettings;
 }
 
 export function FileBrowserDialog({
@@ -36,6 +41,8 @@ export function FileBrowserDialog({
   connectorType,
   connectionId,
   buckets,
+  showShared,
+  ingestSettings,
 }: FileBrowserDialogProps) {
   const [search, setSearch] = useState("");
   const [selectedBucket, setSelectedBucket] = useState<string | undefined>(
@@ -136,6 +143,8 @@ export function FileBrowserDialog({
             mimeType: "",
             size: f.size,
           })),
+          settings: ingestSettings,
+          shared: showShared ? (ingestSettings?.shared ?? false) : undefined,
           ...(hasStale ? { replace_duplicates: true } : {}),
         },
       });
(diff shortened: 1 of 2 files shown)


frontend/components/file-browser-dialog.tsx
mpawlow, line 128: ### (1a) [Normal] Individual-file ingest path skips chunk-settings client-side validation that the bucket-level path enforces

#### Problem

- `SharedBucketView.ingestSelected` (the whole-bucket path, `frontend/components/connectors/shared-bucket-view.tsx:107-111`) validates `ingestSettings` via `getIngestChunkSettingsError` before submitting, blocking with a toast error if `chunkSize < 1` or `chunkOverlap >= chunkSize`. `FileBrowserDialog.handleIngest` (`frontend/components/file-browser-dialog.tsx:128-150`), which now forwards the exact same `ingestSettings` object to the backend via `settings: ingestSettings`, does **not** perform this same validation before calling `syncMutation.mutateAsync`.

#### Background Information

- The `IngestSettings` panel (chunk size / chunk overlap `NumberInput` fields, `frontend/components/cloud-picker/ingest-settings.tsx:206-226`) has no `min`/`max` enforcement, so a user can freely set `chunkOverlap >= chunkSize` in the UI.
- That same settings panel and its session state (`ingestSettings`) are shared between both ingest entry points in `SharedBucketView` — a user can set an invalid chunk overlap, then click "Browse Files" (which opens `FileBrowserDialog`) instead of the whole-bucket "Ingest N Buckets" button, and submit without ever hitting `getIngestChunkSettingsError`.
- Backend search (`src/connectors/service.py`, `src/api/connectors.py`) shows no server-side re-validation of the `chunkSize`/`chunkOverlap` relationship, so an invalid submission would proceed to task creation and likely surface as a confusing downstream ingestion failure rather than the clear, immediate toast the bucket path provides.
- This gap pre-dates the "shared" fix but is newly *exposed* by this PR, since before this change `FileBrowserDialog` didn't forward `settings` at all — it had no reason to validate. Now that it forwards the same object, the two ingest paths have diverged in validation behavior.

#### Code References

- `frontend/components/file-browser-dialog.tsx:128-150` (missing validation before `syncMutation.mutateAsync`)
- `frontend/components/connectors/shared-bucket-view.tsx:107-111` (existing validation pattern to mirror)
- `frontend/components/cloud-picker/types.ts:119-129` (`getIngestChunkSettingsError`)

#### Potential Solution

- In `handleIngest`, call `getIngestChunkSettingsError(ingestSettings)` (guarding for `ingestSettings` being `undefined`) before `syncMutation.mutateAsync`, and short-circuit with `toast.error("Could not start ingest", { description: chunkErr })` on failure — identical to `ingestSelected`'s existing pattern.

#### Alternative Solutions

- Extract the validate-then-toast logic into a small shared helper (e.g. `validateIngestSettingsOrToast(settings)`) used by both `ingestSelected` and `handleIngest`, to prevent this same divergence from recurring as more ingest entry points are added. This is the more idiomatic fix given there are now two call sites duplicating the same submit-guard pattern.

mpawlow (approved):
## Code Review 1

- ✅  Approved / LGTM  🚀 
- See non-blocking PR comment (1a) for consideration

## Functional Review 1

- Not performed yet (I don't have a MinIO instance in HMAC mode setup locally)
   - I can set this up tomorrow OR we can manually verify the fix (using two separate users) in a shared OpenRAG instance with IBM COS already setup)

frontend/components/file-browser-dialog.tsx
edwinjosechittilappilly: @coderabbitai create an issue with the above suggestions. also a potential PR also.

PR #2067, opened 2026-07-09
fix: add warning logs for 10k aggregation limit

yogichipalkatti:
Temporarily fixes Issue #1548

Affected functions:

get_synced_file_ids_for_connector(): Warns when connector_file_id,
get_synced_id_to_filename_map(): Warns when document_id aggregation
reaches the limit

unique values per field

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of large connector result sets so file and name lookups are less likely to return incomplete data.
  * Added warnings when result counts reach a limit, helping surface cases where data may be truncated.

### DIFF
--- a/src/api/connectors.py
+++ b/src/api/connectors.py
@@ -33,6 +33,12 @@ from utils.telemetry import Category, MessageId, TelemetryClient
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


Wallgau (approved):
Great work!

PR #2069, opened 2026-07-09
fix: reset dialog state and abort in-flight test on cancel (#2048)

### LINKED ISSUE #2048: [Bug]: Azure Blog Storage Connector: "Azure Blob Storage Setup" dialog caches previous field state and error messages on cancel
### OpenRAG Version

0.6.2

### Deployment Method

Local development (make dev)

### Operating System

Ubuntu 24.04.4 LTS

### Python Version

Python 3.13.14

### Affected Area

Connectors (Google Drive, OneDrive, SharePoint)

### Bug Description

"Azure Blob Storage Setup" dialog caches previous field state and error messages on cancel

### Steps to Reproduce

- Shutdown Azurite
   - `make azurite-down`
- Go to **Settings** page
- Click **Connect** button on **Azure Blob Storage** connector. 
- Select **Connection String** Auth
- Specify a valid **Connection String**
- Click **Save** button (without clicking **Test Connection & List Containers** button)
   - "Test the connection first to validate credentials." message is shown
- Click **Cancel** button
- Click **Connect** button on **Azure Blob Storage** connector. 
   - BUG: **Connection String** is still persisted
   - BUG" "Test the connection first to validate credentials." message is still shown
- Click **Test Connection & List Containers** button
   - Connection appears to hang (rightfully since Azurite is down)  
   - Note: Eventually the connection will time-out and show an error message
- Click **Cancel** button
- Click **Connect** button on **Azure Blob Storage** connector. 
   - BUG: **Connecting...** status still shown

### Expected Behavior

- "Azure Blob Storage Setup" dialog caches to open fresh in a reset state instead of caching previous field state and error messages on cancel

### Actual Behavior

- "Azure Blob Storage Setup" dialog caches previous field state and error messages on cancel

### Relevant Logs

```shell

```

### Screenshots

Image

Image

### Additional Context

- For **Azure Blog Storage** connector setup instructions, see offline PDF file (`OpenRAG - Azure Blob Storage Connector - Manual Testing Instructions.pdf`)

### Checklist

hunterxtang:
Fixes #2048 on main.

## Root cause

Connector settings dialogs (Azure Blob, S3, IBM COS) are rendered
unconditionally in `connector-cards.tsx` with a stable `key`, so the
component never unmounts when the user closes the dialog. All
`useState` and `useForm` state — the connection-string field value,
the `formError` "Test the connection first…" message, and the
`isFetchingContainers` loading flag — survive Cancel and reappear on
the next open.

Additionally, the Azure Blob test-connection handler called
`fetch("/api/connectors/azure_blob/test")` with no `AbortController`.
Clicking Cancel while a test was in flight left `isFetchingContainers`
stuck at `true` until the underlying TCP connection timed out, keeping
the button locked on "Connecting…" on reopen.

## Fix

**1 — `connector-cards.tsx`: unmount on close**

Render each `SettingsDialog` only while `openDialog === descriptor.connectorType`;
return `null` otherwise. The component unmounts on close, React destroys all
internal state, and every field and error message is fresh on the next open.
This also resolves the same latent stale-state bug in the S3 and IBM COS
dialogs at no extra cost, since they go through the same rendering loop.

**2 — `azure-blob/settings-dialog.tsx`: AbortController on the test fetch**

- A `useRef<AbortController | null>` tracks the currently in-flight test request.
- At the top of `handleTestConnection`, any prior call is aborted before a new
  controller is created — cancels a stuck request if the user clicks the button
  twice.
- `signal: controller.signal` is passed to `fetch`; `AbortError` is caught and
  swallowed silently (no error state set to the user).
- A `useEffect` cleanup aborts on unmount, so closing the dialog mid-request
  immediately cancels the network call rather than letting it run to TCP timeout.
- The `finally` block is guarded with `testAbortRef.current === controller` so
  that an aborted call's `finally` does not clear `isFetchingContainers` while a
  superseding request is still in flight (prevents spinner flicker).

## Known tradeoff

Unmount-on-close skips the Radix exit animation — the dialog disappears
immediately rather than fading out. Intentional; correctness takes priority over
the transition.

## Deferred follow-up

S3 (`s3-settings-dialog.tsx`) and IBM COS (`ibm-cos/settings-dialog.tsx`) both
use a `mutateAsync` + raw `fetch` sequence in their test paths. The raw `fetch`
portion is equally abortable, but adding `AbortController` to both is deferred
to keep this PR scoped to #2048.

## Verification

- `biome check` and `tsc --noEmit` on the two changed files: no new warnings or
  errors vs. the baseline on `main`.
- Manual reproduction of all three symptoms from #2048 against a production
  container build of this branch (Azurite down): connection string, stale error
  message, and stuck loading state all confirmed fixed after Cancel + reopen.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connector settings dialogs so they close and reset cleanly.
  * Cancelled previous connection tests when a new test starts or the dialog closes.
  * Prevented cancelled requests from showing errors or incorrectly changing loading indicators.

### DIFF
--- a/frontend/app/settings/_components/connector-cards.tsx
+++ b/frontend/app/settings/_components/connector-cards.tsx
@@ -126,11 +126,19 @@ export default function ConnectorCards() {
       </div>
 
       {dialogDescriptors.map((descriptor) => {
+        // Render only while open so the component unmounts on close, which
+        // resets all useState/useForm state automatically and eliminates
+        // stale field values, error messages, and stuck loading indicators
+        // on reopen. The exit animation is foregone — an acceptable tradeoff
+        // for correctness.
+        if (openDialog !== descriptor.connectorType) return null;
         const Dialog = descriptor.SettingsDialog!;
         return (
           <Dialog
             key={descriptor.connectorType}
-            open={openDialog === descriptor.connectorType}
+            // mounted only while open; open is always true here,
+            // unmount handles "close"
+            open={true}
             setOpen={(open: boolean) =>
               setOpenDialog(open ? descriptor.connectorType : null)
             }
--- a/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
+++ b/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useQueryClient } from "@tanstack/react-query";
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { FormProvider, useForm } from "react-hook-form";
 import { toast } from "sonner";
 import { Button } from "@/components/ui/button";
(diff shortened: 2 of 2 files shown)


hunterxtang:
Superseded by #2072 — re-landed on an in-repo branch so the secret-dependent checks (build-images, e2e) can run.

PR #2071, opened 2026-07-09
fix: restrict ingestion to containers section shows up correctly and error message isn't duplicated

### LINKED ISSUE #2047: [Bug]: Azure Blob Storage Connector: "Restrict Ingestion to Containers" section is not cleared when toggling between "Connection String" and "Account Key" Auth
### OpenRAG Version

0.6.2

### Deployment Method

Local development (make dev)

### Operating System

Ubuntu 24.04.4 LTS

### Python Version

Python 3.13.14

### Affected Area

Connectors (Google Drive, OneDrive, SharePoint)

### Bug Description

"Restrict Ingestion to Containers" section is not cleared when toggling between "Connection String" and "Account Key" Auth

### Steps to Reproduce

- Go to **Settings** page
- Click **Connect** button on **Azure Blob Storage** connector. 
- Select **Connection String** Auth
- Specify a valid **Connection String**
- Click **Refresh Containers**
- Click **Account Key** Auth
   - BUG: "Restrict Ingestion to Containers" section is not cleared 
- Click **Refresh Containers** (without specifying any values for the **Account Key** fields)
   - BUG: "Restrict Ingestion to Containers" section is not cleared 
- Click **Save** button
   - BUG: Duplicate errors are shown
   - BUG: "Restrict Ingestion to Containers" section is not cleared 

### Expected Behavior

- "Restrict Ingestion to Containers" section should be cleared (reset back to default)
- Duplicate error messages should not be shown (should be consolidated into one site)

### Actual Behavior

- "Restrict Ingestion to Containers" section is not cleared when toggling between "Connection String" and "Account Key" Auth

### Relevant Logs

```shell

```

### Screenshots

Image

Image

Image

### Additional Context

- For **Azure Blog Storage** connector setup instructions, see offline PDF file (`OpenRAG - Azure Blob Storage Connector - Manual Testing Instructions.pdf`)

### Checklist

Vchen7629:
## Summary
Fixes #2047 .This PR adds a showContainers state to control when the "Restrict Ingestion to Containers" section shows up and onAuthModeChange to reset it when the user switches authentication methods, preventing stale container data from persisting across auth mode changes. The section also now hides on a failed test connection, so bad credentials clear any previously visible list. Additionally, setContainersError(null) is called at the start of onSubmit to prevent duplicate error messages when a stale connection error and a save validation error would otherwise both appear simultaneously.

## Before
[screen-capture (11).webm](https://github.com/user-attachments/assets/436bd365-23c9-466c-bbc2-1f534183e6c1)

## After
[screen-capture (13).webm](https://github.com/user-attachments/assets/b4a26979-e1fd-4c46-a2e5-1360839e75a7)

## Changes
- `settings-dialog.tsx`: adds showContainers state, sets it on test success/failure and auth mode change, clears containersError on submit
- `settings-form.tsx`: accepts showContainers + onAuthModeChange props, gates the container section on both, resets on tab switch

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the Azure Blob setup flow so the container selector only appears when containers are available.
  * After a successful “test connection,” container options are fetched, filtered, and displayed automatically.
  * When switching authentication methods, the container selector is hidden and any container selection and related error messages are cleared.
  * If container loading fails, the correct error is shown while keeping the container selector hidden.
  * Saving now clears container-related error state as part of the submission flow.

### DIFF
--- a/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
+++ b/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
@@ -49,10 +49,12 @@ export default function AzureBlobSettingsDialog({
   const [selectedContainers, setSelectedContainers] = useState<string[]>(
     defaults?.container_names ?? [],
   );
+  const [showContainers, setShowContainers] = useState(false);
   useEffect(() => {
     if (defaults?.container_names?.length) {
       setContainers(defaults.container_names);
       setSelectedContainers(defaults.container_names);
+      setShowContainers(true);
     }
   }, [defaults]);
 
@@ -107,6 +109,7 @@ export default function AzureBlobSettingsDialog({
       const fetched: string[] = json.containers;
       setContainers(fetched);
       setSelectedContainers((prev) => prev.filter((c) => fetched.includes(c)));
+      setShowContainers(true);
     } catch (err: unknown) {
       // Ignore cancellations — they are intentional (dialog closed or new test
       // started) and must not surface an error message to the user.
@@ -114,6 +117,7 @@ export default function AzureBlobSettingsDialog({
       setContainersError(
         err instanceof Error ? err.message : "Connection failed",
       );
+      setShowContainers(false);
     } finally {
       // Only clear the loading flag if this invocation is still the active one.
       // If the user started a second test while this one was in flight, the
@@ -127,6 +131,7 @@ export default function AzureBlobSettingsDialog({
 
   const onSubmit = handleSubmit(async (data) => {
     setFormError(null);
+    setContainersError(null);
     if (containers === null) {
       setFormError("Test the connection first to validate credentials.");
       return;
@@ -197,6 +202,8 @@ export default function AzureBlobSettingsDialog({
               connectionStringSet={defaults?.connection_string_set}
               accountKeySet={defaults?.account_key_set}
               formError={formError}
+              showContainers={showContainers}
+              onAuthModeChange={() => setShowContainers(false)}
             />
 
             <DialogFooter className="mt-4">
--- a/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
+++ b/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
@@ -56,11 +56,10 @@ export default function AzureBlobSettingsDialog({
   const [prevDefaults, setPrevDefaults] = useState(defaults);
   if (defaults !== prevDefaults) {
     setPrevDefaults(defaults);
-    if (defaults?.container_names?.length) {
-      setContainers(defaults.container_names);
-      setSelectedContainers(defaults.container_names);
-      setShowContainers(true);
-    }
+    const names = defaults?.container_names ?? [];
+    setContainers(names.length ? names : null);
+    setSelectedContainers(names);
+    setShowContainers(names.length > 0);
   }
 
   const [isFetchingContainers, setIsFetchingContainers] = useState(false);
--- a/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
+++ b/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
@@ -72,13 +72,16 @@ export default function AzureBlobSettingsDialog({
   // can be cancelled if the user closes the dialog or starts a new test.
   const testAbortRef = useRef<AbortController | null>(null);
 
-  // Abort any in-flight test request when the component unmounts (i.e. when
-  // the dialog is closed, since connector-cards renders it only while open).
+  // Abort any in-flight test request when the dialog closes. The dialog
+  // component itself stays mounted (connector-cards renders it unconditionally
+  // and just toggles `open`), so an unmount-only cleanup would never fire here
+  // — without this, a canceled test could still resolve and repopulate the
+  // container list after the user closed the dialog.
   useEffect(() => {
-    return () => {
+    if (!open) {
       testAbortRef.current?.abort();
-    };
-  }, []);
+    }
+  }, [open]);
 
   const handleTestConnection = handleSubmit(async (data) => {
     // Cancel any previous in-flight test before starting a new one.
--- a/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
+++ b/frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
@@ -203,7 +203,10 @@ export default function AzureBlobSettingsDialog({
               accountKeySet={defaults?.account_key_set}
               formError={formError}
               showContainers={showContainers}
-              onAuthModeChange={() => setShowContainers(false)}
+              onAuthModeChange={() => {
+                setShowContainers(false);
+                setContainers(null);
+              }}
             />
 
             <DialogFooter className="mt-4">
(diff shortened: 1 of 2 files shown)


richardz403 (commented):
LGTM

Ran PR locally, the fix does solve the issue of multiple error messages being displayed, removing the excess error messages. The application itself also runs smoothly with no other bugs found.

Ran tests locally, all tests passed, no apparent issues.

frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
```diff
@@ -89,17 +96,20 @@ export default function AzureBlobSettingsDialog({
       const fetched: string[] = json.containers;
       setContainers(fetched);
       setSelectedContainers((prev) => prev.filter((c) => fetched.includes(c)));
+      setShowContainers(true);
     } catch (err: unknown) {
       setContainersError(
         err instanceof Error ? err.message : "Connection failed",
       );
+      setShowContainers(false);
```
mpawlow, line 127: ### (1a) [Major] Failed re-test leaves `containers` non-null, letting Save bypass the "test first" guard

#### Problem

- `handleTestConnection`'s `catch` block (`settings-dialog.tsx:100-107`) sets `containersError` and `setShowContainers(false)` on a failed test, but never resets `containers` back to `null`.
- `onSubmit`'s only safety check is `if (containers === null)` (`settings-dialog.tsx:113`). If a prior test succeeded, `containers` stays populated even after a subsequent failed re-test — so the guard silently passes.
- The failure path is reachable without switching auth tabs at all: test succeeds under mode A → user edits the same-tab credential fields to bad/different values → clicks "Refresh Containers" → test fails → user clicks **Save** anyway. The section is hidden (`showContainers=false`) but the stale `containers`/`selectedContainers` from the earlier successful test are still submitted.
- This contradicts the PR's own description ("bad credentials clear any previously visible list") — the *display* is cleared, but the underlying state is not.

#### Background Information

- `azure_blob_configure` re-validates that the *connection* works (`_list_container_names`) but never checks that the submitted `container_names` actually exist for those credentials — it persists `body.container_names` verbatim. So if the current (possibly different-account) credentials happen to connect successfully, a stale container-name allowlist from an earlier, unrelated test can be persisted without any server-side cross-check.
- This is the same defect class CodeRabbit flagged (🟠 Major) for `onAuthModeChange` in an earlier revision of this PR (review thread now resolved/outdated — see the CodeRabbit section below). That specific tab-switch path is fixed in the current diff (`settings-dialog.tsx:184-189` now resets `containers`, `containersError`, and `formError`, not just `showContainers`), but the parallel same-tab "retest fails" path through `handleTestConnection`'s catch block was not.

#### Code References

- `frontend/enhancements/connectors/azure-blob/settings-dialog.tsx:100-107` (catch block missing `setContainers(null)`)
- `frontend/enhancements/connectors/azure-blob/settings-dialog.tsx:110-116` (`onSubmit` guard relying solely on `containers === null`)
- `enhancements/connectors/azure_blob/api.py:104-152` (`azure_blob_configure` — connectivity-only validation)
- `enhancements/connectors/azure_blob/support.py:61-64` (`container_names` persisted from request body without cross-checking against the live container list)

#### Potential Solution

- In the `catch` block of `handleTestConnection`, also call `setContainers(null)`, mirroring what `onAuthModeChange` already does:
  ```diff
   } catch (err: unknown) {
     setContainersError(
       err instanceof Error ? err.message : "Connection failed",
     );
     setShowContainers(false);
  +  setContainers(null);
   } finally {
  ```

#### Alternative Solutions

- Defense-in-depth: have `azure_blob_configure` validate that every submitted `container_names` entry is present in the freshly-listed containers for the resolved credentials, rejecting the save (400) otherwise. This closes the gap regardless of which frontend state-reset path is missed, and is consistent with the rest of the flow already performing a live connectivity check at save time.

frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
```diff
@@ -49,12 +49,19 @@ export default function AzureBlobSettingsDialog({
   const [selectedContainers, setSelectedContainers] = useState<string[]>(
     defaults?.container_names ?? [],
   );
-  useEffect(() => {
+  const [showContainers, setShowContainers] = useState(
+    !!defaults?.container_names?.length,
+  );
+
+  const [prevDefaults, setPrevDefaults] = useState(defaults);
```
mpawlow, line 56: ### (1b) [Normal] Locally-tested-but-unsaved container state survives dialog close/reopen

#### Problem

- The `settings-dialog.tsx:56-64` derived-state block only *sets* `containers` / `selectedContainers` / `showContainers` when the newly-fetched `defaults.container_names` is non-empty; it never clears them when defaults transitions to empty.
- `AzureBlobSettingsDialog` is permanently mounted by its parent with a stable `key` (`frontend/app/settings/_components/connector-cards.tsx:128-138`) — closing the dialog does not unmount it, so component state (`containers`, `showContainers`, `selectedContainers`) persists across opens.
- `useAzureBlobDefaultsQuery` uses `staleTime: 0` and `enabled: open`, so every reopen triggers a fresh fetch and a new object reference for `defaults`, but if the *saved* config still has no container restriction, the falsy branch is skipped and stale local state is never cleared.
- Repro: open dialog on an unconfigured connection → enter credentials → Test (succeeds, shows containers) → Cancel without saving → reopen the dialog. The previously-tested container list reappears even though nothing was ever persisted, misrepresenting the actual saved state and again allowing Save to bypass the "test first" guard without any test having occurred in the new session.

#### Code References

- `frontend/enhancements/connectors/azure-blob/settings-dialog.tsx:56-64`
- `frontend/app/settings/_components/connector-cards.tsx:128-138` (dialog stays mounted across open/close)
- `frontend/enhancements/connectors/azure-blob/useAzureBlobDefaultsQuery.ts:19-26` (`staleTime: 0`, `enabled: open`)

#### Potential Solution

- In the derived-state block, handle both directions explicitly instead of only the truthy case:
  ```diff
   if (defaults !== prevDefaults) {
     setPrevDefaults(defaults);
  -  if (defaults?.container_names?.length) {
  -    setContainers(defaults.container_names);
  -    setSelectedContainers(defaults.container_names);
  -    setShowContainers(true);
  -  }
  +  const names = defaults?.container_names ?? [];
  +  setContainers(names.length ? names : null);
  +  setSelectedContainers(names);
  +  setShowContainers(names.length > 0);
   }
  ```

#### Alternative Solutions

- Reset all container-related state in `setOpen(false)` (Cancel/close handler) instead of relying on the defaults-diff to self-correct — simpler to reason about, at the cost of losing in-progress (untested) form edits when the dialog is reopened.

mpawlow (changes requested):
## Code Review 1

- See PR comments: (1a) (1b)

## Functional Review 1

- ✅ Successfully verified scenarios outlined in issue's expected behavior
- ⚠️ Additional edge case scenarios identified by (1a) (1b)

frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
mpawlow: ✅  Verified

frontend/enhancements/connectors/azure-blob/settings-dialog.tsx
mpawlow: ✅  Verified

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM  🚀 
- ℹ️ Found a minor nit for an extreme edge case
   - Abort-on-unmount cleanup never fires on dialog close, so a canceled test can still land and repopulate stale containers
   - Addressed with a very small change / commit (since it was trivial)
- ✅ Resolved conflict & rebased branch off of the latest from main

## Functional Review 2

- ✅ Successfully verified all test scenarios

PR #2077, opened 2026-07-13
fix: introduces a user-facing duplicate handling workflow for bucket-based connectors 

edwinjosechittilappilly:
This pull request introduces a user-facing duplicate handling workflow for bucket-based connectors (such as S3, Azure Blob, IBM COS), ensuring that users are prompted to confirm before overwriting already-indexed files during ingestion. It also improves backend duplicate detection for these connectors, adds robust error handling for legacy indices, and unifies the overwrite logic for file selection. The most important changes are summarized below:

### Frontend: Duplicate Handling Workflow

* Added a duplicate detection step to `SharedBucketView` before ingesting selected buckets. If duplicates are found, a dialog prompts the user to choose whether to overwrite them or skip. The ingestion process is updated to respect this choice, and the UI reflects the checking and syncing states. (`frontend/components/connectors/shared-bucket-view.tsx`) [[1]](diffhunk://#diff-4c072ea9c27473c224111c6e765179c580aaec73b3c1498bc493a3cb50d5c6c4R11-R32) [[2]](diffhunk://#diff-4c072ea9c27473c224111c6e765179c580aaec73b3c1498bc493a3cb50d5c6c4R89-R97) [[3]](diffhunk://#diff-4c072ea9c27473c224111c6e765179c580aaec73b3c1498bc493a3cb50d5c6c4L108-R169) [[4]](diffhunk://#diff-4c072ea9c27473c224111c6e765179c580aaec73b3c1498bc493a3cb50d5c6c4R183-R303) [[5]](diffhunk://#diff-4c072ea9c27473c224111c6e765179c580aaec73b3c1498bc493a3cb50d5c6c4L332-R479) [[6]](diffhunk://#diff-4c072ea9c27473c224111c6e765179c580aaec73b3c1498bc493a3cb50d5c6c4R500-R508)
* Updated `FileBrowserDialog` so that already-ingested files are always selectable, and selecting any already-ingested file triggers the overwrite logic (`replace_duplicates`). This ensures consistent behavior for both new and existing files. (`frontend/components/file-browser-dialog.tsx`) [[1]](diffhunk://#diff-a731c4f2da839b9e51cc35d28717a4f4c321d561504e73728a4fbb65e73bf019L81-R83) [[2]](diffhunk://#diff-a731c4f2da839b9e51cc35d28717a4f4c321d561504e73728a4fbb65e73bf019L131-R131) [[3]](diffhunk://#diff-a731c4f2da839b9e51cc35d28717a4f4c321d561504e73728a4fbb65e73bf019L148-R145) [[4]](diffhunk://#diff-a731c4f2da839b9e51cc35d28717a4f4c321d561504e73728a4fbb65e73bf019L314-R315)

### Backend: Duplicate Detection for Buckets

* Implemented `_classify_bucket_connector_duplicates`, which previews a bucket-based sync by classifying remote blobs as new, changed, or unchanged, and returns this information to the frontend for user confirmation. (`src/api/connectors.py`) [[1]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680R742-R818) [[2]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680R875-R886)
* Extended the `/check-duplicates` API to support a `bucket_filter` parameter, enabling duplicate checking for entire buckets, not just individual files. (`src/api/connectors.py`) [[1]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680R603-R607) [[2]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680L720-R833)

### Backend: Robustness and Compatibility

* Improved error handling for OpenSearch aggregations by detecting and recovering from missing keyword mappings on legacy indices. This ensures duplicate detection works even if the index schema is outdated. (`src/api/connectors.py`) [[1]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680R72-R80) [[2]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680R117-R128) [[3]](diffhunk://#diff-e072fa320915014cf46bdf01548a1bd72bd0e1953e556c08fa0849e5fb205680R259-R268)

### Miscellaneous

* Fixed a bug in `process_connector_document` by renaming `document_id` to `connector_file_id` in the document processing call. (`src/connectors/service.py`)

### DIFF
--- a/frontend/components/connectors/shared-bucket-view.tsx
+++ b/frontend/components/connectors/shared-bucket-view.tsx
@@ -350,6 +350,8 @@ export function SharedBucketView({
           connectorType={connector.type}
           connectionId={connector.connectionId}
           buckets={[browseDialogBucket]}
+          showShared={showSharedToggle}
+          ingestSettings={ingestSettings}
         />
       )}
     </>
--- a/frontend/components/connectors/shared-bucket-view.tsx
+++ b/frontend/components/connectors/shared-bucket-view.tsx
@@ -8,11 +8,26 @@ import type { useSyncConnector } from "@/app/api/mutations/useSyncConnector";
 import { useGetSettingsQuery } from "@/app/api/queries/useGetSettingsQuery";
 import { IngestSettings } from "@/components/cloud-picker/ingest-settings";
 import { getIngestChunkSettingsError } from "@/components/cloud-picker/types";
+import { DuplicateHandlingDialog } from "@/components/duplicate-handling-dialog";
 import { FileBrowserDialog } from "@/components/file-browser-dialog";
 import { Button } from "@/components/ui/button";
 import { useAuth } from "@/contexts/auth-context";
 import { useSessionIngestSettings } from "@/hooks/useSessionIngestSettings";
 import { trackProcessFailure, trackStartProcess } from "@/lib/analytics";
+
+interface BucketDuplicateFile {
+  id: string;
+  name: string;
+  mimeType: string;
+  downloadUrl?: string;
+  size?: number;
+}
+
+interface BucketDuplicateCheckResponse {
+  duplicate_names?: string[];
+  duplicate_count?: number;
+  non_duplicate_files?: BucketDuplicateFile[];
+}
 
 export interface SharedBucketViewProps {
   connector: any;
@@ -70,6 +85,14 @@ export function SharedBucketView({
   const [browseDialogBucket, setBrowseDialogBucket] = useState<string | null>(
     null,
   );
+  const [isCheckingDuplicates, setIsCheckingDuplicates] = useState(false);
+  const [duplicateDialogOpen, setDuplicateDialogOpen] = useState(false);
+  const [pendingDuplicates, setPendingDuplicates] = useState<{
+    duplicateNames: string[];
+    duplicateCount: number;
+    nonDuplicateFiles: BucketDuplicateFile[];
+  } | null>(null);
+  const isOverwriteConfirmedRef = useRef(false);
 
   useEffect(() => {
     if (
@@ -105,20 +128,43 @@ export function SharedBucketView({
     });
   };
 
-  const ingestSelected = () => {
-    const chunkErr = getIngestChunkSettingsError(ingestSettings);
-    if (chunkErr) {
-      toast.error("Could not start ingest", { description: chunkErr });
-      return;
-    }
-    trackStartProcess({
+  const onSyncSettled = () => {
+    invalidate();
+  };
+
+  const onSyncError = (err: unknown) => {
+    trackProcessFailure({
       processType: "Ingestion",
       process: "Document Upload",
       category: "Knowledge",
       source: "connector",
       connector_type: connector.type,
-      total_buckets: selectedBuckets.size,
+      resultValue: err instanceof Error ? err.message : "Sync failed",
     });
+    toast.error(err instanceof Error ? err.message : "Sync failed");
+  };
+
+  const onSyncSuccess = (result: { task_ids?: string[]; message?: string }) => {
+    onSyncSettled();
+    if (result.task_ids?.length) {
+      // The container path may return two tasks (new files + changed files);
+      // track them all.
+      for (const id of result.task_ids) {
+        addTask(id, {
+          connectorType: connector.type,
+          source: "connector",
+        });
+      }
+      onDone();
+    } else {
+      toast.info(result.message ?? "No files found in the selected buckets.");
+    }
+  };
+
+  // Full bucket sync: backend classifies new/changed/unchanged itself and
+  // re-ingests both new and changed blobs (used both when there are no
+  // duplicates and when the user chooses to overwrite them).
+  const runBucketSync = () => {
     syncMutation.mutate(
       {
         connectorType: connector.type,
@@ -132,38 +178,118 @@ export function SharedBucketView({
             : undefined,
         },
       },
+      { onSuccess: onSyncSuccess, onError: onSyncError },
+    );
+  };
+
+  // Sync only the given (new, not-yet-ingested) files — used when the user
+  // skips overwriting duplicates.
+  const runSelectedFilesSync = (files: BucketDuplicateFile[]) => {
+    syncMutation.mutate(
       {
-        onSuccess: (result) => {
-          invalidate();
-          if (result.task_ids?.length) {
-            // The container path may return two tasks (new files + changed files);
-            // track them all.
-            for (const id of result.task_ids) {
-              addTask(id, {
-                connectorType: connector.type,
-                source: "connector",
-              });
-            }
-            onDone();
-          } else {
-            toast.info(
-              result.message ?? "No files found in the selected buckets.",
-            );
-          }
-        },
-        onError: (err) => {
-          trackProcessFailure({
-            processType: "Ingestion",
-            process: "Document Upload",
-            category: "Knowledge",
-            source: "connector",
-            connector_type: connector.type,
-            resultValue: err instanceof Error ? err.message : "Sync failed",
-          });
-          toast.error(err instanceof Error ? err.message : "Sync failed");
+        connectorType: connector.type,
+        body: {
+          connection_id: connector.connectionId!,
+          selected_files: files,
+          settings: ingestSettings,
+          shared: showSharedToggle
+            ? (ingestSettings.shared ?? false)
+            : undefined,
         },
       },
+      { onSuccess: onSyncSuccess, onError: onSyncError },
     );
+  };
+
+  const ingestSelected = async () => {
+    const chunkErr = getIngestChunkSettingsError(ingestSettings);
+    if (chunkErr) {
+      toast.error("Could not start ingest", { description: chunkErr });
+      return;
+    }
+    trackStartProcess({
+      processType: "Ingestion",
+      process: "Document Upload",
+      category: "Knowledge",
+      source: "connector",
+      connector_type: connector.type,
+      total_buckets: selectedBuckets.size,
+    });
+
+    setIsCheckingDuplicates(true);
+    try {
+      const checkResponse = await fetch(
+        `/api/connectors/${connector.type}/check-duplicates`,
+        {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({
+            connection_id: connector.connectionId,
+            bucket_filter: Array.from(selectedBuckets),
+          }),
+        },
+      );
+
+      if (!checkResponse.ok) {
+        throw new Error(`Duplicate check failed: ${checkResponse.statusText}`);
+      }
+
+      const checkData =
+        (await checkResponse.json()) as BucketDuplicateCheckResponse;
+      const duplicateNames = checkData.duplicate_names || [];
+      const duplicateCount =
+        typeof checkData.duplicate_count === "number"
+          ? checkData.duplicate_count
+          : duplicateNames.length;
+
+      if (duplicateCount === 0) {
+        runBucketSync();
+        return;
+      }
+
+      setPendingDuplicates({
+        duplicateNames,
+        duplicateCount,
+        nonDuplicateFiles: checkData.non_duplicate_files || [],
+      });
+      setDuplicateDialogOpen(true);
+    } catch (err) {
+      console.error("[Bucket Sync] Duplicate check failed:", err);
+      // Fallback: proceed with the normal full sync (backend still handles
+      // new/changed reconciliation on its own).
+      runBucketSync();
+    } finally {
+      setIsCheckingDuplicates(false);
+    }
+  };
+
+  const handleOverwriteDuplicates = () => {
+    if (!pendingDuplicates) return;
+    isOverwriteConfirmedRef.current = true;
+    runBucketSync();
+    setPendingDuplicates(null);
+  };
+
+  const handleDuplicateDialogOpenChange = (open: boolean) => {
+    if (!open && pendingDuplicates) {
+      if (isOverwriteConfirmedRef.current) {
+        // Overwrite already submitted in handleOverwriteDuplicates; this close
+        // event fires immediately after and would otherwise re-enter the
+        // "skip duplicates" branch.
+        isOverwriteConfirmedRef.current = false;
+      } else {
+        const { nonDuplicateFiles, duplicateCount } = pendingDuplicates;
+        if (nonDuplicateFiles.length > 0) {
+          runSelectedFilesSync(nonDuplicateFiles);
+        } else {
+          toast.info(
+            `All ${duplicateCount} changed file(s) were skipped. Nothing was synced.`,
+          );
+        }
+      }
+      setPendingDuplicates(null);
+    }
+    setDuplicateDialogOpen(open);
   };
 
   return (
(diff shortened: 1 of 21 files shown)


mpawlow:
## Functional Review 1

- ⚠️ One UX question - **Scenario 1** 
   - See: https://github.com/langflow-ai/openrag/pull/2077#issuecomment-4962354993
- ⚠️ One Major issue discovered - **Scenario 2**
   - See: https://github.com/langflow-ai/openrag/pull/2077#issuecomment-4962397694

mpawlow:
### [UX Question] Scenario 1 

- (1) Ingest all files from Azure container
- (2) Update a subset of the files in the Azure container already ingested

image

- (3) Select entire Azure container to ingest

image

- Duplicates Dialog is shown for all 3 files 

image

- `Question`: Should detected duplicates be shown only for the file that does NOT have an update available (and the files with updates should be automatically overridden)?
     - My opinion is current implemented behavior (introduced in this PR) is correct (i.e. all 3 files should be presented in overwrite dialog)

mpawlow:
### [Major] Scenario 2

- (1) Ingest 1 file from Azure Blob Storage container (that was already ingested)

image

- BUG: Document is ingested without interruption or showing the Detect Duplicates confirmation dialog

image

frontend/components/file-browser-dialog.tsx
```diff
@@ -145,7 +142,7 @@ export function FileBrowserDialog({
           })),
           settings: ingestSettings,
           shared: showShared ? (ingestSettings?.shared ?? false) : undefined,
-          ...(hasStale ? { replace_duplicates: true } : {}),
+          ...(hasAlreadyIngested ? { replace_duplicates: true } : {}),
```
mpawlow: ### (1a) [Major] `FileBrowserDialog` allows silent, unconfirmed overwrite of already-ingested files

#### Problem

- `FileBrowserDialog` used to lock already-ingested-and-unchanged files (`locked = file.is_ingested && !file.is_stale`) so they could not be selected or re-ingested by accident. This PR removes that lock entirely (`visibleSelectable = files`, no `disabled` on the checkbox) and instead auto-sets `replace_duplicates: true` whenever any selected file `is_ingested` — but adds **no confirmation dialog** to this component. 
- The `DuplicateHandlingDialog` introduced by this PR is only wired into `SharedBucketView`'s "Ingest N Buckets" button, not into `FileBrowserDialog`'s "Browse files" flow (which `SharedBucketView` also renders per-bucket, `shared-bucket-view.tsx:487-499`).
- Net effect: a user who opens "Browse files" for a bucket, clicks "Select all" (now includes every already-ingested file, changed or not), and clicks "Ingest" will silently overwrite every previously-ingested document in that bucket with `replace_duplicates: true` and zero warning.

#### Background Information

Before this PR:
```tsx
const locked = file.is_ingested && !file.is_stale;
...
<input type="checkbox" checked={selected || locked} disabled={locked} ... />
```
After this PR, `locked` is gone and every file — ingested or not, stale or not — is freely selectable, with overwrite silently implied by re-selecting it.

#### Code References

- `frontend/components/file-browser-dialog.tsx:80-83` (`visibleSelectable = files`)
- `frontend/components/file-browser-dialog.tsx:125-146` (`handleIngest` — sets `replace_duplicates: true` with no confirmation)
- `frontend/components/file-browser-dialog.tsx:311-318` (`locked` guard removed from `FileRow`)
- `frontend/components/connectors/shared-bucket-view.tsx:487-499` (this dialog is reachable outside the new confirm flow)

#### Potential Solution

- Route `FileBrowserDialog`'s "Ingest" action through the same `DuplicateHandlingDialog` (or an equivalent inline confirmation) whenever `hasAlreadyIngested` is true, mirroring what `SharedBucketView.ingestSelected` now does. At minimum, restore visual/interactive friction (e.g., a distinct "re-ingest" affordance) so re-selecting an ingested-and-unchanged file is an explicit, confirmed action rather than indistinguishable from selecting a new file.

#### Alternative Solutions

- Keep unchanged files locked (restore prior behavior) and only unlock `is_stale` files, deferring "overwrite anything" to a future pass — this is the smallest, safest diff.
- Show an inline per-file "already ingested — re-ingest?" badge/checkbox state instead of a blocking dialog, if a modal is considered too heavy for single-file selection.

src/api/connectors.py
```diff
@@ -703,6 +739,83 @@ async def _classify_connector_duplicates(
...
 
+async def _classify_bucket_connector_duplicates(
+    connector,
+    connector_type: str,
+    bucket_filter: list[str],
+    session_manager,
+    user_id: str,
+    jwt_token: str | None,
+) -> dict[str, Any]:
+    """Preview a bucket_filter sync: classify remote blobs new/changed/unchanged
```
mpawlow, line 750: ### (1b) [Minor] Bucket duplicate-check docstring contradicts its own implementation

#### Problem

- `_classify_bucket_connector_duplicates`'s docstring claims a three-way `new`/`changed`/`unchanged` classification mirroring `connector_sync`'s `bucket_filter` branch, with "unchanged" blobs "silently dropped." The actual implementation is purely **existence-based** (`fid in existing_set`) — it does not consult `modified_time` at all, so already-ingested-but-unchanged blobs are *not* dropped; they're reported as duplicates just like genuinely changed ones. The comment a few lines below the docstring even explicitly documents this divergence from the docstring's claim, but the docstring itself was never corrected.

#### Code References

`src/api/connectors.py:742-757` (docstring) vs. `src/api/connectors.py:789-816` (actual logic/comment)

#### Potential Solution

- Rewrite the docstring to describe the actual existence-based behavior, referencing the inline comment's rationale directly instead of the stale new/changed/unchanged framing.

src/api/connectors.py
```diff
@@ -703,6 +739,83 @@ async def _classify_connector_duplicates(
...
+
+    existing_ids, _, _ = await get_synced_file_ids_for_connector(
+        connector_type=connector_type,
+        user_id=user_id,
+        session_manager=session_manager,
+        jwt_token=jwt_token,
+    )
+    existing_set = set(existing_ids)
+
+    # Existence-based, like the OAuth connector duplicate check: any blob
```
mpawlow, line 789: ### (1c) [Normal] Existence-based duplicate check causes the confirm dialog (and a full re-embed) on every routine bucket re-sync, even with zero remote changes

#### Problem

- Because `_classify_bucket_connector_duplicates` flags *any* previously-ingested blob as a duplicate (see 1b), re-clicking "Ingest" on a bucket that was already fully synced — with **no** changes at the source — will always surface the overwrite dialog. If the user picks "Overwrite duplicates" (the more prominent, styled button), every unchanged file gets re-ingested/re-embedded unnecessarily. 
- Contrast this with the pre-existing `bucket_filter` reconciliation in `connector_sync` (`src/api/connectors.py:1189-1238`), which already does proper `modified_time`-based new/changed/unchanged classification and cleanly no-ops ("All files ... already up to date") when nothing changed.
- This turns what should be a rare, meaningful prompt (real conflicting content) into dialog fatigue for the routine "re-sync my bucket" workflow, and risks users reflexively clicking "Overwrite" out of habit, causing wasted embedding/OCR compute on unchanged documents.

#### Code References

- `src/api/connectors.py:789-816` (existence-based classification)
- `src/api/connectors.py:1189-1238` (existing, correct modified-time based reconciliation it should mirror)

#### Potential Solution

- Have `_classify_bucket_connector_duplicates` reuse `classify_remote_file_change` (already used in `connector_sync`, `src/api/connectors.py:1210`) with `get_synced_id_to_modified_time_map`, and only surface true `"changed"` blobs as `duplicate_files`; treat `"unchanged"` the way the docstring already (incorrectly) claims — drop them from both lists so the dialog only appears when there's a real conflict.

frontend/components/connectors/shared-bucket-view.tsx
```diff
@@ -132,40 +180,129 @@ export function SharedBucketView({
...
+
+      const checkData =
+        (await checkResponse.json()) as BucketDuplicateCheckResponse;
+      const duplicateNames = checkData.duplicate_names || [];
+      const duplicateCount =
+        typeof checkData.duplicate_count === "number"
+          ? checkData.duplicate_count
+          : duplicateNames.length;
+
+      if (duplicateCount === 0) {
```
mpawlow, line 254: ### (1d) [Normal] Redundant full bucket listing on every "no duplicates" ingest

#### Problem

- `ingestSelected` in `SharedBucketView` calls `check-duplicates`, which lists every object across the selected buckets (`_classify_bucket_connector_duplicates`, unbounded pagination). When `duplicateCount === 0`, the frontend discards that listing and calls `runBucketSync()`, which re-triggers `connector_sync`'s `bucket_filter` branch — listing the same buckets from scratch again. For large buckets this doubles listing latency/cost on every ingest click, when the check step already has the complete `non_duplicate_files` list in hand.

#### Potential Solution

When `duplicateCount === 0`, call `runSelectedFilesSync(checkData.non_duplicate_files, false)` directly instead of `runBucketSync()`, avoiding the second full listing.

frontend/components/connectors/shared-bucket-view.tsx
```diff
@@ -132,40 +180,129 @@ export function SharedBucketView({
...
+      console.error("[Bucket Sync] Duplicate check failed:", err);
+      // Fallback: proceed with the normal full sync (backend still handles
+      // new/changed reconciliation on its own).
+      runBucketSync();
+    } finally {
+      setIsCheckingDuplicates(false);
+    }
+  };
+
+  const handleOverwriteDuplicates = () => {
```
mpawlow, line 276: ### (1e) [Minor] Fragile ref-based workaround for dialog close/overwrite race, with no test coverage

#### Problem

- `isOverwriteConfirmedRef` exists to suppress `handleDuplicateDialogOpenChange`'s "skip" branch from re-firing when `DuplicateHandlingDialog.handleOverwrite` calls `onOpenChange(false)` right after `onOverwrite()`. It works (verified by tracing the closure/microtask order), but it's a subtle pattern: it depends on `pendingDuplicates` still being non-null in the *stale* closure at the moment `onOpenChange(false)` fires synchronously after `onOverwrite()`, before React flushes the `setPendingDuplicates(null)` update. Any future change to `DuplicateHandlingDialog`'s call order (e.g., making `onOverwrite` awaited before closing, or vice versa) could silently break this. No frontend/e2e test exercises either the overwrite or skip path.

#### Potential Solution

- Prefer a single source of truth for "why did the dialog close" — e.g., have `onOverwrite` itself own closing the dialog and clearing `pendingDuplicates` in one state update, and have the dialog's own Cancel/backdrop-close path call a distinct `onSkip` callback rather than overloading `onOpenChange`. Add a Playwright test (or component test) covering: no-duplicates auto-sync, duplicates→overwrite, duplicates→skip, duplicates→dismiss-without-choice.

src/api/connectors.py
```diff
@@ -69,6 +69,15 @@ def _connector_sync_should_replace(connector_type: str) -> bool:
     return connector_type in ["google_drive", "sharepoint", "onedrive"]
 
 
+def _is_unmapped_keyword_agg_error(err: Exception) -> bool:
```
mpawlow, line 72: ### (1f) [Minor] Exception-message string matching for the keyword-mapping fallback is brittle

#### Problem

- `_is_unmapped_keyword_agg_error` detects the legacy-mapping case by substring-matching the raw exception text (`"Text fields are not optimised"` / `"fielddata"`). This is coupled to OpenSearch's current English error wording and could silently stop matching (falling through to `raise`) after an OpenSearch version bump changes the message. Low risk since the outer `try/except` in both callers fails safe (returns empty results / defaults `id_field` to `"document_id"`), but it's worth a code comment noting the coupling, or matching on the OpenSearch/Elasticsearch structured error type/reason code if available in the client library.

#### Potential Solution

If the OpenSearch async client surfaces a structured error (e.g., `RequestError` with `.info["error"]["root_cause"][0]["reason"]` / `.error` type `"illegal_argument_exception"`), match on that instead of a free-text substring for resilience across versions.

mpawlow (changes requested):
## Code Review 1

- See blocking PR comment (1a)  - Same issue as Functional Test 1 - Scenario 2
- See non-blocking PR comments (1c) (1d) for consideration
- See Minor PR comments (1b), (1e), (1f) for optional consideration

edwinjosechittilappilly:
https://github.com/user-attachments/assets/f08843f1-4d8d-4f7e-abf3-2d77c7f47b5f

frontend/components/file-browser-dialog.tsx
```diff
@@ -125,169 +137,270 @@ export function FileBrowserDialog({
     [allFiles, selectedFileIds],
   );
 
+  const [isCheckingDuplicates, setIsCheckingDuplicates] = useState(false);
```
mpawlow, line 140: ### (2a) [Minor] Duplicate-confirmation flow is now copy-pasted near-verbatim across two components

#### Problem

- The fix for issue (1a) — routing `FileBrowserDialog`'s ingest action through a duplicate-check + confirm dialog — was implemented by copying the entire pattern already present in `SharedBucketView` (`shared-bucket-view.tsx:213-304`) into `FileBrowserDialog` (`file-browser-dialog.tsx:140-263`) almost line-for-line: the same `fetch("/api/connectors/{type}/check-duplicates")` call, the same `pendingSync`/`pendingDuplicates` state shape, the same `submitSync`/`runSelectedFilesSync` mutate wrapper, the same `handleOverwriteDuplicates`, and the same `isOverwriteConfirmedRef` race workaround (see 2b).
- This is a legitimate DRY violation: any future fix to the duplicate-check flow (e.g., addressing 1c/1e from Review 1) now has to be applied in two places, and they will silently drift apart over time (they already differ slightly — `FileBrowserDialog` omits the `duplicateFiles` field `SharedBucketView` tracks, since it doesn't need to re-list duplicate files separately).
- This also (partially) explains the react-doctor bot's new warning on this file: `file-browser-dialog.tsx:52` — "Large component is hard to read and change" (`no-giant-component`). The component grew from ~290 to ~468 lines, largely due to this inlined duplicate-check logic.

#### Code References

- `frontend/components/file-browser-dialog.tsx:140-263` (new duplicate-check state/handlers)
- `frontend/components/connectors/shared-bucket-view.tsx:213-304` (the original, now-duplicated pattern)

#### Potential Solution

- Extract a shared hook, e.g. `useDuplicateConfirmSync({ connectorType, checkPayload, onSync })`, that owns the `check-duplicates` fetch, the `pending`/`dialogOpen` state, the overwrite/skip handlers, and the ref-based close-race workaround. Both `SharedBucketView` and `FileBrowserDialog` would then call the hook and render `<DuplicateHandlingDialog>` with its returned props — fixing 1e once fixes it everywhere.

#### Alternative Solutions

- If a shared hook feels like too large a refactor for this PR, leave as-is and file a fast-follow ticket referencing both call sites — acceptable given this is a bug-fix PR, but the duplication should be tracked, not silently left.

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM  🚀 
   - See Minor PR comment (2a) for optional consideration
   - Note: Code Review 1 PR comments still applicable: (2b) ... (2f)  

## Functional Review 2

- ✅ Successful 
   - Verified resolution for **Functional Review 1 - Scenario 2**

PR #2078, opened 2026-07-13
fix: show confirmation dialog before disconnecting a connector

yogichipalkatti:
File changed: [frontend/app/settings/_components/connector-cards.tsx](vscode-webview://1udpbjun8njpsks0b39nrb88e947nvlr9u1akej9ikgr2jppgg8u/frontend/app/settings/_components/connector-cards.tsx)

What was done
Added disconnectTarget state — holds the Connector the user clicked "disconnect" on, or null when the dialog is closed.

handleDisconnect no longer fires the mutation directly — it now just calls setDisconnectTarget(connector), which opens the confirmation dialog.

confirmDisconnect — new function that runs disconnectMutation.mutateAsync and clears the target on success. Uses mutateAsync so the isLoading spinner in the dialog works correctly.

<DeleteConfirmationDialog> — rendered once at the bottom of the component tree (outside the grid), controlled by disconnectTarget !== null. It shows:

Title: "Disconnect <connector name>?"
Description: "This will remove the <name> connection from OpenRAG."
Body warning: explains that previously ingested documents remain but syncing stops until reconnection
A destructive Disconnect confirm button with a loading spinner while the mutation is in flight
A Cancel button to abort without any side-effects

Screenshot 2026-07-13 at 11 05 20 AM

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a confirmation step before disconnecting a connector.
  * Uses a connector-specific confirmation dialog to help prevent accidental disconnections.
  * Shows a loading state while the disconnect is being processed.
  * Disconnects only after you confirm, then closes the dialog.
  * Cancelling or closing the dialog leaves the connector connected.


frontend/app/settings/_components/connector-cards.tsx
```diff
@@ -62,8 +66,16 @@ export default function ConnectorCards() {
...
+  const handleDisconnect = (connector: Connector) => {
+    setDisconnectTarget(connector);
+  };
+
+  const confirmDisconnect = async () => {
+    if (!disconnectTarget) return;
+    await disconnectMutation.mutateAsync(
+      disconnectTarget as unknown as QueryConnector,
+    );
+    setDisconnectTarget(null);
```
Wallgau: confirmDisconnect awaits mutateAsync with no error handling. If the disconnect request fails, the promise rejects and bubbles out of the dialog's handleConfirm as an unhandled rejection (console noise). The error toast still fires via the mutation's onError, but we should swallow the rejection here and only clear the target on success.
Suggested change:
const confirmDisconnect = async () => {
    if (!disconnectTarget) return;
    try {
      await disconnectMutation.mutateAsync(
        disconnectTarget as unknown as QueryConnector,
      );
    } finally {
      setDisconnectTarget(null);
    }
  };

Wallgau (changes requested):

Wallgau (approved):
Nice work! Thanks @yogichipalkatti

PR #2095, opened 2026-07-14
fix: missing ingest settings for individual files (backport of #2066)

edwinjosechittilappilly:
## Summary
Backports #2066 from `release-saas-ga-0.6.2` to `main`.

**Depends on #2094** (backport of #2050), #2093 (#2035), #2092 (#2025) — stacked because it touches `shared-bucket-view.tsx`/`file-browser-dialog.tsx` in the same regions. Merge order: #2092 → #2093 → #2094 → this PR.

Note: cherry-pick had conflicts in `frontend/components/file-browser-dialog.tsx` and `frontend/components/connectors/shared-bucket-view.tsx` against an unrelated `onIngestSuccess` feature already on `main`; resolved by keeping both additive changes (verified with `tsc --noEmit`, no errors in either file).

## Test plan
- [ ] CI passes
- [ ] Manually verify ingest settings are applied when ingesting individual files via the file browser dialog

## Summary by CodeRabbit

* **New Features**
  * Added IBM Cloud Object Storage connector configuration and local-development enablement options.
  * Added support for shared/private document visibility during bucket uploads and synchronization.
  * Added a configurable shared-upload toggle in settings.
  * Added controls to show or hide advanced ingestion options.
  * Updated existing documents when their shared/private visibility changes.

* **Bug Fixes**
  * Ensured shared-upload settings are consistently applied to new, changed, and existing files.

* **Tests**
  * Added coverage for IBM COS availability, shared uploads, visibility reconciliation, and configuration toggles.

PR #2094, opened 2026-07-14
fix: cos share all (backport of #2050)

edwinjosechittilappilly:
## Summary
Backports #2050 from `release-saas-ga-0.6.2` to `main`.

**Depends on #2093** (backport of #2035) and #2092 (backport of #2025) — this branch is stacked on `backport/2035-shareall-default-true` because it touches `src/config/settings.py` in the same region. Merge order: #2092 → #2093 → this PR.

## Test plan
- [ ] CI passes

## Summary by CodeRabbit

* **New Features**
  * Added IBM Cloud Object Storage as an available bucket connector.
  * Added configuration options for IBM COS authentication and local development access.
  * Added a configurable shared-upload toggle for bucket uploads.
  * Advanced ingestion settings can now be hidden while keeping shared uploads available.

* **Bug Fixes**
  * Shared-upload settings are now consistently applied when syncing new, changed, duplicate, or unchanged files.
  * File ownership metadata is updated correctly when shared status changes.

PR #2093, opened 2026-07-14
chore: default OPENRAG_SHOW_SHARED_UPLOAD_TOGGLE to true (backport of #2035)

edwinjosechittilappilly:
## Summary
Backports #2035 from `release-saas-ga-0.6.2` to `main`.

Changes the default value of `OPENRAG_SHOW_SHARED_UPLOAD_TOGGLE` to `true`.

**Depends on #2092** (backport of #2025, which introduces the flag) — this branch is stacked on top of `backport/2025-shared-upload-toggle-flag` and should be merged after it, or rebased onto `main` once #2092 lands.

## Test plan
- [ ] CI passes
- [ ] Manually verify the toggle defaults to enabled

## Summary by CodeRabbit

- **New Features**
  - Added an independent setting to control visibility of the “Make documents available to all users” upload toggle.
  - Advanced ingestion options can now be hidden while keeping the shared-upload option available.
  - Shared-upload behavior is now configurable separately from advanced ingestion settings.

- **Bug Fixes**
  - Improved upload setting visibility and handling based on configured preferences.

- **Tests**
  - Added coverage for default, enabled, disabled, and independent toggle behavior.

PR #2092, opened 2026-07-14
feat: add dedicated OPENRAG_SHOW_SHARED_UPLOAD_TOGGLE flag for COS "share all" toggle (backport of #2025)

edwinjosechittilappilly:
## Summary
Backports #2025 from `release-saas-ga-0.6.2` to `main`.

Adds a dedicated `OPENRAG_SHOW_SHARED_UPLOAD_TOGGLE` flag to gate the COS "share all" upload toggle.

## Test plan
- [ ] CI passes
- [ ] Manually verify the toggle is gated by the new flag

## Summary by CodeRabbit

* **New Features**
  * Added configuration support for independently showing a shared-upload toggle.
  * Added settings visibility control for advanced ingestion options, including embedding, chunking, OCR, and image descriptions.
  * Shared upload controls now appear independently from other ingestion settings.

* **Bug Fixes**
  * Improved handling of shared-upload settings so they are only displayed and submitted when enabled.

* **Tests**
  * Added coverage for default, enabled, and independent configuration behavior.

PR #2096, opened 2026-07-14
fix: user-facing duplicate handling workflow for bucket-based connectors (backport of #2077)

edwinjosechittilappilly:
## Summary
Backports #2077 from `release-saas-ga-0.6.2` to `main`.

**Depends on #2095** (backport of #2066), #2094 (#2050), #2093 (#2035), #2092 (#2025) — stacked because it substantially rewrites the ingest flow in `frontend/components/file-browser-dialog.tsx` and `frontend/components/connectors/shared-bucket-view.tsx`, both touched by earlier backports. Merge order: #2092 → #2093 → #2094 → #2095 → this PR.

Note on conflict resolution: `main` has an `onIngestSuccess` callback prop on `FileBrowserDialog` (added independently of the release branch) that release's version of this file doesn't have. Resolved by taking release's file (duplicate-check flow) as the base and re-threading `onIngestSuccess` through `submitSync` so it still fires with the mutation result on successful ingest, matching its original main behavior. Verified with `tsc --noEmit` — no errors in either touched file.

## Test plan
- [ ] CI passes
- [ ] Manually verify: ingesting files with existing duplicates in a bucket connector shows the duplicate-handling dialog (skip/overwrite), and `onIngestSuccess` still fires (task tracking / invalidation) on success

## Summary by CodeRabbit

- **New Features**
  - Added duplicate detection before ingesting files and bucket selections.
  - Users can overwrite duplicates or skip them during syncing.
  - Added a duplicate-handling dialog with duplicate vs non-duplicate counts and names.
  - Added shared-ingest options that appear only when appropriate, affecting what gets sent during ingest.
  - Improved ingest UX by reflecting duplicate-check progress in button state/label.
- **Bug Fixes**
  - Improved OpenSearch compatibility by retrying lookups when index mappings differ.
  - Added clearer handling when all selected items are already imported (with informative messaging).
- **UX Improvements**
  - Expanded selection behavior to allow re-ingesting already imported files.


src/api/connectors.py
mpawlow, line 613: ### (1a) [Major] `connection_id` is not threaded through the real sync request, only the duplicate-check preview

#### Problem

- `ConnectorCheckDuplicatesBody` (used by `/check-duplicates`) has a `connection_id` field, so the duplicate-check preview runs against a specific connection when the user has more than one active connection of the same `connector_type`. 
- `ConnectorSyncBody` (used by the actual `connector_sync` ingest call) has **no** `connection_id` field at all — it always falls back to "the first connection that actually works." `shared-bucket-view.tsx` sends `connection_id` in its sync body anyway, but since `ConnectorSyncBody` doesn't declare that field, Pydantic's default `extra="ignore"` behavior silently drops it. 
- `file-browser-dialog.tsx`'s `submitSync` doesn't even attempt to send it.

#### Background Information

- This is a pre-existing gap in `connector_sync`'s connection-resolution logic (`src/api/connectors.py:1026` picks the first authenticating connection), not something newly introduced by the backport — but #2077's new duplicate-preview flow is the first place in the codebase that lets a user preview against a specific connection and then act on that preview via a follow-up sync call, so it's the first place the mismatch is user-visible: a user with two S3/COS/Azure connections could see duplicates computed against connection A, choose "overwrite," and have the actual overwrite happen against connection B instead. This was also flagged independently by CodeRabbit on PR #2096.

#### Code References

- `src/api/connectors.py:612-629` (`ConnectorSyncBody` — no `connection_id`)
- `src/api/connectors.py:632-639` (`ConnectorCheckDuplicatesBody` — has `connection_id`)
- `src/api/connectors.py:1026,1075-1090` (`connector_sync` connection resolution — first-working fallback only)
- `frontend/components/connectors/shared-bucket-view.tsx:180,206` (sends `connection_id` that the backend ignores)
- `frontend/components/file-browser-dialog.tsx:152-186` (`submitSync` — no `connection_id` sent at all)

#### Potential Solution

- Add `connection_id: str | None = None` to `ConnectorSyncBody`, and in `connector_sync` prefer `body.connection_id` (mirroring the `if body.connection_id: ... else: first-working` pattern already in `connector_check_duplicates`) before falling back to "first working connection."

#### Alternative Solutions

- Have the duplicate-check response echo back the resolved `connection_id`, and require the frontend to pass it through unmodified on the follow-up sync call (makes the coupling explicit at the API boundary).

src/api/connectors.py
```diff
@@ -735,6 +771,83 @@ async def _classify_connector_duplicates(
...
+) -> dict[str, Any]:
+    """Preview a bucket_filter sync: classify remote blobs new/changed/unchanged
+    without ingesting anything, mirroring the reconciliation in connector_sync's
+    bucket_filter branch. "changed" blobs are reported as duplicates (they would
+    overwrite an already-indexed version); "unchanged" blobs are silently
+    dropped (the real sync would skip them too); "new" blobs are returned as
+    ``non_duplicate_files`` so the caller can sync just those when the user
+    chooses to skip duplicates.
+    """
+    original_buckets = connector.bucket_names
```
mpawlow, line 793: ### (1b) [Normal] Duplicate-check preview mutates a cached, shared connector's `bucket_names` without per-request isolation

#### Problem

- `_classify_bucket_connector_duplicates` temporarily overwrites `connector.bucket_names = bucket_filter`, calls `connector.list_files(...)`, then restores it in a `finally` block. `connector_service.get_connector(connection_id)` (via `connection_manager`) returns a **cached** connector instance keyed by `connection_id` (`self.active_connectors[connection_id]`), not a fresh instance per request. If two requests for the same connection race (e.g. two browser tabs, or a preview check overlapping with a real sync), one request's temporary `bucket_names` mutation can leak into the other's `list_files()` call, causing it to list the wrong buckets.

#### Background Information

- This mutate/restore-in-`finally` pattern is not unique to the new code — the same pattern already exists in `connector_sync`'s own `bucket_filter` branch (`src/api/connectors.py:1198-1209`) and elsewhere in the file (e.g. `:2381-2386`), so this PR extends an existing architectural shortcut rather than introducing a new one. 

#### Code References

- `src/api/connectors.py:790-801` (`_classify_bucket_connector_duplicates`)
- `src/api/connectors.py:1198-1209` (same pattern in `connector_sync`, pre-existing)
- `src/connectors/connection_manager.py:431-440` (`get_connector` connector caching by `connection_id`)

#### Potential Solution

- Pass `bucket_filter` as a parameter into `connector.list_files(bucket_filter=..., page_token=...)` instead of mutating shared instance state, so no shared mutable state is touched across concurrent requests.

#### Alternative Solutions

- Acquire a per-`connection_id` asyncio lock around the mutate/restore window as a smaller, lower-risk interim fix if changing the `list_files` signature is out of scope for this backport.

src/api/connectors.py
```diff
@@ -75,6 +75,15 @@ def _connector_sync_should_replace(connector_type: str) -> bool:
     return connector_type in ["google_drive", "sharepoint", "onedrive"]
 
 
+def _is_unmapped_keyword_agg_error(err: Exception) -> bool:
```
mpawlow, line 79: ### (1c) [Normal] `_is_unmapped_keyword_agg_error`'s broad substring match can mask unrelated aggregation failures

#### Problem

- `_is_unmapped_keyword_agg_error` returns `True` on `"Text fields are not optimised" in msg or "fielddata" in msg`. The bare `"fielddata"` substring check is broader than the specific legacy-mapping error this helper is meant to detect, so an unrelated OpenSearch error that happens to mention "fielddata" would also trigger the `.keyword`-field retry path instead of surfacing as a real failure.

#### Background Information

- CodeRabbit flagged this as Major; but it is Normal because the retry re-runs the same query with a renamed field — if the actual problem is unrelated, the retry itself will very likely still raise (there's no swallowing `except` around the retry call), so the most common failure mode is a wasted retry rather than a silently wrong result. The narrower risk is a transient/unrelated fielddata error that happens to succeed on retry, returning results computed against a re-issued query that may not reflect the original triggering condition.

#### Code References

- `src/api/connectors.py:78-84` (`_is_unmapped_keyword_agg_error`)
- `src/api/connectors.py:127-137` (first call site / retry)
- `src/api/connectors.py:291-297` (second call site / retry)

#### Potential Solution

- Narrow the check to the specific mapping-error text (`"Text fields are not optimised"` / the associated OpenSearch error type), dropping the standalone `"fielddata"` substring match.

#### Alternative Solutions

- Match on the specific OpenSearch exception type/`error.type` field (e.g. `illegal_argument_exception` with a `caused_by` mapping-related reason) instead of a raw string substring match, which is generally more robust against message-text drift across OpenSearch versions.

src/api/connectors.py
```diff
@@ -735,6 +771,83 @@ async def _classify_connector_duplicates(
     }
 
 
+async def _classify_bucket_connector_duplicates(
```
mpawlow, line 777: ### (1d) [Minor] Docstring on `_classify_bucket_connector_duplicates` contradicts its actual (correct) behavior

#### Problem

- The docstring claims the function classifies blobs as "new/changed/unchanged" mirroring `connector_sync`'s modified-time reconciliation. The implementation — and its own inline comment a few lines below — explicitly does existence-only classification (any blob whose ID already exists is a "duplicate", regardless of modification time). A future maintainer reading only the docstring will misunderstand the semantics.

#### Code References

- `src/api/connectors.py:774-789` (docstring)
- `src/api/connectors.py:820-827` (inline comment describing the actual, correct behavior)

#### Potential Solution

- Update the docstring to describe existence-only classification, as CodeRabbit's suggested diff does — align it with the inline comment already in the function body.

src/api/connectors.py
```diff
@@ -735,6 +771,83 @@ async def _classify_connector_duplicates(
...
+    """Preview a bucket_filter sync: classify remote blobs new/changed/unchanged
+    without ingesting anything, mirroring the reconciliation in connector_sync's
+    bucket_filter branch. "changed" blobs are reported as duplicates (they would
+    overwrite an already-indexed version); "unchanged" blobs are silently
+    dropped (the real sync would skip them too); "new" blobs are returned as
+    ``non_duplicate_files`` so the caller can sync just those when the user
+    chooses to skip duplicates.
+    """
+    original_buckets = connector.bucket_names
+    connector.bucket_names = bucket_filter
```
mpawlow, line 794: ### (1f) [Normal] Existence-based duplicate check triggers the confirm dialog (and a full re-embed) on every routine bucket re-sync, even with zero remote changes

#### Problem

- Because `_classify_bucket_connector_duplicates` flags *any* previously-ingested blob as a duplicate (see 1d), re-clicking "Ingest" on a bucket that was already fully synced — with no changes at the source — always surfaces the overwrite dialog. If the user picks "Overwrite duplicates" (the more prominent, styled button), every unchanged file gets re-ingested/re-embedded unnecessarily.

#### Background Information

- Identified in an earlier code review of the source PR #2077 (thread still unresolved/not outdated there) and independently re-confirmed here as still present in #2096's code, since `_classify_bucket_connector_duplicates` is byte-identical between the two PRs. Contrast with the pre-existing `bucket_filter` reconciliation in `connector_sync` (`src/api/connectors.py:1198-1238`), which already does proper `modified_time`-based new/changed/unchanged classification and cleanly no-ops when nothing changed. This turns what should be a rare, meaningful prompt (real conflicting content) into dialog fatigue for the routine "re-sync my bucket" workflow, and risks users reflexively clicking "Overwrite" out of habit, wasting embedding/OCR compute on unchanged documents.

#### Code References

- `src/api/connectors.py:790-827` (existence-based classification)
- `src/api/connectors.py:1198-1238` (existing, correct modified-time-based reconciliation it should mirror)

#### Potential Solution

- Have `_classify_bucket_connector_duplicates` reuse the existing `modified_time`-based reconciliation helper already used in `connector_sync`'s `bucket_filter` branch, and only surface true "changed" blobs as `duplicate_files`; drop "unchanged" blobs from both lists so the dialog only appears when there's a real conflict.

frontend/components/connectors/shared-bucket-view.tsx
```diff
@@ -128,38 +181,134 @@ export function SharedBucketView({
...
+        },
+      );
+
+      if (!checkResponse.ok) {
+        throw new Error(`Duplicate check failed: ${checkResponse.statusText}`);
+      }
+
+      const checkData =
+        (await checkResponse.json()) as BucketDuplicateCheckResponse;
+      const duplicateNames = checkData.duplicate_names || [];
```
mpawlow, line 254: ### (1g) [Normal] Redundant full bucket listing on every "no duplicates" ingest

#### Problem

- `ingestSelected` in `SharedBucketView` calls `check-duplicates`, which lists every object across the selected buckets (`_classify_bucket_connector_duplicates`, unbounded pagination). When `duplicateCount === 0`, the frontend discards that listing and calls `runBucketSync()`, which re-triggers `connector_sync`'s `bucket_filter` branch — listing the same buckets from scratch again. For large buckets this doubles listing latency/cost on every ingest click, when the check step already has the complete `non_duplicate_files` list in hand.

#### Background Information

- Identified in an earlier code review of the source PR #2077 (thread still unresolved/not outdated there); re-verified here that `runBucketSync()` is still called unconditionally on the `duplicateCount === 0` path in #2096's `shared-bucket-view.tsx:260-261`, and `non_duplicate_files` is available on `checkData` but unused in that branch.

#### Potential Solution

- When `duplicateCount === 0`, call `runSelectedFilesSync(checkData.non_duplicate_files, false)` directly instead of `runBucketSync()`, avoiding the second full listing.

frontend/components/connectors/shared-bucket-view.tsx
```diff
@@ -65,6 +86,15 @@ export function SharedBucketView({
...
   );
+  const [isCheckingDuplicates, setIsCheckingDuplicates] = useState(false);
+  const [duplicateDialogOpen, setDuplicateDialogOpen] = useState(false);
+  const [pendingDuplicates, setPendingDuplicates] = useState<{
+    duplicateNames: string[];
+    duplicateCount: number;
+    duplicateFiles: BucketDuplicateFile[];
+    nonDuplicateFiles: BucketDuplicateFile[];
+  } | null>(null);
+  const isOverwriteConfirmedRef = useRef(false);
```
mpawlow, line 97: ### (1h) [Minor] Fragile ref-based workaround for dialog close/overwrite race, with no test coverage

#### Problem

- `isOverwriteConfirmedRef` exists to suppress `handleDuplicateDialogOpenChange`'s "skip" branch from re-firing when `DuplicateHandlingDialog`'s overwrite handler calls `onOpenChange(false)` right after `onOverwrite()`. It works, but depends on `pendingDuplicates` still being non-null in a stale closure at the moment `onOpenChange(false)` fires synchronously after `onOverwrite()`, before React flushes the state update clearing it. Any future change to `DuplicateHandlingDialog`'s call order could silently break this, and no frontend/e2e test exercises either the overwrite or skip path.

#### Background Information

- Identified in an earlier code review of the source PR #2077 (thread still unresolved/not outdated there); re-verified here that `isOverwriteConfirmedRef` is still present and used identically in #2096's `shared-bucket-view.tsx:97,284,292,296`.

#### Potential Solution

- Have a single source of truth for "why did the dialog close" — e.g., have the overwrite handler own closing the dialog and clearing `pendingDuplicates` in one state update, and have the dialog's own Cancel/backdrop-close path call a distinct `onSkip` callback rather than overloading `onOpenChange`. Add a Playwright test covering: no-duplicates auto-sync, duplicates→overwrite, duplicates→skip, duplicates→dismiss-without-choice.

mpawlow (commented):
## Code Review 1

- See Major PR comment (1a) for assessment / evaluation
- See Normal PR comments (1b) (1c) (1f) (1g) for consideration
- See Minor PR comments (1d) (1e) (1h) for optional consideration
- ⚠️ Note: Most of these PR review comments are carried over from PR #2077
- ✅  Verified: All relevant changes from PR #2077 are cleanly applied to this PR #2096.
   - Given that this PR's changes exactly represents what will be deployed to SaaS, let me know if it makes sense to address the remaining issues in a subsequent PR  👀

mpawlow (approved):

src/api/connectors.py
edwinjosechittilappilly: @coderabbitai can you create an issue wrt to the review comments above and assign it to me and @mpawlow

PR #2103, opened 2026-07-15
refactor: Connector sync improvements

mfortman11:
## Summary by CodeRabbit

* **New Features**
  * Added `replace_duplicates` support for bucket uploads to control replacement vs skipping existing object keys.
  * Connector sync now uses connector-declared change-detection to drive updates-only sync and improves selected-file expansion for more accurate incremental behavior.

* **Bug Fixes**
  * Reduced incorrect orphan deletions by reconciling and cleaning up using both supported file-id layouts, with centralized chunk cleanup.
  * Improved duplicate/filename detection using bulk OpenSearch checks (including automatic index creation when needed) and consistent filename-alias handling.
  * Strengthened shared connector sync authorization by requiring the anonymous delete permission earlier when applicable.


edwinjosechittilappilly:
@coderabbitai what are your thoughts on moving the apis of specific connectors also into the enhancement folders, so we can completely make the connectors pluggable step by step ?

src/models/processors.py
```diff
@@ -333,61 +389,22 @@ async def _delete_connector_chunks(
...
-                        },
-                        owner_filter,
-                    ]
-                }
-            }
-            if keep_filenames:
-                query["bool"]["must_not"] = [{"terms": {"filename": keep_filenames}}]
-
-            chunk_ids = await collect_visible_document_ids(
+            return await delete_connector_file_chunks(
```
Wallgau, line 407: I think we should scope _delete_connector_chunks by connector_type (like delete_orphan_documents), let's add connector_type to the function signature. That way the rename/404 cleanup only removes chunks for the relevant connector and doesn't accidentally delete chunks from another connector that happens to share the same ID.

mpawlow (commented):
## Pre-Code Review / Pre-Functional Review

@mfortman11 

- `Request`: Can you rebase your branch off of the latest from **main** and resolve the file conflicts (when you get a chance)?
- I'm noticing some related / residual issues in your PR that were surfaced and addressed by @edwinjosechittilappilly in his Pull Request (already merged into *main*)
   - https://github.com/langflow-ai/openrag/pull/2096

src/connectors/base.py
```diff
@@ -175,6 +193,32 @@ async def list_files(
...
+
+        Connectors are cached and shared across requests by ConnectionManager,
+        so we operate on a per-call shallow copy with its own cfg instead of
+        mutating (and racing on) the shared config.
+        """
+        cfg = getattr(self, "cfg", None)
+        if cfg is None:
+            return await self.list_files(**kwargs)
+        scoped = copy.copy(self)
+        scoped.cfg = copy.copy(cfg)
```
mpawlow, line 230: ### (2a) [Critical] CI is currently red — mypy fails on the `list_selected_files()` fix 

#### Problem

- The race-condition fix for Review #1 finding `1b` replaced in-place `cfg` mutation with a shallow-copy approach: `scoped = copy.copy(self); scoped.cfg = copy.copy(cfg)`.
- `BaseConnector` never declares a `cfg` attribute (no `self.cfg = ...` in `__init__`, no class-level annotation) — only subclasses set it. mypy's `attr-defined` check fails on assigning to an attribute the base class doesn't declare.
- Actual CI output (`Ruff and mypy on changed files`, run `29881364725`, 2026-07-22):
  ```
  connectors/base.py:217: error: "BaseConnector" has no attribute "cfg"  [attr-defined]
  connectors/base.py:218: error: "BaseConnector" has no attribute "cfg"  [attr-defined]
  connectors/base.py:219: error: "BaseConnector" has no attribute "cfg"  [attr-defined]
  Found 3 errors in 1 file (checked 11 source files)
  ##[error]Process completed with exit code 1.
  ```

#### Potential Solution

- Declare `cfg: Any = None` (or a proper protocol/base type) as a class attribute on `BaseConnector` so mypy recognizes the attribute exists on the base type, then let subclasses override with their concrete `cfg` type.

mpawlow (commented):
## Code Review 2 - Part 1

- See PR comment: (2a)
- Added a commit addressing (2a)

enhancements/connectors/azure_blob/connector.py
mpawlow, line 10: ### (1d) [Minor] Stale docstring references a removed method

#### Problem

- This PR deletes `ConnectorService.process_connector_document()` from `src/connectors/service.py` (confirmed dead code — zero remaining callers anywhere in `src/`, `tests/`, or `enhancements/`) along with its regression test (`tests/unit/connectors/test_langflow_connector_txt_to_md.py`).
- `enhancements/connectors/azure_blob/connector.py` (not touched by this PR) still documents its owner-assignment DLS model in terms of the now-deleted method: *"DLS (v1): owner-based... `service.process_connector_document` assigns `owner` to the ingesting OpenRAG user"* (module docstring, lines ~9-11) and a matching inline comment at line ~337.
- Not a functional bug (ownership assignment still happens correctly via `ConnectorFileProcessor`/`process_document_standard`), but the comment is now actively misleading for anyone tracing the DLS ownership flow for Azure Blob.

#### Code References

- `enhancements/connectors/azure_blob/connector.py:337`

#### Potential Solution

- Update both references to point at the current mechanism (`ConnectorFileProcessor.process_document_standard` / `_update_connector_metadata`) as part of this PR, since it's the PR that invalidated them.

mpawlow: Fixed.

src/connectors/base.py
mpawlow: Fixed.

src/models/processors.py
mpawlow: I concur with @Wallgau here 

### (1c) [Major] Rename/404 chunk cleanup not scoped by `connector_type`

#### Problem

- `connectors/chunk_cleanup.py`'s `build_connector_file_chunks_query()` accepts an optional `connector_type` filter, and `reconcile_orphans_for_connector_type` → `delete_orphan_documents` correctly passes it.
- However, `TaskProcessor._delete_connector_chunks()` (`src/models/processors.py:382-415`) — used for both the per-file 404 cleanup (line ~999) and rename cleanup (line ~1070) inside `ConnectorFileProcessor` — never passes `connector_type`, even though `self.connector_type` is already available on the instance (set at `src/models/processors.py:867`, resolved at `:938`).
- Without this scoping, a connector file ID that happens to collide with an ID from a *different* connector type (e.g. a Langflow-path `document_id` hash colliding with another connector's stable ID) could have its chunks incorrectly deleted or preserved during 404/rename cleanup.

#### Code References

- `src/models/processors.py:382-415` (`_delete_connector_chunks` — no `connector_type` param)
- `src/models/processors.py:999-1001` (404 cleanup call site — has `self.connector_type` available, doesn't pass it)
- `src/models/processors.py:1069-1077` (rename cleanup call site — same gap)
- `src/connectors/chunk_cleanup.py:22-50` (query builder already supports `connector_type`)

#### Potential Solution

- Add `connector_type: str | None = None` to `_delete_connector_chunks()`'s signature, thread it through to `delete_connector_file_chunks()`, and pass `self.connector_type` (already resolved in `sync_specific_files`) at both call sites.

mpawlow (changes requested):
## Code Review 2 - Part 2

- See Major PR comment (1c) - Concur with @Wallgau 
- See Minor PR comment (1d) 
   - Added a commit to address (1d) since it was trivial

## Functional Review 1

- ⚠️  Encountered a problem / regression ingesting a single container using Azure Blob Storage
    - There are 2 containers available to ingest
    - Selected first container to ingest
    - BUG: First and second containers were ingested
    - Will send recording offline

src/models/processors.py
mpawlow: Verified.

mpawlow (approved):
## Code Review 3

- ✅ Approved / LGTM  🚀 

## Functional Review 2

- ✅ All manual test scenarios were successful (Azure Blob Storage)

PR #2106, opened 2026-07-16
fix: speed up CI image builds with BuildKit layer caching and a shared cross-PR cache

Wallgau:
Speed up CI image builds with BuildKit layer caching + a shared cross-PR cache

Summary
Reworks the build-images job in test-ci.yml to stop rebuilding unchanged Docker layers and to share a global layer cache across PRs. The expensive images (Langflow's torch/pip layer, backend's uv sync layer) are now restored from cache instead of rebuilt from scratch.

What changed
Replaced whole-image tar caching with BuildKit layer caching. Dropped the get_image_hash.py → actions/cache → docker save/docker load tar shuttle and the custom build_or_load_image/pull_image bash. Builds now go through docker/build-push-action@v6 with cache-from/cache-to, so only the layers that actually changed rebuild. A flows-only change no longer invalidates the dependency-install layer.
Direct push to GHCR. Non-fork builds push straight to ghcr.io/<repo>/openrag-<image>:${sha} — no local load, no tarball artifact for internal PRs.

Global shared registry cache. Each image keeps a fixed cache tag ghcr.io/<repo>/openrag-<image>:buildcache (type=registry, not branch-scoped, no 10 GB Actions-cache cap). main writes it; every internal PR reads it — so a PR that doesn't touch an image gets that image's layers for free on its first run.
main seeds the cache cheaply. Added a push: [main] trigger so a merge re-warms :buildcache. The test-suite, e2e-run, e2e-merge-reports, and the tests/e2e gate jobs are gated off push, so a merge to main runs only build-images (cache warming) — not the full 8-shard e2e suite.
Per-PR cache retained. PRs still write a per-PR gha cache, so a second commit in the same PR reuses its own layers even for images it did change.
Fork behavior unchanged. Fork PRs still build locally and upload the image as an artifact (the :buildcache package stays private, so forks build cold — same as today).

Impact
First push to main after merge does one cold build (~7 min for Langflow) to seed :buildcache.
After that, a fresh PR off main that doesn't modify an image restores it straight from the shared cache — Langflow drops from ~7 min to ~1–2 min on the first run, with no per-PR warm-up needed.
Merges to main no longer pay for a full test/e2e run.
Notes / follow-ups
Cache freshness tracks main: if main bumps an image's deps, the next main push rebuilds that layer once (~7 min on main), and subsequent PRs inherit the warm layer.
scripts/ci/get_image_hash.py is no longer referenced by CI and can be removed in a follow-up if unused elsewhere.

Test plan
 Open this PR → confirm build-images runs 4 images in parallel and passes.
 Confirm test-suite + e2e run on the PR as before.
 Merge → confirm the main push runs only build-images (no e2e) and populates the openrag-*:buildcache packages.
 Open a follow-up PR off main that doesn't touch Langflow → confirm the Langflow build is a cache hit and finishes in ~1–2 min.

## Summary by CodeRabbit

* **Chores**
  * Updated CI image workflows to build, cache, pull, and publish images more efficiently.
  * CI image builds now run on pushes to the main branch, while heavier test and end-to-end jobs remain excluded.
  * Added retry handling when pulling prebuilt images.

* **Style**
  * Improved spacing in the environment configuration section.

* **Documentation**
  * Added an internal TODO comment for CI testing.

### DIFF
--- a/.github/workflows/test-ci.yml
+++ b/.github/workflows/test-ci.yml
@@ -2,6 +2,10 @@ name: CI Tests
 
 on:
   pull_request:
+  # Push to main only warms the shared layer cache (build-images); the test/e2e
+  # jobs are gated off `push` below so a merge doesn't run the full suite.
+  push:
+    branches: [main]
   workflow_dispatch:
     inputs:
       use_local_images:
@@ -73,7 +77,7 @@ jobs:
 
   build-images:
     needs: changes
-    if: ${{ needs.changes.outputs.e2e == 'true' || needs.changes.outputs.integration == 'true' || github.event_name == 'workflow_dispatch' }}
+    if: ${{ needs.changes.outputs.e2e == 'true' || needs.changes.outputs.integration == 'true' || github.event_name == 'workflow_dispatch' || github.event_name == 'push' }}
     runs-on:
       labels: "ubuntu-latest"
     permissions:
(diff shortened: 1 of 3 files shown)


lucaseduoli (approved):
Nice work!!

PR #2109, opened 2026-07-16
feat: Scope document chunk IDs by owner for isolation and updates

edwinjosechittilappilly:
This pull request improves how document chunk IDs are managed in the OpenSearch index to ensure that identical chunks uploaded by different owners do not conflict, and that upserts remain idempotent within each ownership scope. It introduces a new chunk ID scoping mechanism, updates the logic for deleting documents, and adds comprehensive unit tests to verify the new behavior.

**Ownership-scoped chunk IDs and upsert logic:**

* Added a `_scoped_chunk_id` method to `DocumentIndexWriter` that prefixes chunk IDs with a hash of the ownership scope, ensuring chunk upserts are isolated per owner and preventing cross-owner conflicts. [[1]](diffhunk://#diff-d131634fa6bbaeccb67906110fcbe7368a3e9ad5720a8afe1ce014b8bbfb8db6R152-R158) [[2]](diffhunk://#diff-d131634fa6bbaeccb67906110fcbe7368a3e9ad5720a8afe1ce014b8bbfb8db6L115-R123)
* Updated the `index_chunks` method to use the scoped chunk ID when indexing, so identical chunks from different owners do not overwrite each other.
* Updated documentation in `index_chunks` to clarify idempotency now applies within the same ownership scope.

**Ownership-aware document deletion:**

* Modified `delete_document_by_filename` calls to pass `owner_user_id`, ensuring that deletions are correctly scoped to the owner and do not affect documents belonging to other users.

**Testing and validation:**

* Added a new test module `test_document_index_writer_ownership.py` with in-memory OpenSearch mocks to verify that identical chunks remain available to each owner and that re-ingesting a chunk for the same owner or for shared documents updates it in place.
* Updated existing tests to expect the new chunk ID format (scoped with a hash) and to check that the correct parameters are used in deletion calls. [[1]](diffhunk://#diff-5e054a2de8ed81867c4e7dad1deac803eb84452d181043fb5e1a8050177a5251L135-R136) [[2]](diffhunk://#diff-e2995d5db9f6d58541fd2bde7fec93f74ee431163f55d6e8f212317165c2cae7L90-R94)
* Minor test import update for improved assertion matching.Make document chunk upserts ownership-aware so identical chunk IDs from different users do not overwrite each other. Update duplicate-delete flow to pass the owner context, and adjust tests to cover per-owner isolation plus in-place updates within the same scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate file replacement so existing documents are properly removed within the correct ownership scope.
  * Prevented documents from different ownership scopes from overwriting one another during indexing.
  * Preserved in-place updates when the same document is indexed again.

* **Tests**
  * Added coverage for ownership isolation, shared documents, re-ingestion, and duplicate replacement behavior.

### DIFF
--- a/src/services/document_index_writer.py
+++ b/src/services/document_index_writer.py
@@ -8,6 +8,7 @@ indexing chunks into the documents index.
 from __future__ import annotations
 
 import datetime
+import hashlib
 import json
 from dataclasses import dataclass, field
 from typing import Any
@@ -80,8 +81,8 @@ class DocumentIndexWriter:
     ) -> dict[str, Any]:
         """Index one batch of chunks.
 
-        Repeated calls with the same chunk ids are idempotent because the write
-        operation is an index/upsert.
+        Repeated calls with the same chunk ids in the same ownership scope are
+        idempotent because the write operation is an index/upsert.
         """
         from config.settings import get_index_name
 
@@ -112,7 +113,14 @@ class DocumentIndexWriter:
                     "Embedding dimension mismatch in batch: "
                     f"expected {dimensions}, got {len(chunk.vector)} for {chunk.chunk_id}"
                 )
-            bulk_body.append({"index": {"_index": index_name, "_id": chunk.chunk_id}})
+            bulk_body.append(
+                {
+                    "index": {
+                        "_index": index_name,
+                        "_id": self._scoped_chunk_id(context, chunk.chunk_id),
+                    }
+                }
+            )
             bulk_body.append(
                 self._build_chunk_document(
                     context=context,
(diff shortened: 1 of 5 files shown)


src/services/document_index_writer.py
```diff
@@ -141,6 +149,13 @@ async def index_chunks(
             "document_id": context.document_id,
         }
 
+    @staticmethod
+    def _scoped_chunk_id(context: DocumentIndexContext, chunk_id: str) -> str:
```
mpawlow, line 153: ### (1a) [Info] No remediation path for documents already corrupted by the pre-fix collision bug

#### Problem

- This PR stops **new** cross-owner collisions but does nothing for chunks that were already overwritten in a live deployment before this fix shipped.
- If owner A and owner B previously uploaded identical file content, A's document metadata (owner, ACLs, filename) may already have been silently replaced by B's, or vice versa — that data loss already happened and isn't reversible by this change alone.
- Once deployed, the fix prevents recurrence but gives no way to detect or repair documents that are currently mis-owned as a result of the bug.

#### Code References

- `src/models/processors.py:591-593` (`resolve_shared_owner_fields`) — confirms `owner` is the sole discriminator that was previously ignored in the chunk `_id`

#### Potential Solution

- ℹ️  Out of scope for this PR
- Maybe worth a tracked follow-up: a one-off audit/remediation script (matching the existing precedent of `scripts/migrate_embedding_model_field.py`) to identify documents whose current owner may not match the original uploader, for affected deployments.

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
- 1 observational / info PR comment: (1a) - related but orthogonal to this PR

mfortman11 (approved):

src/services/document_index_writer.py
edwinjosechittilappilly: @coderabbitai can you create an issue and also mentin to have a set of startup tasks for this migration meniton the plan above and also add a potential plan for the fix. and tag @mpawlow  and me

edwinjosechittilappilly: Thank you for the feedback.

PR #2113, opened 2026-07-16
feat: Migrate CPD to Universal Token

edwinjosechittilappilly:
This pull request refactors the handling of OpenSearch credentials and RBAC enforcement across SaaS, on-prem, and OSS deployment modes. The main change is to treat both SaaS and on-prem modes as requiring a platform service token (JWT) for OpenSearch administration and RBAC, removing prior fallback logic and clarifying error handling and logging. OSS mode continues to use OpenSearch basic authentication. This unifies and simplifies the authentication model, making requirements stricter and more predictable.

**OpenSearch client and credential handling:**
- Both SaaS and on-prem now require `OPENRAG_SERVICE_TOKEN` for OpenSearch admin and writer clients; missing the token raises a `RuntimeError` and logs a clear error message. OSS continues to use basic auth. [[1]](diffhunk://#diff-4d75b7c1443c4f51a4fa22fb59df90c20bf2ccf2c76a4f76fc66e17bb1f35dccR1105-R1122) [[2]](diffhunk://#diff-7256a966ec47d12af4482369f083f68d5a08f112439093b0d54f0864722d962fR264-R274) [[3]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690R896-L925) [[4]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L1569-L1611)
- The logic for creating OpenSearch clients is updated throughout the codebase to consistently treat SaaS/on-prem as requiring the service token, and OSS as using basic auth. [[1]](diffhunk://#diff-4d75b7c1443c4f51a4fa22fb59df90c20bf2ccf2c76a4f76fc66e17bb1f35dccR1105-R1122) [[2]](diffhunk://#diff-7256a966ec47d12af4482369f083f68d5a08f112439093b0d54f0864722d962fR264-R274) [[3]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L1569-L1611) [[4]](diffhunk://#diff-6a7ded7417c902e8e0d51d5008a58e007373b8f09a340320bc09ff3646fa4167L244-R245)

**RBAC enforcement and authentication:**
- RBAC enforcement is unified for SaaS and on-prem: in either mode, with RBAC enabled, the gateway must forward the end-user JWT on every request. If missing, the API fails fast with a 401 error and descriptive message, and does not fall back to basic credentials. [[1]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL292-R299) [[2]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL313-R347) [[3]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL495-R507) [[4]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL574-R587) [[5]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL597-R604) [[6]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL659-R666)
- All logging and error messages are updated to refer to "platform" (SaaS/on-prem) modes and to include the current run mode for clarity. [[1]](diffhunk://#diff-7256a966ec47d12af4482369f083f68d5a08f112439093b0d54f0864722d962fL296-R288) [[2]](diffhunk://#diff-7256a966ec47d12af4482369f083f68d5a08f112439093b0d54f0864722d962fL310-R311) [[3]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL313-R347) [[4]](diffhunk://#diff-1365801e155896d9da186ce9f0261ece6a9fdbbf11fd4fa69cdcab07b5c8a06dL574-R587)

**Code and documentation cleanup:**
- Removes legacy or backward-compatibility paths that allowed user JWTs or unauthenticated clients in SaaS/on-prem; these modes now always require the service token. [[1]](diffhunk://#diff-4d75b7c1443c4f51a4fa22fb59df90c20bf2ccf2c76a4f76fc66e17bb1f35dccL1130-R1134) [[2]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L1569-L1611)
- Updates docstrings and comments to reflect the new, unified platform authentication model. [[1]](diffhunk://#diff-6ff8f6f489c49cfc2185aa6bdee4b6ee87be730175eed7d6a67fc11194861690L1569-L1611) [[2]](diffhunk://#diff-6a7ded7417c902e8e0d51d5008a58e007373b8f09a340320bc09ff3646fa4167L244-R245)

## Summary by CodeRabbit

* **Authentication**
  * Platform RBAC now applies to both **SaaS** and **on-prem** with **fail-fast 401** responses when the forwarded user JWT is missing or invalid.
  * When RBAC is enabled, the gateway JWT path takes precedence and skips other credential fallback behavior.

* **Configuration**
  * **SaaS** and **on-prem** OpenSearch admin setup/index bootstrap now require **`OPENRAG_SERVICE_TOKEN`** and use a JWT-derived admin identity.
  * **OSS** continues to use OpenSearch basic authentication.

* **Bug Fixes**
  * Startup/security reconciliation warning messaging now reflects the active run mode.

* **Tests**
  * Updated and expanded unit coverage for RBAC and OpenSearch client selection across run modes.

### DIFF
--- a/src/api/settings/endpoints.py
+++ b/src/api/settings/endpoints.py
@@ -1102,38 +1102,36 @@ async def onboarding(
                 )
                 from main import init_index
                 from utils.run_mode_utils import (
+                    get_run_mode,
                     is_run_mode_on_prem,
-                    is_run_mode_oss,
                     is_run_mode_saas,
                 )
 
                 # Choose the admin identity for index/security setup, by run mode:
-                #   saas    -> platform service token (JWT); admin from its claim
-                #   on_prem -> OpenSearch basic auth; OpenSearch username as admin
-                #   oss     -> OpenSearch basic auth; OpenSearch username as admin
+                #   saas/on_prem -> platform service token (JWT); admin from its claim
+                #   oss          -> OpenSearch basic auth; OpenSearch username as admin
                 # The matching client comes from the shared run-mode-aware helper.
                 admin_username = None
-                if is_run_mode_saas():
+                if is_run_mode_saas() or is_run_mode_on_prem():
                     service_token = get_openrag_service_token()
-                    if service_token:
-                        # Prefer the platform service token so onboarding pins the
-                        # same admin identity as the startup security bootstrap
-                        # (see app/lifespan.py).
-                        from auth.ibm_auth import admin_username_from_service_jwt
+                    if not service_token:
+                        raise RuntimeError(
+                            "OPENRAG_SERVICE_TOKEN is required to determine the "
+                            "OpenSearch admin identity during onboarding in "
+                            f"{get_run_mode()} mode."
+                        )
+                    # Prefer the platform service token so onboarding pins the
+                    # same admin identity as the startup security bootstrap
+                    # (see app/lifespan.py).
+                    from auth.ibm_auth import admin_username_from_service_jwt
 
-                        admin_username = admin_username_from_service_jwt(service_token)
-                        if not admin_username:
-                            raise RuntimeError(
-                                "OPENRAG_SERVICE_TOKEN has no 'username' or 'sub' claim; "
-                                "cannot determine OpenSearch admin during onboarding"
-                            )
-                    elif user:
-                        # TODO: backward-compatibility fallback for saas deployments
-                        # without OPENRAG_SERVICE_TOKEN — pins the onboarding user as
-                        # admin instead of the platform service identity. Remove once
-                        # the service token is always provided.
-                        admin_username = user.user_id
-                elif is_run_mode_on_prem() or is_run_mode_oss():
+                    admin_username = admin_username_from_service_jwt(service_token)
+                    if not admin_username:
+                        raise RuntimeError(
+                            "OPENRAG_SERVICE_TOKEN has no 'username' or 'sub' claim; "
+                            "cannot determine OpenSearch admin during onboarding"
+                        )
+                else:
                     admin_username = get_opensearch_username()
 
                 opensearch_client = app_clients.create_index_admin_opensearch_client(
--- a/src/auth/request_identity.py
+++ b/src/auth/request_identity.py
@@ -289,14 +289,14 @@ async def _get_ibm_user(request: Request, required: bool) -> Optional["User"]:
                 # existing DB roles untouched). RBAC on -> extract + 401 if none.
                 _stage_jwt_roles(request, claims, user_id)
 
-    from utils.run_mode_utils import is_run_mode_saas
+    from utils.run_mode_utils import get_run_mode, is_run_mode_on_prem, is_run_mode_saas
 
-    saas_rbac = is_run_mode_saas() and jwt_roles_enabled()
+    platform_rbac = (is_run_mode_saas() or is_run_mode_on_prem()) and jwt_roles_enabled()
 
     # The forwarded end-user JWT is the OpenSearch credential — OpenSearch's
     # openid_auth_domain validates it via the backend JWKS. Build it once; it's
-    # used both as the saas+rbac authoritative credential and as the header-less
-    # fallback in other modes.
+    # used both as the platform+rbac authoritative credential and as the
+    # header-less fallback in other modes.
     jwt_user = None
     if ibm_token and user_id:
         jwt_user = User(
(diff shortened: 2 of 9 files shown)


src/auth/request_identity.py
mpawlow, line 227: ### (1a) [Major] `PLATFORM_USERNAME`/`PLATFORM_PASSWORD` env override bypasses mandatory platform RBAC JWT enforcement

#### Problem

- `src/auth/request_identity.py` returns a `User` built from `PLATFORM_USERNAME`/`PLATFORM_PASSWORD` (Basic-auth env vars) at the very top of `_get_ibm_user`, **before** `platform_rbac` is even computed.
- This PR's entire premise is that SaaS/on-prem + RBAC must always be authenticated by a forwarded end-user JWT, and any missing/invalid JWT must fail loud (401) with no fallback.
- The `PLATFORM_USERNAME`/`PLATFORM_PASSWORD` override completely sidesteps that contract: if those env vars are set, the request authenticates via static Basic credentials regardless of RBAC mode, run mode, or whether a JWT was forwarded at all.

#### Background Information

- Concur with similar feedback provided by CodeRabbit.ai here

#### Code References

- `src/auth/request_identity.py:294` (`platform_rbac` computed after the override already returned)

#### Potential Solution

- Compute `platform_rbac` before the `PLATFORM_USERNAME`/`PLATFORM_PASSWORD` check, and skip/disable the override whenever `platform_rbac` is `True`, so the mandatory-JWT contract holds uniformly.

#### Alternative Solutions

- If the override is intentionally meant as a break-glass/local-dev mechanism that should never be reachable in a platform-RBAC deployment, gate it explicitly behind `not platform_rbac` and add a startup-time warning/assertion if both are configured simultaneously, to make the precedence auditable.

src/services/dls_principal_service.py
mpawlow, line 298: ### (1b) [Major] `RuntimeError` for missing `OPENRAG_SERVICE_TOKEN` is silently swallowed in DLS principal sync

#### Problem

- `src/services/dls_principal_service.py::_get_opensearch_client` now raises `RuntimeError` when `OPENRAG_SERVICE_TOKEN` is missing in SaaS/on-prem mode — matching the "required, fail loud" pattern used in `lifespan.py`, `config/settings.py`, and `endpoints.py`.
- However, that `raise` sits inside a `try:` block whose enclosing `except Exception as e:` (lines ~295-301) catches it, logs a `logger.warning`, and returns `None`.
- Net effect: the "fail loud" intent stated in the updated comment ("SaaS/on-prem -> platform service-token JWT (required)") is not actually enforced here — DLS principal synchronization instead degrades silently, same as the old behavior this PR was meant to replace.
- Verified: `tests/unit/services/test_dls_principal_client_selection.py::test_saas_without_service_token_returns_none` (unchanged by this PR) still passes, which masks the discrepancy — the test asserts `None` is returned, not that a `RuntimeError` occurs.

#### Background Information

- Concur with the corresponding CodeRabbit.ai feedback

#### Code References

- `src/services/dls_principal_service.py:264-270` (raise, now caught)
- `tests/unit/services/test_dls_principal_client_selection.py:54-62` (`test_saas_without_service_token_returns_none`, now stale/misleading given the new intent)

#### Potential Solution

- Re-raise (or don't catch) the missing-token `RuntimeError` specifically, letting it propagate as a hard failure consistent with the other three call sites; keep the broad `except Exception` only around the actual client-construction call, not the credential-selection logic.
- Update `test_saas_without_service_token_returns_none` (and add an on-prem mirror) to assert `RuntimeError` is raised instead of `None` returned.

src/config/settings.py
```diff
@@ -906,36 +906,32 @@ def __init__(self):
...
-        #   saas    -> platform service token (JWT) when available, else unauthenticated
-        #   on_prem -> OpenSearch basic auth
-        #   oss     -> OpenSearch basic auth
-        service_token = get_openrag_service_token() if is_run_mode_saas() else None
-        if service_token:
+        #   saas/on_prem -> platform service token (JWT); required, raises if unset
+        #   oss          -> OpenSearch basic auth
+        if is_run_mode_saas() or is_run_mode_on_prem():
+            service_token = get_openrag_service_token()
+            if not service_token:
```
mpawlow, line 919: ### (1c) [Info][Question] New unconditional hard failure on startup if `OPENRAG_SERVICE_TOKEN` is unset for SaaS/on-prem

#### Potential Problem

- `Clients.initialize()` (`src/config/settings.py:907-928`) now unconditionally raises `RuntimeError` for any SaaS/on-prem deployment missing `OPENRAG_SERVICE_TOKEN`, and this is called directly (not conditionally) from `initialize_services()` in `src/app/container.py:73`, which re-raises after logging — i.e., the backend process fails to start entirely.
- Previously, SaaS without a service token fell back to an unauthenticated client (with a warning), so this is a behavior change from "degrades but boots" to "cannot boot."
- This differs from the more surgical `app/lifespan.py` treatment, where a missing token is only fatal when security bootstrap (`OPENRAG_BOOTSTRAP_OS_SECURITY_ON_STARTUP`-style logic) is explicitly enabled — that path is more conservative.

#### Code References

- `src/app/container.py:71-79` (`initialize_services`, re-raises on failure)

#### Questions

- ❓ `Question`: Just to confirm, is this new hard failure behavior intended?

mpawlow (changes requested):
## Code Review 1

- See Major PR comments: (1a) (1b) for evaluation
- See Question in PR comment (1c)

src/auth/request_identity.py
edwinjosechittilappilly: This is a predated change, The current PR focus on enabling the RBAC for the on prem mode.
This mode using platform username and password is completeley for Dev Testing , I am thinking if we can remove this mode with platform user name and password since RBAC is from the JWT and only in saas and onprem mode.

@coderabbitai thoughts?

src/services/dls_principal_service.py
edwinjosechittilappilly: Good Catch Will look into this shortly.

@coderabbitai, suggest potential possibilities of fixes.

src/config/settings.py
edwinjosechittilappilly: Yes! Without the service token the saas and onprem should fail since its a pre requisite for lot of default operations and dangerous to proceed without it.

@coderabbitai add this to the learning and reviews wrt to the startup

src/auth/request_identity.py
edwinjosechittilappilly: https://github.com/langflow-ai/openrag/pull/2136 to be merged after this PR #2113 merges

src/config/settings.py
mpawlow: Okay, makes sense

src/api/settings/endpoints.py
```diff
@@ -1104,38 +1104,36 @@ async def onboarding(
...
-                            )
-                    elif user:
-                        # TODO: backward-compatibility fallback for saas deployments
-                        # without OPENRAG_SERVICE_TOKEN — pins the onboarding user as
-                        # admin instead of the platform service identity. Remove once
-                        # the service token is always provided.
-                        admin_username = user.user_id
-                elif is_run_mode_on_prem() or is_run_mode_oss():
+                    if not service_token:
+                        raise RuntimeError(
```
mpawlow, line 1120: ### (2a) [Minor] `create_index_admin_opensearch_client`'s None→RuntimeError contract change removes graceful degradation for two upload endpoints

#### Problem

- This PR changed `Clients.create_index_admin_opensearch_client()` (`src/config/settings.py:1574-1605`) from "return `None` when no suitable admin credentials exist, let the caller fall back to the global writer client" to "raise `RuntimeError` when `OPENRAG_SERVICE_TOKEN` is missing in saas/on_prem mode."
- The PR's own primary caller (`src/api/settings/endpoints.py` onboarding handler) was updated to catch this correctly — the whole block is wrapped in `try/except Exception` and returns a clean `JSONResponse(..., status_code=500)`.
- However, `create_index_admin_opensearch_client()` has a second call path: `src/api/documents.py::_ensure_index_exists()` → called from 5 places. Of those, 3 (`documents.py::check_filename_exists`, `router.py::_traditional_upload_ingest_task`, `connectors.py::connector_sync`) have an enclosing `try/except Exception` that degrades gracefully to a JSON error response.
- Two do **not**: `src/api/upload.py::upload_path` and `src/api/upload.py::upload_bucket` call `await _ensure_index_exists(jwt_token)` with no surrounding exception handling at all. A `RuntimeError` here propagates unhandled out of the FastAPI route (no `add_exception_handler` is registered anywhere in `app/factory.py`/`main.py`), producing a generic unhandled-exception 500 instead of a clean JSON error body — and, more importantly, losing the previous silent-fallback-to-global-writer-client behavior these two endpoints relied on.

#### Code References

- `src/api/upload.py:68-104` (`upload_path`, no exception handling around `_ensure_index_exists`)
- `src/api/upload.py:177-225` (`upload_bucket`, same gap)
- `src/api/documents.py:240-249` (`_ensure_index_exists`, the shared helper)

#### Potential Solution

- Wrap the bodies of `upload_path` and `upload_bucket` in the same `try/except Exception` pattern already used by `upload()` (`src/api/upload.py:35-58`) and `connector_sync`, returning a clean `{"error": ...}` JSON 500.

#### Alternative Solutions

- Keep `create_index_admin_opensearch_client()` raising (it's the right contract going forward), but audit and update all 5 call sites of `_ensure_index_exists` in one pass rather than patching individually, to guarantee no caller is left depending on the old `None`-fallback contract.

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM  🚀 
- See Minor (nit) PR comment: (2a) for optional consideration
- Note: (1a) deferred to https://github.com/langflow-ai/openrag/pull/2136

src/api/settings/endpoints.py
edwinjosechittilappilly: @coderabbitai can you investigate and create an issue with the plan for the above and assign it to me

PR #2115, opened 2026-07-18
feat: ship ingest review UI with settings, Docling preview, and chunk search

Wallgau:
Summary

Add the ingest review dialog: Docling document view (or DocPage skeleton while loading), index pipeline steps, file-scoped searchable chunks, expand fill layout, and header search / auto-open footer.

Add Ingest preview settings (auto-open Every / First run / Never, chunk/pipeline toggles, sample ingest) gated by OSS/SaaS run mode, with prefs shared via useIngestPreviewSettings.

Reuse FileChunksPanel + controlled search on /knowledge/chunks and in review so paste search filters locally after a file-scoped load.
Wire preview=true uploads from knowledge; extend preview backend for Docling page images / index-proof and related tests.

Test plan
1.  OPENRAG_INGEST_PREVIEW_ENABLED=true + OPENRAG_RUN_MODE=oss (or saas); 
--follow onboarding steps, upload a doc--
and/or
2. upload a PDF and confirm review auto-opens per setting
3.  Confirm Docling skeleton → annotated pages, index steps complete, chunk click highlights page
4.  Paste chunk text into header search and confirm in-panel filter (not empty “No knowledge”)
5.  Expand/collapse: body grows; title/toolbar/footer stay compact
6.  Folder upload: file dropdown switches files; search resets on file change
7.  Settings → Ingest preview: change auto-open / toggles; run sample ingest
8.  /knowledge/chunks?filename=…: same chunk search UI and filter behavior
9.  Auto-open Never / First run only / Every upload behave as expected after close + re-upload

## Summary by CodeRabbit

* **New Features**
  * Added ingest preview/review dialog with Docling parsed previews (including page/image rendering when available) and indexing pipeline progress.
  * Added ingest-preview settings (new tab, auto-open policy, and “Run a sample ingest”).
  * Enhanced chunk browsing with file-scoped chunk loading and optional highlighting/search.
* **Bug Fixes**
  * Improved preview retrieval behavior for missing/expired data and disabled preview mode; safer polling completion/error handling.
* **Tests**
  * Added unit coverage for parse-preview endpoint, Docling preview caching, layout summary counts, and per-task cache limits.


frontend/components/docling-preview.tsx
```diff
@@ -0,0 +1,1033 @@
...
+        if (pageAnchor) {
+          finish([pageAnchor], false);
+          return;
+        }
+      }
+
+      if (attempt < 18) {
+        schedule(() => place(attempt + 1), 32);
+      }
+    };
```
mfortman11: Where do these number values come from 6,10,18,32 if these are specific values let's define constants for them to help others understand what they are and why we chose them

frontend/components/docling-preview.tsx
mfortman11, line 1: This file is quite large. It would be great to split it up some.
Maybe pull out types or util function

frontend/components/docling-preview.tsx
```diff
@@ -0,0 +1,1033 @@
...
+        (beforeSignature === "" || signature !== beforeSignature);
+
+      if (hitsAreFresh) {
+        finish(hitEls, true);
+        return;
+      }
+
+      // Still the previous selection's parts — wait briefly for Lit.
+      if (hitEls.length > 0 && attempt < 10) {
+        schedule(() => place(attempt + 1), 32);
```
mfortman11: **Severity**: Low (efficiency)

**Issue**: The `place()` retry loop schedules up to 18 sequential 32ms timers to poll for Lit Shadow DOM geometry updates. Rapid chunk selection starts a new 18-step cascade each time (the previous one is cancelled, but the pattern is wasteful). The stale-hit branch here retries up to 10 times at 32ms, then falls through to a second retry path that goes up to attempt 18 — totaling ~576ms of busy polling.

**Suggested fix**: Reduce max attempts from 18 to ~10, use exponential backoff (32ms for early attempts, 64ms for later), and lower the page-fallback threshold from attempt 6 to 4. Same total wait budget with fewer timer allocations. Alternatively, if the `docling-img-page` Lit element exposes `updateComplete`, awaiting that promise would eliminate the polling entirely.

frontend/components/knowledge-dropdown.tsx
```diff
@@ -370,44 +448,103 @@ export function KnowledgeDropdown() {
...
+    const taskIdsByBatch: (string | undefined)[] = [];
+    const failedBatchIndexes: number[] = [];
+    // Cap parallel batch uploads so large folders don't open every request at once.
+    const batchConcurrency = Math.min(2, batches.length);
+    let nextBatchIndex = 0;
+    // Shared across concurrent workers: a previewMode=false batch must not
+    // close the dialog after a sibling batch already confirmed preview.
+    let previewModeConfirmed = false;
+
+    const runNextBatch = async () => {
```
mfortman11, line 467: **Severity**: Medium (correctness)

**Issue**: `uploadFolderBatches` shares a `let previewModeConfirmed` flag between 2 concurrent runNextBatch workers (`knowledge-dropdown.tsx:465`). The interleaving depends on network timing: Worker A's fetch can resolve with previewMode=false, read `previewModeConfirmed` as `false`, and call `setPreview(EMPTY_PREVIEW)` — closing the dialog — right before Worker B's fetch resolves with `previewMode=true`. The comment on line `464` acknowledges this ("a concurrent false must not wipe a successful open") but the guard doesn't fully prevent it since both workers race through the same if/else between awaits.

**Suggested fix**: Remove the per-batch `setPreview(EMPTY_PREVIEW)` call from inside the worker loop entirely. Instead, after `Promise.all` resolves, check whether any batch confirmed preview mode and only close the dialog at that point:

```
await Promise.all(
  Array.from({ length: batchConcurrency }, () => runNextBatch()),
);

if (!previewModeConfirmed) {
  setPreview(EMPTY_PREVIEW);
}
```

This makes the close decision a single synchronous check after all concurrent work is done, eliminating the race.

mfortman11 (dismissed):
Image

I was getting blocked on this screen the preview seemed to be stuck but the ingest completed

frontend/components/docling-preview.tsx
Wallgau: those numbers are empirical wait budgets, let's remove those arbitrary values and await updateComplete instead

Wallgau:
> Image
> I was getting blocked on this screen the preview seemed to be stuck but the ingest completed

The preview skeleton got stuck because we stopped polling after ~90s and cached a null (404 while Docling was still running) forever, even after ingest finished. Fix: refetch when the task phase becomes ready, and show an unavailable state with Retry instead of an infinite skeleton. Pushing the fix in few minutes

lucaseduoli (changes requested):
Image

The description is inconsistent with the buttons, and I personally didn't understand quite well what each button does...

Image

You can put a <div> in the text of this button so that it doesn't get affected by the button's class that capitalizes every word

Image

Big names are overflowing the chunks horizontally

And the three last steps are being shown at the same time for me. Maybe we can show them briefly for a sec before moving to the next even though they're completed at the same time

lucaseduoli (approved):

PR #2120, opened 2026-07-20
fix: pass index name at runtime to langflow

lucaseduoli:
This pull request updates the way OpenSearch configuration is injected into Langflow headers to ensure flows always use the correct endpoint and index name. The most important changes are:

**OpenSearch Configuration Injection:**

* The OpenSearch index name is now included in the headers by importing and using the `get_index_name` function from `config.settings`, and setting the `X-LANGFLOW-GLOBAL-VAR-OPENSEARCH_INDEX_NAME` header if an index name is available.
* The import statement for OpenSearch settings now includes `get_index_name` in addition to `LANGFLOW_OPENSEARCH_HOST` and `LANGFLOW_OPENSEARCH_PORT`.

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenSearch requests now include the configured index name when available.
  * Preserved support for setting the OpenSearch connection URL.


edwinjosechittilappilly (approved):
LGTM

PR #2123, opened 2026-07-20
fix: tenant ID and remove unused components

edwinjosechittilappilly:
This pull request updates authentication handling for the Docling service and makes a small change to the ingestion flow configuration. The main focus is on improving how authentication headers are set based on the deployment mode, making the logic more robust and less dependent on a single environment variable.

Authentication logic improvements:
* Replaced the use of `IBM_AUTH_ENABLED` with checks for `is_run_mode_saas()` and `is_run_mode_on_prem()` to determine when to add authentication headers in `src/api/docling.py`. [[1]](diffhunk://#diff-9b2e0d4eecb17971a285069e8086cf8c710711e7d67bbcab7ba4cd1a77e3c0adL11-R15) [[2]](diffhunk://#diff-9b2e0d4eecb17971a285069e8086cf8c710711e7d67bbcab7ba4cd1a77e3c0adL34-L38)
* Removed the addition of the `X-Tenant-Id` header in the health check endpoint, streamlining the headers sent to the Docling service.

Configuration change:
* Set the `X-Tenant-Id` value to an empty string in the `flows/ingestion_flow.json` configuration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docling service health checks across supported runtime modes.
  * Authorization headers are now forwarded only when a valid token is available.
  * Removed unnecessary tenant header forwarding from health-check requests.

### DIFF
--- a/flows/ingestion_flow.json
+++ b/flows/ingestion_flow.json
@@ -666,7 +666,7 @@
                       "value": true
                     },
                     "key": "X-Tenant-Id",
-                    "value": "OWNER"
+                    "value": ""
                   }
                 ]
               },
--- a/src/api/docling.py
+++ b/src/api/docling.py
@@ -8,11 +8,11 @@ from config.settings import (
     DOCLING_HOST_IP,
     DOCLING_SERVE_URL,
     DOCLING_SERVE_VERIFY_SSL,
-    IBM_AUTH_ENABLED,
 )
 from dependencies import get_optional_user
 from session_manager import User
 from utils.logging_config import get_logger
+from utils.run_mode_utils import is_run_mode_on_prem, is_run_mode_saas
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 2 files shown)


flows/ingestion_flow.json
```diff
@@ -666,7 +666,7 @@
                       "value": true
                     },
                     "key": "X-Tenant-Id",
-                    "value": "OWNER"
+                    "value": ""
```
mpawlow, line 669: ### (2a) [Normal] Recommend removing the entire"X-Tenant-Id" HTTP header if possible

#### Problem

- May cause downstream parsing / validation problems sending the blank value
- Concur with CodeRabbit.AI comment below

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM  🚀 
- See non-blocking PR comment (2a) for consideration

flows/ingestion_flow.json
edwinjosechittilappilly: Yes thats the next step, We need a langflow component Update, in discussion with @lucaseduoli already on that step.

This PR is to un block the CPD Follow up PR require to remove tenant id @coderabbitai create an issue for removing tenant id from docling serve langflow component and assign it to @lucaseduoli

zzzming (approved):

PR #2131, opened 2026-07-21
feat: langflow 1.11.1 upgrade

lucaseduoli:
This pull request introduces several improvements across the development workflow, Docker images, and components, focusing on developer experience, database reliability, and data handling. The most notable changes are the addition of new Makefile targets for flexible local development, enhancements to the Dockerfile for Langflow, improved SQLite database configuration for concurrency, and updates to component interfaces for better downstream compatibility.

**Development workflow and infrastructure:**

* Added new Makefile targets (`dev-branch-local`, `dev-branch-local-cpu`) for starting local development infrastructure with custom Langflow branches, supporting both GPU and CPU-only setups. These enable running backend/frontend on the host while using containerized OpenSearch and Langflow, improving iteration speed and flexibility. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L108-R108) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R270-R272) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R527-R558) [[4]](diffhunk://#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L174-R185)
* Updated Makefile help output and documentation to reflect new development commands and usage patterns. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R270-R272) [[2]](diffhunk://#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L174-R185)

**Docker and environment improvements:**

* Refactored `Dockerfile.langflow` to use the upstream Langflow base image, simplified dependency installation, enabled custom components by default, and removed unnecessary user/group manipulations for improved build speed and maintainability.
* Increased frontend build memory allocation in `Dockerfile.langflow.dev` for more reliable builds on modern systems.

**Database reliability and configuration:**

* Configured SQLite to use WAL (Write-Ahead Logging) mode and set a busy timeout in the `langflow-entrypoint.py` script, improving database concurrency and robustness in containerized environments.
* Updated `docker-compose.yml` to set a longer SQLite timeout and connection pool settings, and exposed additional environment variables for rate limiting and custom component support. [[1]](diffhunk://#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3L181-R182) [[2]](diffhunk://#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R215-R223)

**Component and data handling improvements:**

* Changed the `search_documents` method in `opensearch_multimodal.py` to return a list of `Data` objects instead of a DataFrame, preventing string truncation and improving compatibility with downstream components. [[1]](diffhunk://#diff-e68e1cd62cf086d2f2f2ca544c9705121233caa5696aed34694347b8a2d8e4b8L2411-R2423) [[2]](diffhunk://#diff-e68e1cd62cf086d2f2f2ca544c9705121233caa5696aed34694347b8a2d8e4b8L2440-R2445)
* Refactored the `export_docling_document.py` component to use utility functions for document coercion and image mode handling, streamlined metadata extraction, and improved error handling for export operations. [[1]](diffhunk://#diff-6b925b6ed5d63aae5c201e25b93bc607ef355acaa5dee6578455eff1c3cc0968L3-R3) [[2]](diffhunk://#diff-6b925b6ed5d63aae5c201e25b93bc607ef355acaa5dee6578455eff1c3cc0968L88-R91) [[3]](diffhunk://#diff-6b925b6ed5d63aae5c201e25b93bc607ef355acaa5dee6578455eff1c3cc0968L147-R131)

**Frontend improvements:**

* Ensured chat messages are always sorted chronologically in the chat UI by explicitly sorting by timestamp, fixing potential ordering issues.

## Summary by CodeRabbit

* **New Features**
  * Added provider-aware global headers and expanded per-node model/provider overrides for chat and ingestion flows.
  * Enhanced OpenRAG flows with structured agent responses and improved vector-store ingestion/search behavior.
  * Added new local custom-branch development targets (including CPU-only).
* **Bug Fixes**
  * Conversation messages now load in chronological order.
  * Improved tool-call/retrieval parsing for more consistent outputs and citations.
  * Increased stability of Langflow startup (SQLite WAL/busy timeout) and container behaviors (rate limits/timeouts, SSRF allowlist).
* **Documentation**
  * Expanded custom Langflow branch development instructions.
* **Tests**
  * Added unit coverage for provider mapping, Langflow global-variable sync, and chunk parsing.


edwinjosechittilappilly:
@coderabbitai can you explaing the changes made to the docling and openserach multimodela components

lucaseduoli:
> @coderabbitai can you explaing the changes made to the docling and openserach multimodela components

I just updated it to the updated component in Langflow, because it seems like it fixed the previous problems with VLM

flows/openrag_agent.json
mpawlow, line 1: ### (1b) [Critical] Hardcoded OpenSearch password still committed in `flows/openrag_agent.json`

#### Problem

- The `password` `SecretStrInput` on the embedded OpenSearch node carries a literal credential value instead of the `OPENSEARCH_PASSWORD` global-variable placeholder used everywhere else in the same file.
- Committing a real secret to source control leaks it into git history and any distributed copy of the flow (clones, forks, CI artifacts), independent of the current `auth_mode`.
- CodeRabbit flagged this exact issue as **Critical** on 2026-07-21 (review `4749094870`) with a proposed diff to reset the value to `"OPENSEARCH_PASSWORD"`. It was not applied — the literal value is still present in the current HEAD of this branch.

#### Background Information

- Every other `password` field of type `SecretStrInput` in this file (e.g. line 1096, 2477, 3103, 3983) correctly uses `load_from_db`/global-variable indirection rather than a literal value — this one node is the outlier.
- Given the value looks like a real generated password (mixed case, symbols), it should be treated as compromised and rotated regardless of whether it fixes the file.

#### Code References

- `flows/openrag_agent.json:2683-2700` (the `password` input block, `"value": "8$8@SqVR&ZvbGBrd"` at line 2700)

#### Potential Solution

- Replace the literal value with `"OPENSEARCH_PASSWORD"` to match the pattern used elsewhere in the same file.
- Rotate the exposed OpenSearch password in any environment where it was ever valid, and scrub it from git history if it was ever a real production/shared credential.

#### Alternative Solutions

- Add a pre-commit check (e.g. extend the existing `detect-secrets` baseline scan to include `flows/*.json`, if not already covered) so a literal secret value in an embedded flow node can't silently pass CI again.

flows/ingestion_flow.json
```diff
@@ -2867,602 +3152,694 @@
...
                 "trace_as_metadata": true,
                 "track_in_telemetry": true,
-                "type": "int",
-                "value": 1000
+                "type": "str",
+                "value": "lowercase"
               },
-              "number_of_results": {
-                "_input_type": "IntInput",
+              "code": {
```
mpawlow, line 3459: ### (2b) [Major] `flows/ingestion_flow.json`'s `Operations` node had entirely wrong code embedded 

#### Problem

- Diffing the "fix flows with correct code" commit reveals that `Operations-VUCJC` in `flows/ingestion_flow.json` previously had the **`OpenSearchVectorStoreComponentMultimodalMultiEmbedding` component's Python source pasted into it** — i.e. a node declared/wired in the flow graph as a generic "Operations" (JSON/Table/Text) transform was actually carrying the class definition for an unrelated OpenSearch vector-store component (107,427 chars of the wrong code).
- This class/metadata mismatch: Langflow failing to parse/register this node at flow-load time would plausibly abort loading the whole `ingestion_flow.json` file
- The commit replaces the wrong code with a legitimate ~55KB "Unified Operations Component" implementation (Text/JSON/Table operations), which is structurally consistent with what an `Operations` node should contain.

#### Background Information

- The other three flows' `OpenSearchVectorStoreComponentMultimodalMultiEmbedding` node code was also synced in this same commit and is now byte-identical across all four flow files (`ingestion_flow.json`, `openrag_agent.json`, `openrag_nudges.json`, `openrag_url_mcp.json`) 

#### Code References

- `flows/ingestion_flow.json:3459-3476` (node `Operations-VUCJC`, `template.code.value` — line 3459 is the `"code": {` field, line 3476 is the `"value"` line carrying the entire embedded Python source)

#### Potential Solution

- Once CI confirms green, add a lightweight guard (e.g. a unit/smoke test that loads each `flows/*.json`, extracts every node's declared component `type`/`base_classes`, and asserts the embedded `code` actually defines a class matching that declaration) so a copy-paste mismatch like this can't silently ship again without a component-level assertion failing fast, rather than only manifesting as a 404 deep in e2e/integration CI.

flows/openrag_url_mcp.json
mpawlow, line 1: ### (1d) [Major] Multi-page URL crawls corrupt per-document `filename`/`mimetype` metadata

#### Problem

- In `flows/openrag_url_mcp.json`, `ParserComponent-1eim1` collapses all page titles from a multi-page crawl into a single newline-joined message, and the downstream `Operations-AJiaz` node then broadcasts that single joined value to every row's `filename`/`mimetype` column instead of keeping titles row-aligned per document.
- With `max_depth > 1` (i.e. any crawl that follows more than one page), every ingested chunk ends up tagged with the same combined string instead of its own page's title — and per CodeRabbit's second review pass, this same misrouting also overwrites the `mimetype` column with page-title text instead of an actual MIME type.
- This breaks OpenSearch's `document_types` faceting/filtering and any UI that relies on per-file `filename` metadata for multi-page URL-sourced content.

#### Background Information

- CodeRabbit flagged this as Major/"Heavy lift" on 2026-07-21; the flagged nodes (`ParserComponent-1eim1`, `Operations-AJiaz`) are still present and unchanged in the current flow.
- This is scoped to the URL/MCP ingestion connector specifically (`flows/openrag_url_mcp.json`), and only manifests once a crawl spans more than one page — single-page URL ingests are unaffected. 

#### Code References

- `flows/openrag_url_mcp.json:179-206` (node wiring: `ParserComponent-1eim1.parsed_text` → `Operations-AJiaz.new_column_value`)

#### Potential Solution

- Keep the per-page title row-aligned: route it directly into the filename/title column per document rather than through the parser's collapsed single-message output, and leave `mimetype` on its component default rather than driving it from `parsed_text`.

src/agent.py
```diff
@@ -417,14 +352,33 @@ async def async_response_stream(
...
+
+                    # Also check for direct tool_call chunks
+                    elif chunk_data.get("type") == "tool_call" and "results" in chunk_data:
+                        from utils.langflow_utils import parse_knowledge_chunks
+
+                        chunk_data["results"] = parse_knowledge_chunks(chunk_data["results"])
+
                 # Middleware: Detect implicit tool calls and inject standardized events
                 # This helps Granite 3.3 8b and other models that don't emit standard markers
                 if isinstance(chunk_data, dict) and not detected_tool_call:
```
mpawlow, line 372: ### (1e) [Normal] `agent.py` retrieval-chunk detection loosened to accept non-list `results`

#### Problem

- In `async_response_stream`, the implicit-tool-call detection previously required `"results" in chunk_data and isinstance(chunk_data.get("results"), list)` before treating a chunk as containing retrieval results. The refactor dropped the `isinstance(..., list)` check, now triggering on `"results" in chunk_data` alone.
- A chunk where `results` is present but `None` (or a non-list type) will now spuriously satisfy `has_results`, triggering injection of a synthetic `Retrieval` tool-call event with an empty/near-empty `results` list (since `parse_knowledge_chunks(None)` safely returns `[]`).
- Functionally low-impact (`parse_knowledge_chunks` guards against bad input), but it does widen when synthetic events get injected into the stream, which could confuse frontend tool-call-trace rendering with a spurious empty "Retrieval" step for models/payloads that include a `results: null` key for unrelated reasons.

#### Code References

- `src/agent.py:376-385` (`has_results = any([...])` check inside `async_response_stream`, under the "Middleware: Detect implicit tool calls" comment; line 380 is the specific `"results" in chunk_data` condition)

#### Potential Solution

- Restore a type guard, e.g. `isinstance(chunk_data.get("results"), (list, dict, str))`, or check truthiness (`chunk_data.get("results")`) rather than mere key presence, before treating a chunk as carrying retrieval results.

src/agent.py
```diff
@@ -417,14 +352,33 @@ async def async_response_stream(
...
+                # Format native tool call results
+                if isinstance(chunk_data, dict):
+                    # Format: {"type": "response.output_item.added", "item": {"type": "tool_call", "results": ...}}
+                    item = chunk_data.get("item")
+                    if (
+                        isinstance(item, dict)
+                        and item.get("type") == "tool_call"
+                        and "results" in item
+                    ):
+                        from utils.langflow_utils import parse_knowledge_chunks
```
mpawlow: ### (1f) [Minor] `agent.py` repeats the same function-local import four times

#### Problem

- `from utils.langflow_utils import parse_knowledge_chunks` is imported inline, mid-function, at four separate call sites within `src/agent.py` (inside `async_response_stream` twice, and inside `async_langflow_chat`), rather than once at module scope alongside the existing top-level `from utils.langflow_utils import strip_untrusted_fence_recursive`.
- Purely a code-quality/consistency nit — no functional issue — but it's inconsistent with the sibling import from the same module already hoisted to the top of the file, and repeats the same statement needlessly.

#### Code References

- `src/agent.py:364, 370, 392, 811` (`from utils.langflow_utils import parse_knowledge_chunks` — lines 364, 370, 392 inside `async_response_stream`; line 811 inside `async_langflow_chat`)

#### Potential Solution

- Move `parse_knowledge_chunks` into the existing top-level import from `utils.langflow_utils` and drop the four inline imports.

flows/openrag_nudges.json
```diff
@@ -4090,11 +3375,10 @@
             ],
             "frozen": false,
             "icon": "type",
-            "last_updated": "2026-02-27T18:44:23.265Z",
-            "legacy": false,
-            "lf_version": "1.8.0",
+            "last_updated": "2026-07-17T19:15:11.463Z",
+            "legacy": true,
```
mpawlow, line 3379: ### (1g) [Minor] `openrag_nudges.json` still wired through a deprecated legacy component

#### Problem

- The embedded `TextInput` component used as a passthrough for `OPENRAG-QUERY-FILTER` into the OpenSearch node's `filter_expression` sets `legacy = True` with `replacement = ["input_output.ChatInput"]` in its own component metadata.
- Since upstream Langflow may drop legacy components in a future release, this flow will need migrating to the recommended replacement before the next Langflow upgrade, or it risks silently breaking on update.

#### Background Info

- CodeRabbit flagged this as Trivial on 2026-07-21; still present, unaddressed (reasonably deprioritized given it's not currently broken).

#### Code References

- `flows/openrag_nudges.json:3340-3480`

#### Potential Solution

- Track this as a fast-follow: migrate the `OPENRAG-QUERY-FILTER` passthrough onto `input_output.ChatInput` (or Langflow's then-current recommended replacement) ahead of the next Langflow version bump, rather than blocking this PR on it.

mpawlow (commented):
## Code Review 1 + 2

- ⚠️ See PR comments (1b), (2b), (1d), (1e) for evaluation
- See Minor PR comments (1f), (1g) for optional consideration

## Functional Review 1

- 🚧 Still Pending

flows/ingestion_flow.json
lucaseduoli: this was already fixed

flows/openrag_nudges.json
lucaseduoli: we can't use chat input since it only allows one component at a time. Langflow team told me to continue using TextInput

mpawlow:
## Functional Review 1

- ✅ Langflow / flows sanity test scenarios passed
- ❌ URL ingestion flow does not work for me
   - Prompt: "Ingest this URL: https://ontheline.trincoll.edu/images/bookdown/sample-local-pdf.pdf"

image


- ✅ URL ingested worked for non-PDF target:
   - Prompt: `Ingest this URL: https://www.gutenberg.org/cache/epub/79225/pg79225.txt`

image

flows/openrag_url_mcp.json
mpawlow, line 1: ### (3a) [Info] URL-ingestion flow cannot ingest PDF (or any non-HTML) URLs — pre-existing gap, confirmed by live user testing on this PR

#### Problem

- Ingesting `https://ontheline.trincoll.edu/images/bookdown/sample-local-pdf.pdf` via chat fails with "the URL ingestion component appears outdated, and the flow could not load the document," while a non-PDF URL (a Gutenberg `.txt` link) ingests successfully.
- Root cause: `URLComponent-lnA0q`'s embedded code (`flows/openrag_url_mcp.json`) has no content-type branching. `_process_response()` always does `html_content = response.text` and feeds it to `BeautifulSoup` regardless of the response's `content-type` header — there's no path that routes binary/PDF content through Docling (unlike the file-upload ingestion path, which is Docling-aware per `frontend/lib/supported-file-types.ts` / `docling_remote.py`).
- For a single-URL (non-crawl) request, if nothing usable is extracted, `fetch_url_contents()` raises `"No documents were successfully loaded from any URL"` — the exact failure the user observed.

#### Background Information

- **Not a regression introduced by this PR**: diffing `URLComponent-lnA0q`'s code at the `main` merge-base (`bb95570e`) shows the same absence of PDF/binary handling — this gap predates the Langflow 1.11 upgrade.
- However, it's directly relevant to this review because (a) this PR heavily rewires this exact flow (the `(1d)`/`(2d)`-adjacent `Operations` nodes were just replaced this round, see below) and (b) it was discovered via live testing performed specifically against this PR's branch, not a report against `main`.
- Distinct from the previously-tracked `(1d)` metadata-corruption finding: that was about a crawl's filename/mimetype columns getting collapsed across rows; this is a hard failure — zero documents ingested — for any URL whose content isn't HTML.

#### Code References

- `flows/openrag_url_mcp.json` (node `URLComponent-lnA0q`, `template.code.value`; see `_process_response` — no `content_type` branch — and `fetch_url_contents`'s `"No documents were successfully loaded from any URL"` raise)

#### Potential Solution

- Detect non-HTML `content-type` (e.g. `application/pdf`) in `_process_response` and route the raw bytes through the same Docling remote component used by file-upload ingestion, rather than attempting HTML parsing on binary content.
- Short term: fail with a clear, user-facing error ("this URL points to a PDF; use file upload instead") instead of the current opaque "component appears outdated" message, until Docling routing is added.

#### Alternative Solutions

- Document the limitation (URL ingestion is HTML/text-only) in user-facing docs/UI copy and defer actual PDF-over-URL support to a follow-up, since this is pre-existing behavior rather than something this PR broke.

mpawlow (approved):
## Code Review 3

- ✅ Approved / LGTM  🚀 
- FYI: PR Comment: (3a) - Pre-existing gap as per **Functional Review 1**

PR #2139, opened 2026-07-21
fix: fixed tui service status to now properly show container status instead of missing

### LINKED ISSUE #2134: [Bug] TUI is not reflecting state of the containers
When starting TUI for latest nightly, when starting the services, it starts the services but looks for the "opensearch", "openrag-backend" etc containers for the status, so it shows as Missing Container.
We need to update the TUI logic to look for the new container names, respecting the new logic introduced in https://github.com/langflow-ai/openrag/pull/1986

Vchen7629:
## Summary
Fixes #2134 The TUI status screen was showing all container services as "Missing" on Podman because the container name resolution relied on reconstructing a {project_name}-{service} prefix, but Compose derives the project name from the compose file's directory (~/.openrag/tui/), producing tui-* containers instead of the assumed openrag-*. The fix introduces a _resolve_service_name helper that keys off the stable com.docker.compose.service label (Podman) and the Service field (Docker) before falling back to the existing name map, matching the pattern already used in welcome.py. This makes status resolution robust regardless of what project name Compose derives at runtime.

## Changes
- `src/tui/managers/container_manager.py`: add _resolve_service_name helper and wire it into both Docker and Podman status paths to resolve containers by compose label before falling back to the name map
- `tests/unit/test_container_manager_status.py`: add regression and fallback tests for the new resolution logic

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container status detection for Docker and Podman environments.
  * Service names are now resolved reliably from compose metadata, even when container naming conventions differ.
  * Preserved fallback support for legacy container-name mappings.


lucaseduoli (approved):
LGTM but please merge to main instead of release-0.6.0. I will cherry pick the commit later

PR #2149, opened 2026-07-23
feat: add upgrade flows functionality

### LINKED ISSUE #2157: Langflow Flows Upgrade Path
## Motivation
Every time there is an update to the Langflow flows in the OpenRAG repo, the user has to factory reset their environment so that the flows are updated. This can cause issues if the flow update was a breaking change, which is happening often with each optimization that we do to the OpenRAG ingestion and retrieval.
This needs to be fixed for the 0.6 OSS upgrade, since there was a breaking change on the flows from one version to another.

## Solution
To stay backwards-compatible with the previous version of Langflow and OpenRAG, the upgrade is detected if the flow file's `Modified Date` is after the last `Modified Date` of the flow in OpenRAG. This solution is not perfect, but it works with our backend and Langflow.

- If the flow file is updated and the Langflow flow is outdated, a check will run when the backend is started. If the flow is detected as not custom (locked property is true), the flow will be updated automatically.
- If the flow is detected as custom, a popup will appear at the OpenRAG frontend to the person that has access to the flows, and the person can update and back up their flow to restore later:

Image

- If the user dismiss this, a banner will appear in the Settings page for when he wants to update it:

Image

## Caviats
- If the user does a modification to the flow, but locks the flow afterwards, the flow won't be detected as Custom. We don't have anything to track if the flow is custom besides that, since the backend still updates the flows to update the prompt/model, even if the flow is not custom.
    - Because of that, all flows that are updated are backed up locally, even if they are locked. If the user experiences the flows loss, they can still access their flows in the local backup.

- If the user ignores the flow update and then changes a model in the settings, the banner will not be shown again in the next Backend startup, since the new `Modified Date` will be after the flow's `Modified Date`.
    - This can be fixed if we make upgrading the custom flows mandatory, instead of optional.

lucaseduoli:
This pull request introduces a new system for notifying users about available Langflow custom flow updates and provides UI components and API hooks to manage updating or dismissing these notifications. The changes include new React Query hooks for fetching and mutating flow update state, two user-facing notification components (a global dialog and a settings page banner), and supporting UI and utility improvements.

The most important changes are:

**Langflow Flow Update Notification UI:**

* Added `FlowsUpdateDialog` component to globally prompt users about available Langflow flow updates, allowing them to update or dismiss updates, with backup options and error handling (`frontend/components/flows-update-dialog.tsx`).
* Added `LangflowUpdatesBanner` to the Langflow settings page to notify users about available updates and provide an update action with backup and error handling (`frontend/app/settings/_components/langflow-updates-banner.tsx`, `frontend/app/settings/[tab]/page.tsx`). ([frontend/app/settings/_components/langflow-updates-banner.tsxR1-R106](diffhunk://#diff-173acf28bc8ffa63fc1ede20c17875cd7a3a15fa5ad122b02261d42e99cdabefR1-R106), [frontend/app/settings/[tab]/page.tsxR15](diffhunk://#diff-09455844472841a91c96e5553088900c5b7c188213da8da67b5037f1eda110b7R15), [frontend/app/settings/[tab]/page.tsxR147](diffhunk://#diff-09455844472841a91c96e5553088900c5b7c188213da8da67b5037f1eda110b7R147))

**API Integration for Flow Updates:**

* Introduced React Query hooks for fetching available flow updates (`useGetFlowsUpdatesQuery`), updating flows (`useUpdateFlowsMutation`), and dismissing updates (`useDismissFlowsUpdateMutation`), with appropriate cache invalidation and error handling (`frontend/app/api/queries/useGetFlowsUpdatesQuery.ts`, `frontend/app/api/mutations/useUpdateFlowsMutation.ts`, `frontend/app/api/mutations/useDismissFlowsUpdateMutation.ts`). [[1]](diffhunk://#diff-09128188639dc9aadc9e5929f246e01681241a842c2d7d987f3f4db684b0832cR1-R23) [[2]](diffhunk://#diff-be00c7a7ea1aca7b09513548e50147b80b439618e6cd5efb9f1eff590f17ec25R1-R39) [[3]](diffhunk://#diff-43773ce325e88cee0a327f69f3c61321b481619d8c7a163baddebb9a24de05dbR1-R31)

**Supporting UI Components and Utilities:**

* Added a new `Checkbox` UI component and improved the `AlertDescription` styling for better notification presentation (`frontend/components/ui/checkbox.tsx`, `frontend/components/ui/alert.tsx`). [[1]](diffhunk://#diff-cd2c96d1493bbb25c0d566d6b386dd7f84a99381802494c63e1fdfb269155e72R1-R29) [[2]](diffhunk://#diff-5ef5589d230ecb61792ed20fda463e7b3a3189134a67a9fa9e4ffa0ba1a4a376L58-R58)
* Added a utility function `formatFlowName` to display flow names in a user-friendly format (`frontend/lib/utils.ts`).

**Dependency Updates:**

* Added `@radix-ui/react-checkbox` to dependencies to support the new checkbox component (`frontend/package-lock.json`).


frontend/app/settings/_components/langflow-updates-banner.tsx
```diff
@@ -0,0 +1,73 @@
...
+      console.error("Failed to update flows", e);
+    } finally {
+      setIsUpdating(false);
+    }
+  };
+
+  return (
+    <div className="mb-6">
+      <Alert
+        variant="warning"
```
mpawlow: ### (1a) [Critical] Invalid `Alert` variant breaks the frontend TypeScript build

#### Problem

- `LangflowUpdatesBanner` passes `variant="warning"` to `Alert`, but the `Alert` component (`frontend/components/ui/alert.tsx`) only defines `"default"` and `"destructive"` variants.
- This is a hard TypeScript compile error, not a lint nitpick.

#### Background Information

- Originally flagged by CodeRabbit as Critical, and corroborated independently by the `TypeScript type check` / `0_TypeScript type check.txt` CI job logs (`TS2322: Type '"warning"' is not assignable to type '"default" | "destructive" | null | undefined'`).

#### Potential Solution

- Change `variant="warning"` to `variant="default"` and keep the existing amber `className` overrides (border/background colors are already applied via `className`, so visual styling is preserved).

#### Alternative Solutions

- Add a proper `"warning"` variant to `alertVariants` in `frontend/components/ui/alert.tsx` if a warning-styled alert is expected to be reused elsewhere in the app.

src/services/flows_service.py
```diff
@@ -1445,3 +1445,142 @@ def _get_provider_component_ids(self, provider: str):
...
+                                )
+                            else:
+                                logger.warning(
+                                    f"Failed to create backup flow in Langflow for {flow_type}: HTTP {lf_create_res.status_code}"
+                                )
+                except Exception as e:
+                    logger.error(f"Failed to check/backup flow {flow_id} before update: {e}")
+
+            try:
+                reset_res = await self.reset_langflow_flow(flow_type)
```
mpawlow: ### (1b) [Critical] Flow reset proceeds even when the custom-flow backup fails, silently discarding user changes

#### Problem

- In `bulk_update_flows`, when `backup_custom=True`, the code attempts to (a) write a local backup file via `_backup_flow` and (b) create a backup flow directly in Langflow via `POST /api/v1/flows/`.
- Both of these are best-effort: `_backup_flow`'s return value isn't checked, the Langflow POST failure is only logged (`logger.warning`), and the whole backup `try` block swallows exceptions via a bare `except Exception` that also only logs.
- Regardless of whether either backup step succeeded, `self.reset_langflow_flow(flow_type)` is called unconditionally afterward — overwriting the user's custom flow.

#### Background Information

- This matches a Critical finding already raised by CodeRabbit with a concrete proposed fix (guard on `backup_path is None and backup_flow_id is None`, append a failed result, and `continue` instead of calling `reset_langflow_flow`).

#### Potential Solution

- After the backup attempt, if `backup_custom` is `True` and both `backup_path` and `backup_flow_id` are `None`, append a failed result (`"error": "Backup failed; update aborted to avoid losing custom changes"`) and `continue` to the next flow without calling `reset_langflow_flow`.
- Apply the same abort-and-continue behavior in the outer `except Exception` block that wraps the backup attempt.

#### Alternative Solutions

- Make backup failure a hard error surfaced to the caller (400/409) instead of silently skipping that flow in a bulk response, so the UI can retry explicitly.

src/services/flows_service.py
```diff
@@ -1445,3 +1445,142 @@ def _get_provider_component_ids(self, provider: str):
...
+
+                        try:
+                            dt = datetime.datetime.fromisoformat(langflow_updated_at_str)
+                            langflow_mtime = dt.timestamp()
+                        except ValueError:
+                            pass
+
+                    # 3. Compare timestamps
+                    # If the date modified on the local files is greater than what's in Langflow, it needs an update
+                    if local_mtime > langflow_mtime:
```
mpawlow: ### (1c) [Major] Update detection uses file `mtime` instead of content comparison, causing false positives

#### Problem

- `get_flows_updates_available` compares the local flow file's `os.path.getmtime()` against Langflow's `updated_at` timestamp to decide whether an "update available" notification should fire.
- Deployment/build pipelines (Docker image builds, git checkouts, container copies) routinely change file modification times without changing file content, which will cause false-positive "update available" banners/dialogs after every deploy — even when nothing actually changed.
- The codebase already has `_compare_flow_with_file(flow_id)` (`src/services/flows_service.py:818`), which does a normalized, content-based structural comparison (sorts nodes/edges, does a full equality check) and is used elsewhere for exactly this kind of "has this flow drifted from its file" check.

#### Background Information

- CodeRabbit raised this as Major and pointed at the same existing `_compare_flow_with_file` helper as the correct mechanism to reuse
- Discussed in a prior meeting: lightweight MD5 content comparison mechanism
- The "Alternative Solution" appears to be a great hybrid, optimized solution

#### Code References

- Existing correct mechanism: `src/services/flows_service.py:818-881` (`_compare_flow_with_file`)

#### Potential Solution

- Replace the `local_mtime > langflow_mtime` check with `not await self._compare_flow_with_file(flow_id)` (i.e., flag an update only when content actually differs), preserving the existing `is_custom` derivation from `locked`.

#### Alternative Solutions

- If timestamp comparison is kept for performance reasons, gate it as a cheap pre-filter and fall back to `_compare_flow_with_file` only when timestamps differ, to avoid a full content diff on every flow on every check.

src/api/flows.py
```diff
@@ -52,6 +45,46 @@ async def reset_flow_endpoint(
...
+
+async def bulk_update_flows_endpoint(
+    request: BulkUpdateFlowsRequest,
+    flows_service=Depends(get_flows_service),
+    user: User = Depends(require_permission("flows:edit")),
+):
+    """Bulk update multiple flows and optionally backup custom flows"""
+    try:
+        results = await flows_service.bulk_update_flows(request.flow_types, request.backup_custom)
+        return JSONResponse({"success": True, "results": results}, status_code=200)
```
mpawlow: ### (1d) [Major] `bulk_update_flows_endpoint` hardcodes `"success": true` regardless of per-flow results

#### Problem

- `flows_service.bulk_update_flows` returns a list of per-flow results, each with its own `success` boolean.
- `bulk_update_flows_endpoint` wraps this in `{"success": True, "results": results}` unconditionally — even if every single flow update failed, the top-level `success` is still `True`.
- Any caller (including the new frontend hooks) that only checks the top-level flag will report success when nothing actually succeeded.
- The codebase already has the correct pattern for this in `flows_service.py`'s `change_langflow_model_value`, which computes `all(r.get("success", False) for r in results)`.

#### Potential Solution

```python
results = await flows_service.bulk_update_flows(request.flow_types, request.backup_custom)
overall_success = all(r.get("success", False) for r in results)
return JSONResponse({"success": overall_success, "results": results}, status_code=200)
```

frontend/components/layout-wrapper.tsx
```diff
@@ -184,6 +185,8 @@ export function LayoutWrapper({ children }: { children: React.ReactNode }) {
           </div>
         </div>
       </div>
+
+      {(isAuthenticated || isNoAuthMode) && <FlowsUpdateDialog />}
```
mpawlow: ### (1e) [Major] Global update dialog is shown to users who lack permission to actually perform the update

#### Problem

- `FlowsUpdateDialog` is mounted in `LayoutWrapper` for **every** authenticated (or no-auth-mode) user on every page: `{(isAuthenticated || isNoAuthMode) && <FlowsUpdateDialog />}`. There is no RBAC gating on the frontend mount.
- `GET /settings/flows/updates-available` requires `flows:read`, which `admin`, `developer`, `user`, and `viewer` roles all have (`src/db/seed.py:129,154,168`) — so the dialog will pop up for all of them.
- `POST /settings/flows/update` requires `flows:edit`, which only `admin` and `developer` have — `user` and `viewer` roles do **not**.
- When `OPENRAG_RBAC_ENFORCE=true`, a `user`/`viewer`-role account will see the global "Flow Updates Available" dialog on login, and clicking "Update Flows" will 403.

#### Background Information

- RBAC is opt-in (default off), so this only manifests in deployments that have turned on `OPENRAG_RBAC_ENFORCE`. 

#### Code References

- `src/api/flows.py:80` (`flows:edit` requirement)
- `src/db/seed.py:129,154,168` (role → permission mapping)

#### Potential Solution

- Gate the `FlowsUpdateDialog` mount (and the settings banner, for consistency) on the user having `flows:edit`, using whatever permission-check hook/context the frontend already uses elsewhere (e.g., the `permissions` set built in `frontend/app/settings/[tab]/page.tsx:43-44`).

#### Alternative Solutions

- Keep the dialog visible to all roles as a passive notice, but disable the "Update Flows" button (with a tooltip) for users lacking `flows:edit`, rather than letting them hit a 403.

src/api/flows.py
```diff
@@ -1,12 +1,13 @@
 """Reset Flow API endpoints"""
+
 from typing import Literal
 
 from fastapi import Depends
 from fastapi.responses import JSONResponse
-from utils.logging_config import get_logger
 
-from dependencies import get_flows_service, get_current_user, require_permission
+from dependencies import get_current_user, get_flows_service, require_permission
```
mpawlow: ### (1f) [Normal] Unused `get_current_user` import — ruff F401, currently failing CI

#### Problem

- Concur with CodeRabbit.ai
- `get_current_user` is imported in `src/api/flows.py` but never referenced anywhere in the file.
- Confirmed as an active CI failure: `gh pr checks 2149` shows `Ruff and mypy on changed files → fail`, and CodeRabbit's captured log shows `F401: dependencies.get_current_user imported but unused`.

#### Potential Solution

```python
from dependencies import get_flows_service, require_permission
```

src/api/flows.py
```diff
@@ -52,6 +45,46 @@ async def reset_flow_endpoint(
     except Exception as e:
         logger.error("Unexpected error in flow reset", error=str(e))
         return JSONResponse(
-            {"success": False, "error": f"Internal server error: {str(e)}"},
-            status_code=500
+            {"success": False, "error": f"Internal server error: {str(e)}"}, status_code=500
+        )
+
+
+from typing import List
```
mpawlow: ### (1g) [Normal] Mid-file imports violate import ordering (E402), unused/deprecated `typing.List` (F401/UP035)

#### Problem

- `from typing import List` and `from pydantic import BaseModel` are declared mid-file (after `reset_flow_endpoint`), triggering `E402` (module-level import not at top of file).
- `List` is imported but unused since the code uses builtin `list[str]` — `F401`, plus `UP035` (deprecated in favor of builtin `list`).
- This is part of the same failing `Ruff and mypy on changed files` CI check as (1f).

#### Potential Solution

- Move `from pydantic import BaseModel` to the top-level import block (alongside `fastapi`/`dependencies` imports) and delete the mid-file `from typing import List` / `from pydantic import BaseModel` block entirely, keeping the existing `list[str]` annotation.

src/api/flows.py
```diff
@@ -52,6 +45,46 @@ async def reset_flow_endpoint(
...
+    user: User = Depends(require_permission("flows:edit")),
+):
+    """Bulk update multiple flows and optionally backup custom flows"""
+    try:
+        results = await flows_service.bulk_update_flows(request.flow_types, request.backup_custom)
+        return JSONResponse({"success": True, "results": results}, status_code=200)
+    except Exception as e:
+        logger.error("Error bulk updating flows", error=str(e))
+        return JSONResponse(
+            {"success": False, "error": f"Internal server error: {str(e)}"}, status_code=500
```
mpawlow: ### (1h) [Normal] Raw exception text returned to API clients (information exposure)

#### Problem

- All three exception handlers in `src/api/flows.py` (`reset_flow_endpoint`, the new `get_flows_updates_endpoint`, and the new `bulk_update_flows_endpoint`) return `f"Internal server error: {str(e)}"` directly in the JSON response body.
- GitHub Advanced Security / CodeQL flagged this three times (`Information exposure through an exception`) — once on the pre-existing endpoint and twice on the two new endpoints this PR adds.
- Stack-trace/exception text can leak internal paths, library versions, or other implementation details to any authenticated caller.

#### Background Information

- The pattern predates this PR (`reset_flow_endpoint` already did this), so it's not a regression introduced here — but the PR doubles the exposure surface by copying the same anti-pattern into two new endpoints instead of fixing it.

#### Code References

- `src/api/flows.py:48` (pre-existing), `src/api/flows.py:73`, `src/api/flows.py:89` (new)

#### Potential Solution

- Log the full exception server-side (already done via `logger.error`) and return a generic, non-identifying error message to the client (e.g., `"Internal server error"`), consistent with typical FastAPI error-handling conventions.

frontend/components/flows-update-dialog.tsx
```diff
@@ -0,0 +1,140 @@
...
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
+
+export function FlowsUpdateDialog() {
+  const { data: updates, isLoading } = useGetFlowsUpdatesQuery();
+  const updateMutation = useUpdateFlowsMutation();
+  const [isOpen, setIsOpen] = useState(false);
+  const [hasDismissed, setHasDismissed] = useState(false);
```
mpawlow: ### (1i) [Minor] React anti-patterns in `FlowsUpdateDialog` (extra renders / effect chaining)

#### Problem

- `hasDismissed` is `useState` but only ever set inside handlers and never rendered — it should be a `useRef` to avoid an unnecessary re-render.
- The initial-open logic (`useEffect` reading `sessionStorage` and calling `setIsOpen(true)`) is really event-driven ("when updates finish loading, open the dialog") but is implemented as a `useEffect` watching derived state, adding a render cycle and running one tick later than necessary.
- These are automated `react-doctor` warnings (score 88/100, 3 warnings, all in this new file) — style/perf nitpicks rather than correctness bugs, but worth cleaning up since they're isolated to code newly added in this PR.

#### Code References

- `frontend/components/flows-update-dialog.tsx:23` (`hasDismissed` state)
- `frontend/components/flows-update-dialog.tsx:35-36` (effect-driven open logic)

#### Potential Solution

- Replace `const [hasDismissed, setHasDismissed] = useState(false)` with a `useRef(false)`.
- Move the "open dialog when updates are available and not dismissed" logic to run directly where `updates` finishes loading (e.g., derive `isOpen` from query state + ref instead of a separate effect + state variable).

mpawlow (commented):
## Code Review 1

- See PR comments (a) to (h) for evaluation / consideration
   - ✅ Added a commit to fix (1f) and (1g) since it was non-controversial / trivial
- See Minor PR comment (i) for optional consideration

## Functional Review 1

- Deferred until next review (not performed yet)

src/services/flows_service.py
```diff
@@ -1445,3 +1478,177 @@ def _get_provider_component_ids(self, provider: str):
...
+                    {"flow_type": flow_type, "success": False, "error": "Invalid flow_type"}
+                )
+                continue
+
+            backup_path = None
+            backup_flow_id = None
+            is_custom = False
+
+            try:
+                response = await clients.langflow_request("GET", f"/api/v1/flows/{flow_id}")
```
mpawlow: ### (2a) [Critical] `bulk_update_flows` resets the flow when the pre-flight GET fails or raises, bypassing the backup-abort guard entirely

#### Problem

- The abort guard added to fix Review #1's (1b) only fires when `is_custom` is `True`:
  ```python
  if backup_custom and is_custom and backup_path is None and backup_flow_id is None:
      ...
      continue
  ```
- `is_custom` is initialized to `False` and is only set inside the `if response.status_code == 200:` branch. If the initial `GET /api/v1/flows/{flow_id}` returns a non-200 status, or the surrounding `try` block raises for any reason (network error, malformed JSON, etc.), `is_custom` stays `False` — so the guard's `is_custom` condition is `False`, the guard doesn't fire, and execution falls through to `reset_langflow_flow(flow_type)` unconditionally.
- In this failure mode the code never learns whether the flow was actually custom, and **no backup was attempted at all** (not even the local file backup) — this reproduces the exact silent-data-loss scenario that (1b)'s fix was meant to close, just through a different trigger (verification failure vs. backup failure).

#### Background Information

- This was detected by CodeRabbit.AI: https://github.com/langflow-ai/openrag/pull/2149/changes#r3647563373

#### Code References

- `src/services/flows_service.py:1571-1633`

#### Potential Solution

- Track verification success explicitly (e.g., a `flow_check_ok` flag set only inside the `status_code == 200` branch), and abort whenever `not flow_check_ok` **or** the existing backup-failure condition is true:
  ```python
  if not flow_check_ok or (
      backup_custom and is_custom and backup_path is None and backup_flow_id is None
  ):
      ...
      continue
  ```
- CodeRabbit's inline suggestion at comment id `3647563373` implements this directly and is close to a drop-in patch.

#### Alternative Solutions

- Treat a failed pre-flight GET as a hard per-flow error returned immediately (skip the "attempt reset anyway" path entirely for any flow whose current state couldn't be read), rather than folding it into the same abort/continue branch as the backup-failure case.

src/services/flows_service.py
```diff
@@ -1445,3 +1478,177 @@ def _get_provider_component_ids(self, provider: str):
...
+                                "is_custom": not is_locked,
+                                "dismissed": is_dismissed,
+                            }
+                        )
+            except Exception as e:
+                logger.error(f"Failed to fetch flow {flow_id} for update check: {e}")
+
+        return updates
+
+    async def bulk_update_flows(self, flow_types: list[str], backup_custom: bool = True):
```
mpawlow, line 1494: ### (2b) [Major] `bulk_update_flows` / `reset_langflow_flow` do not use the existing per-flow lock, allowing a concurrent provider/model update to race with a flow reset

#### Problem

- `FlowsService` already has a per-flow `asyncio.Lock` mechanism (`_get_flow_lock` / `_flow_locks`, `src/services/flows_service.py:108-122`), used by `_update_provider_components` (`flows_service.py:1099-1100`) to serialize the GET→modify→PATCH cycle when switching a flow's provider/model.
- `bulk_update_flows`'s own GET→backup→reset sequence (`flows_service.py:1560-1652`) never acquires this lock. Two concurrent triggers of this feature (e.g., a user bulk-updating flows while another request calls `_update_provider_components` on the same flow, or two concurrent bulk-update calls) can interleave: a provider/model change could land on the flow after this code reads/backs it up but before `reset_langflow_flow` overwrites it, silently discarding that concurrent edit, or two racing bulk updates could double-backup/double-reset.
- This is architecturally the same protection the codebase already applies elsewhere in this file for flow mutation — it just wasn't extended to the new bulk-update path.

#### Background Information

- Also, flagged by CodeRabbit.AI

#### Code References

- `src/services/flows_service.py:1547-1654` 
- `src/services/flows_service.py:109` 

#### Potential Solution

- Acquire `await self._get_flow_lock(flow_id)` around the GET→backup→reset sequence per flow, refactoring `reset_langflow_flow` (or introducing an internal `_reset_langflow_flow_locked` used by both callers) so it does not attempt to re-acquire the same lock.

#### Alternative Solutions

- If a full refactor is out of scope for this PR, at minimum document the race as a known limitation and file a fast-follow — the risk window is narrow (provider/model switching + flow bulk-update are both infrequent admin actions) but not zero.

src/services/flows_service.py
```diff
@@ -25,6 +25,19 @@
 
 
 class FlowsService:
+    def dismiss_flows_updates(self, flow_types: list[str] | None = None):
```
mpawlow: ### (2c) [Major] Dismissed-update state is a single process-wide set, not per-user — one user's "Skip for Now" hides the notification for every user

#### Problem

- The new dismissal feature stores dismissed flow types on `self._dismissed_updates`, a plain `set[str]` attribute on the `FlowsService` instance (`flows_service.py:28-37`, `__init__` at `:104-105`).
- `FlowsService` is constructed exactly once at app startup (`src/app/container.py:115`) and injected via `get_flows_service` as a shared singleton for the whole process — it is not scoped per-request, per-session, or per-user.
- Consequently, when any single user clicks "Skip for Now" (`handleDismiss` → `POST /settings/flows/dismiss-update`), the update notification disappears for **every** user hitting this backend process, not just the user who dismissed it. It only reappears when `bulk_update_flows` is later called for that flow type (`clear_dismissed_update`, `flows_service.py:1561`) or the process restarts.
- This undermines the purpose of surfacing the notification in the first place: one admin dismissing "for now" can silently suppress the update prompt for every other admin/developer with `flows:edit`, with no indication to them that it was dismissed (by whom, or that it's still pending).

#### Background Information

- This is a new gap introduced by this PR's own follow-up commit (`7be82384` "make dismissed modal ephemeral"), replacing the previous per-browser `sessionStorage` approach (which was at least scoped to one user's browser tab) with a mechanism that is *less* scoped (global across all users) while being marketed as more ephemeral. 

#### Code References

- `src/services/flows_service.py:28-39` 
- `src/services/flows_service.py:104-105` 
- `src/app/container.py:115` 

#### Potential Solution

- Scope dismissal to the requesting user: key `_dismissed_updates` by `user.id` (`dict[str, set[str]]`), using the `User` already injected into `dismiss_flows_update_endpoint` via `require_permission`, and thread the same `user` into `get_flows_updates_available` so it can check the per-user set.

#### Alternative Solutions

- Persist dismissal in the existing DB-backed `WorkspaceConfig`/per-user settings layer instead of in-memory, if the intent is for a dismissal to survive a backend restart — trades ephemerality for durability and per-user correctness in one change.
- If a single global "dismiss for everyone" is actually the intended UX (e.g., only `flows:edit` holders can dismiss, and it's meant as a workspace-wide snooze), state that explicitly in the PR description/UI copy so it's a documented decision rather than an apparent bug.

frontend/components/flows-update-dialog.tsx
```diff
@@ -0,0 +1,170 @@
...
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
+import { usePermissions } from "@/hooks/use-permissions";
+import { formatFlowName } from "@/lib/utils";
+
+export function FlowsUpdateDialog() {
+  const { can } = usePermissions();
+  const { data: updates, isLoading } = useGetFlowsUpdatesQuery();
```
mpawlow: ### (2d) [Normal] Update-availability query fires for every user regardless of `flows:edit` permission

#### Problem

- Both `FlowsUpdateDialog` and `LangflowUpdatesBanner` call `useGetFlowsUpdatesQuery()` unconditionally before their `can("flows:edit")` early-return, so the `GET /settings/flows/updates-available` request fires for every authenticated user on every mount (dialog: every page, via `LayoutWrapper`), including `user`/`viewer` roles who can never act on the result.
- The endpoint itself is gated by the weaker `flows:read` permission (correct per its own scope), so this isn't a broken-access-control bug — but it's wasted backend work (Langflow API calls per flow, per user, per page load) for a UI element that immediately discards the result for most roles.

#### Background Information

- Raised by CodeRabbit (comment id `3647563359`) against `langflow-updates-banner.tsx`; the same pattern exists in `flows-update-dialog.tsx` (both call the query hook before the permission check). Practically low-impact under the default permissive RBAC config but relevant once `OPENRAG_RBAC_ENFORCE=true` is set, which is exactly the deployment mode where `flows:edit` is restricted to admin/developer.

#### Code References

- `frontend/components/flows-update-dialog.tsx:25` (unconditional `useGetFlowsUpdatesQuery()` call) and `:86` (the `can("flows:edit")` early return it precedes); 
- `frontend/app/settings/_components/langflow-updates-banner.tsx:15` (query call) and `:20` (`can("flows:edit")` check); 

#### Potential Solution

- Pass `enabled: can("flows:edit")` to `useGetFlowsUpdatesQuery`'s underlying `useQuery` call so the fetch itself is skipped for users who can never act on the result.

frontend/app/settings/_components/langflow-updates-banner.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const [isUpdating, setIsUpdating] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+
+  if (!can("flows:edit") || isLoading || !updates || updates.length === 0) {
+    return null;
+  }
+
+  const hasCustomFlows = updates.some((u) => u.is_custom);
+
+  const handleUpdate = async () => {
```
mpawlow: ### (2e) [Normal] Update button re-enables before the updates-available list has refetched, allowing a duplicate submit

#### Problem

- In `LangflowUpdatesBanner.handleUpdate`, `finally { setIsUpdating(false) }` runs as soon as `updateMutation.mutateAsync(...)` resolves. `useUpdateFlowsMutation`'s `onSuccess` only *starts* `queryClient.invalidateQueries({ queryKey: ["flows", ...] })` — it isn't awaited by the caller, so the button re-enables and the (still-stale) "updates available" list can remain on screen for a brief window before the refetch lands, allowing a user to click "Update Flows" a second time against data that hasn't refreshed.
- Same shape of issue exists in `FlowsUpdateDialog` (it doesn't await invalidation before allowing the dialog to be reopened/re-triggered either), though the dialog's `updateMutation.isPending` disables its button until the mutation itself settles, narrowing (but not closing) the window.

#### Background Information

- Raised by CodeRabbit (comment id `3647563369`) specifically against the banner. Impact is bounded — worst case is a redundant `bulk_update_flows` call that re-backs-up and re-resets already-updated flows, not data loss, since (2a)'s fix (once applied) would still protect each individual call.

#### Code References

- `frontend/app/settings/_components/langflow-updates-banner.tsx:26-57` (`handleUpdate`, including the `finally { setIsUpdating(false) }` block at `54-56`); 
- `frontend/app/api/mutations/useUpdateFlowsMutation.ts` 

#### Potential Solution

- Await the query invalidation/refetch before clearing `isUpdating` in the `finally` block (or move `setIsUpdating(false)` into a `.then()` chained after invalidation resolves).

mpawlow (commented):
## Code Review 2

- See PR comments (2a), (2b), (2c) for consideration / evaluation
- See non-blocking PR comments (2d), (2e) for consideration

src/services/flows_service.py
```diff
@@ -460,9 +482,38 @@ async def reset_langflow_flow(self, flow_type: str):
             raise ValueError(
                 "flow_type must be either 'nudges', 'retrieval', 'ingest', or 'url_ingest'"
             )
-
         if not flow_id:
             raise ValueError(f"Flow ID not configured for flow_type '{flow_type}'")
+        return flow_id
+
+    async def reset_langflow_flow(self, flow_type: str):
```
mpawlow, line 489: ### (4a) [Normal] `reset_langflow_flow` self-deadlocks by reacquiring its own per-flow lock via `reapply_all_settings`

#### Problem

- `reset_langflow_flow` (`flows_service.py:489-512`) acquires the per-flow `asyncio.Lock` for `flow_id` at line 499 and holds it for the entire method body, including a new call to `reapply_all_settings()` at line 509 (added to replace the per-flow config-update logic that used to live inline in `_reset_langflow_flow_locked`).
- `reapply_all_settings()` (`src/api/settings/langflow_sync.py:278`) calls `_update_langflow_model_values(config, flows_service)` (line 300), which — when called with no explicit provider args, exactly as it is here — iterates **every configured LLM and embedding provider across all 4 flows** via `flows_service.change_langflow_model_value(...)` (`langflow_sync.py:211`, `:227`).
- `change_langflow_model_value` (`flows_service.py:971-1035`) defaults `flow_configs` to all 4 flows (nudges/retrieval/ingest/url_ingest) and fans out via `asyncio.gather` to `self._update_provider_components(config, provider, ...)` for each one (`flows_service.py:1004-1018`).
- `_update_provider_components` (`flows_service.py:1037-1057`) itself does `flow_lock = await self._get_flow_lock(flow_id); async with flow_lock:`. Because `_get_flow_lock` (`flows_service.py:132-137`) returns the **same** `asyncio.Lock` instance for a given `flow_id` (LRUCache-backed), the flow that was just reset — still locked by the outer `reset_langflow_flow` call — is one of the 4 flows `change_langflow_model_value` will try to update.
- `asyncio.Lock` is non-reentrant: a second `acquire()` on an already-locked lock blocks until `release()` is called — but `release()` can only happen when the outer `async with` block in `reset_langflow_flow` exits, which is itself blocked awaiting this inner acquire. **The task deadlocks permanently.** Since `asyncio.gather` (used internally by `change_langflow_model_value`, and by `asyncio.gather` semantics never returns until all children complete) never resolves, `reapply_all_settings()` never returns, `reset_langflow_flow()` never returns, and — critically — **the lock for that `flow_id` is never released**, so every subsequent operation on that same flow (another reset, a provider-component update, a bulk update) will also hang forever, until the backend process is restarted.

#### Code References

- `src/services/flows_service.py:489-512` (`reset_langflow_flow` — lock held across `reapply_all_settings()`)
- `src/services/flows_service.py:132-137` (`_get_flow_lock` — confirms same-`flow_id` lock reuse)
- `src/services/flows_service.py:1037-1057`, `:971-1035` (`_update_provider_components` / `change_langflow_model_value` — re-acquires the lock)
- `src/api/settings/langflow_sync.py:278-323`, `:149-235` (`reapply_all_settings` → `_update_langflow_model_values`)
- `src/api/flows.py:18-45` (reachable endpoint)
- `src/api/settings/endpoints.py:1583-1596` (reachable rollback path)

#### Potential Solution

- Move the `reapply_all_settings()` call in `reset_langflow_flow` to **after** the `async with flow_lock:` block exits (mirroring the pattern already used correctly in `bulk_update_flows` and `ensure_flows_exist`):
  ```python
  async def reset_langflow_flow(self, flow_type: str):
      flow_id = self._get_flow_id_by_type(flow_type)
      flow_lock = await self._get_flow_lock(flow_id)
      async with flow_lock:
          result = await self._reset_langflow_flow_locked(flow_type, flow_id)
      if result.get("success"):
          self._custom_updates_cache = None
          try:
              config = get_openrag_config()
              if config.edited:
                  from api.settings import reapply_all_settings
                  await reapply_all_settings()
          except Exception as e:
              logger.error(f"Error reapplying settings after flow reset: {e}")
      return result
  ```

#### Alternative Solutions

- Add an `exclude_flow_id` / `exclude_flow_ids` parameter threaded through `reapply_all_settings` → `_update_langflow_model_values` → `change_langflow_model_value` → `_update_provider_components`, so the currently-held flow is skipped during reapply. More invasive, and not obviously more correct than simply reapplying after releasing the lock.
- Have `_update_provider_components` accept an already-held lock (or a lock-free `_update_provider_components_locked` call site keyed on "am I already inside this flow's lock") the same way `_reset_langflow_flow_locked` does. Adds complexity for no real benefit over just narrowing the lock's scope.

src/services/flows_service.py
```diff
@@ -1445,3 +1442,182 @@ def _get_provider_component_ids(self, provider: str):
...
+                            "backup_path": None,
+                            "backup_flow_id": None,
+                        }
+                    )
+                    continue
+
+                try:
+                    from unittest.mock import AsyncMock, MagicMock
+
+                    if isinstance(self.reset_langflow_flow, (AsyncMock, MagicMock)):
```
mpawlow: ### (4b) [Normal] Production code branches on `isinstance(self.reset_langflow_flow, (AsyncMock, MagicMock))`

#### Problem

- `bulk_update_flows` (`flows_service.py:1585-1591`) contains:
  ```python
  from unittest.mock import AsyncMock, MagicMock

  if isinstance(self.reset_langflow_flow, (AsyncMock, MagicMock)):
      reset_res = await self.reset_langflow_flow(flow_type)
  else:
      reset_res = await self._reset_langflow_flow_locked(flow_type, flow_id)
  ```
- This imports `unittest.mock` inside a production request-handling code path and uses it to detect whether the instance's `reset_langflow_flow` attribute has been monkeypatched by a test, in order to decide which code path to run. This is a testability anti-pattern: production behavior should never branch on "am I being unit tested."
- The practical consequence is exactly what caused (4a) to ship undetected: because `bulk_update_flows`' existing tests (`test_bulk_update_flows_creates_backup_flow_in_langflow`, `test_bulk_update_flows_aborts_reset_on_backup_failure`, `test_bulk_update_flows_saves_local_file_backup_for_default_flows`) all `patch.object(service, "reset_langflow_flow", ...)`, they take the `isinstance(...)` **true** branch and call the mock — meaning **none of them ever execute `_reset_langflow_flow_locked` through `bulk_update_flows`'s real code path**. The `else` branch (the one that actually runs in production) has zero direct test coverage.

#### Potential Solution

- Always call `self._reset_langflow_flow_locked(flow_type, flow_id)` directly (no `isinstance` branch), and update the three affected tests in `tests/unit/services/test_flows_service_bulk_update.py` to `patch.object(service, "_reset_langflow_flow_locked", ...)` instead of `reset_langflow_flow`.

src/services/flows_service.py
```diff
@@ -1445,3 +1442,182 @@ def _get_provider_component_ids(self, provider: str):
...
+        """
+        results: list[dict[str, Any]] = []
+        flow_configs_dict = {
+            "nudges": NUDGES_FLOW_ID,
+            "retrieval": LANGFLOW_CHAT_FLOW_ID,
+            "ingest": LANGFLOW_INGEST_FLOW_ID,
+            "url_ingest": LANGFLOW_URL_INGEST_FLOW_ID,
+        }
+
+        for flow_type in flow_types:
```
mpawlow, line 1509: ### (4c) [Normal] `bulk_update_flows` clears per-user dismissal state unconditionally, even when the update aborts

#### Problem

- `bulk_update_flows`'s per-flow-type loop begins with `self.clear_dismissed_update(flow_type)` (`flows_service.py:1497`), called **before** the pre-flight GET check, before the backup attempt, and before the reset — i.e., before it's known whether the update will actually succeed.
- If the update subsequently aborts (pre-flight GET fails → (2a)'s now-fixed guard at `:1554-1567`, or backup fails → `:1570-1583`), the dismissal state has already been wiped for **every user** who had previously dismissed that flow's update notification. The bulk-update result correctly reports `success: False`, but the side effect (clearing dismissals) is not rolled back.
- Net effect: a failed or aborted update attempt causes the "Flow Updates Available" banner/dialog to reappear for users who had explicitly dismissed it, even though nothing about the flow's actual state changed.

#### Code References

- `src/services/flows_service.py:61-68` (`clear_dismissed_update` definition)

#### Potential Solution

- Move `self.clear_dismissed_update(flow_type)` to only run after a successful reset, alongside the existing `self._custom_updates_cache = None` invalidation at `flows_service.py:1612-1613` (which already only runs `if results and any(r.get("success") for r in results)`), or scope it per-flow to only fire when that specific flow's `reset_res.get("success")` is `True`.

mpawlow (approved):
## Code Review 3

- ✅  Approved / LGTM  🚀 
- See non-blocking PR comments (4a), (4b), (4c) for optional consideration

## Functional Review 1

- ✅ Verified existing flows replaced and flow back-ups successfully occurred when manually modifying the JSON of the flows in the development workspace
- ⚠️ Second modification to the flows JSON did not trigger another replace and back-up operation
- ⚠️ Unable to verify whether dismissing the replace / back-up operation shows up in the **Langflow** Settings page

frontend/components/flows-update-dialog.tsx
```diff
@@ -0,0 +1,204 @@
...
+          <div className="py-2">
+            <p className="text-sm text-muted-foreground leading-relaxed">
+              Updating Langflow will overwrite any custom changes made. Backup
+              copies of your flows will be created located in Langflow, Do you
+              want to continue?
+            </p>
+          </div>
+
+          <DialogFooter>
+            <Button variant="outline" onClick={() => setShowConfirm(false)}>
```
mpawlow: ### (5a) [Minor] Clicking "Cancel" on the update-confirmation dialog silently suppresses the prompt for the rest of the browser session

#### Problem

- `FlowsUpdateDialog` is mounted once at the root layout (`frontend/app/layout.tsx` → `LayoutWrapper`), so it persists across all client-side navigation for the life of the browser tab — it is not remounted per page.
- Its auto-open logic only fires when `isLoading` or `hasUndismissed` **changes value** (`flows-update-dialog.tsx:53-66`, comparing against `prevIsLoading`/`prevHasUndismissed`).
- The two-step flow is: first dialog → "Update flow" → `handleInitialUpdateClick` closes the first dialog and opens a confirm dialog (`showConfirm`) → user clicks "Cancel" → `onClick={() => setShowConfirm(false)}` (`flows-update-dialog.tsx:190`) simply closes the confirm dialog. Nothing calls `dismissMutation`, and nothing re-opens the first dialog or flips `hasUndismissed`.
- Net effect: after Cancel, both dialogs are closed, the update is **not** recorded as dismissed on the backend, `hasUndismissed` is still `true` — but since its boolean value never changes again, the effect-like block at lines 56-66 never re-fires, so the dialog cannot reopen itself. The only way to see the prompt again is a full page reload (client-side navigation won't do it, since the component isn't remounted).
- A user who opens the dialog, gets to the confirmation step, and backs out with "Cancel" (a very natural "not right now" action) effectively loses all further visibility into a pending flow update until they hard-refresh — worse than the explicit "Skip action"/dismiss path, which at least records the dismissal server-side and is intentional.

#### Code References

- `frontend/components/flows-update-dialog.tsx:53-66` (auto-open effect-like logic, keyed on value changes only)
- `frontend/components/flows-update-dialog.tsx:81-84` (`handleInitialUpdateClick`)
- `frontend/components/flows-update-dialog.tsx:189-192` (Cancel button — no dismiss, no reopen)
- `frontend/app/layout.tsx` (confirms `LayoutWrapper`, and thus `FlowsUpdateDialog`, is mounted at the root and persists across navigation)

#### Potential Solution

- Treat "Cancel" as "go back to the first dialog" rather than "close everything": `onClick={() => { setShowConfirm(false); setIsOpen(true); }}`, so the user lands back where they can explicitly Skip or retry Update.

#### Alternative Solutions

- If "Cancel" is meant to be equivalent to a soft dismiss, call `handleDismiss()` from the Cancel button so it's at least recorded server-side and consistent with the "Skip action" path.

frontend/components/flows-update-dialog.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+      <Dialog open={showConfirm} onOpenChange={setShowConfirm}>
+        <DialogContent className="sm:max-w-[540px]">
+          <DialogHeader>
+            <DialogTitle>Confirm Langflow Update</DialogTitle>
+          </DialogHeader>
+
+          <div className="py-2">
+            <p className="text-sm text-muted-foreground leading-relaxed">
+              Updating Langflow will overwrite any custom changes made. Backup
```
mpawlow, line 183: ### (5c) [Minor] Malformed sentence in the update-confirmation dialog copy

#### Confidence

90

#### Problem

- The confirmation dialog reads: *"Backup copies of your flows will be created located in Langflow, Do you want to continue?"* — "will be created located in" is a grammatical error (duplicate verb phrase), and "Do you want to continue?" is incorrectly capitalized mid-sentence after a comma.

#### Code References

- `frontend/components/flows-update-dialog.tsx:182-186`

#### Potential Solution

```
Updating Langflow will overwrite any custom changes made. Backup copies of
your flows will be created in Langflow. Do you want to continue?
```

mpawlow (approved):
## Code Review 5

- ✅ Approved / LGTM  🚀 
- See Minor PR comments (5a) (5c) for optional consideration

frontend/app/settings/_components/langflow-updates-banner.tsx
```diff
@@ -4,15 +4,18 @@ import { AlertTriangle, X } from "lucide-react";
...
+import { DEV_IBM_COS } from "@/lib/brand";
 import { cn } from "@/lib/utils";
 
 export function LangflowUpdatesBanner() {
   const { brand } = useBrand();
   const isIbm = brand === "ibm";
+  const { isIbmAuthMode } = useAuth();
   const { can } = usePermissions();
-  const canEdit = can("flows:edit");
+  const canEdit = can("flows:edit") && !(isIbmAuthMode || DEV_IBM_COS);
```
edwinjosechittilappilly: @mfortman11  we are passing the run mode in api I believe can we use that here ? instead of the auth mode, or ir is the same ?

mfortman11: We can do 
```
const { runMode } = useAuth();
const { can } = usePermissions();
const canEdit = can("flows:edit") &&  runMode === 'oss';
```

edwinjosechittilappilly (approved):
code LGTM

mfortman11 (approved):

PR #2152, opened 2026-07-24
fix: provide real error for onboarding apiKey setup and settings

Wallgau:
issue: https://github.ibm.com/lakehouse/tracker/issues/77752#
Summary
Surface real provider credential errors (sanitized, no wrapper prefixes) when validating API keys in onboarding and settings, instead of generic copy like “Invalid OpenAI API key” / “Connection failed.”
Propagate cleaned messages from /api/models/* and settings provider validation using the same formatting helpers as chat/ingest (sanitize_provider_error_content / formatProviderErrorMessage).
Fix env-key toggle validation so switching back to “Use environment API key” does not keep sending a previously typed (e.g. disabled) custom key (useEnvKey omits api_key from the request).
Test plan

 Onboarding watsonx/OpenAI/Anthropic: enter an invalid/disabled key → UI shows the real provider message

 Toggle off env key, enter a bad key, then toggle env back on → revalidates env credentials (no stale disabled-key error)

 Settings provider dialogs: save with a bad key → field shows the real error

 Confirm successful env/config keys still load models and clear the error

 Unit: tests/unit/test_models_api_errors.py, tests/unit/test_provider_error_formatting.py

## Summary by CodeRabbit

* **New Features**
  * Added an env-key mode for fetching provider models, using environment credentials instead of user-entered keys.
* **Bug Fixes**
  * Improved onboarding/settings flows to gate validation until loading completes and to display the exact server-provided error message.
  * Standardized model-fetch error handling (including safer credential redaction) and improved request/caching behavior by tracking resolved provider parameters.
  * Updated settings persistence to clear stored keys when env-key mode is enabled.
* **Tests**
  * Added unit coverage for sanitized model endpoint errors and adjusted connection/config error assertions.

### DIFF
--- a/frontend/app/api/queries/useGetModelsQuery.ts
+++ b/frontend/app/api/queries/useGetModelsQuery.ts
@@ -65,11 +65,10 @@ export const useGetOpenAIModelsQuery = (
   return useQuery(
     {
       queryKey: ["models", "openai", useEnvKey, apiKey] as const,
-      queryFn: async ({ queryKey }) => {
-        const [, , envMode, key] = queryKey;
+      queryFn: async (): Promise<ModelsResponse> => {
         const body: { api_key?: string } = {};
-        if (!envMode && key) {
-          body.api_key = key;
+        if (!useEnvKey && apiKey) {
+          body.api_key = apiKey;
         }
 
         const response = await fetch("/api/models/openai", {
@@ -80,7 +79,7 @@ export const useGetOpenAIModelsQuery = (
         if (response.ok) {
           return (await response.json()) as ModelsResponse;
         }
-        await throwModelsFetchError(response, "Failed to fetch OpenAI models");
+        return throwModelsFetchError(response, "Failed to fetch OpenAI models");
       },
       staleTime: 0,
       gcTime: 0,
@@ -102,11 +101,10 @@ export const useGetAnthropicModelsQuery = (
   return useQuery(
     {
       queryKey: ["models", "anthropic", useEnvKey, apiKey] as const,
-      queryFn: async ({ queryKey }) => {
-        const [, , envMode, key] = queryKey;
+      queryFn: async (): Promise<ModelsResponse> => {
         const body: { api_key?: string } = {};
-        if (!envMode && key) {
-          body.api_key = key;
+        if (!useEnvKey && apiKey) {
+          body.api_key = apiKey;
         }
 
         const response = await fetch("/api/models/anthropic", {
@@ -117,7 +115,7 @@ export const useGetAnthropicModelsQuery = (
         if (response.ok) {
           return (await response.json()) as ModelsResponse;
         }
-        await throwModelsFetchError(
+        return throwModelsFetchError(
           response,
           "Failed to fetch Anthropic models",
         );
@@ -141,18 +139,17 @@ export const useGetOllamaModelsQuery = (
   return useQuery(
     {
       queryKey: ["models", "ollama", endpoint] as const,
-      queryFn: async ({ queryKey }) => {
-        const [, , ep] = queryKey;
+      queryFn: async (): Promise<ModelsResponse> => {
         const url = new URL("/api/models/ollama", window.location.origin);
-        if (ep) {
-          url.searchParams.set("endpoint", ep);
+        if (endpoint) {
+          url.searchParams.set("endpoint", endpoint);
         }
 
         const response = await fetch(url.toString());
         if (response.ok) {
           return (await response.json()) as ModelsResponse;
         }
-        await throwModelsFetchError(response, "Failed to fetch Ollama models");
+        return throwModelsFetchError(response, "Failed to fetch Ollama models");
       },
       staleTime: 0,
       gcTime: 0,
@@ -183,21 +180,20 @@ export const useGetIBMModelsQuery = (
         projectId,
         apiKey,
       ] as const,
-      queryFn: async ({ queryKey }) => {
-        const [, , envMode, ep, proj, key] = queryKey;
+      queryFn: async (): Promise<ModelsResponse> => {
         const body: {
           endpoint?: string;
           api_key?: string;
           project_id?: string;
         } = {};
-        if (ep) {
-          body.endpoint = ep;
+        if (endpoint) {
+          body.endpoint = endpoint;
         }
-        if (proj) {
-          body.project_id = proj;
+        if (projectId) {
+          body.project_id = projectId;
         }
-        if (!envMode && key) {
-          body.api_key = key;
+        if (!useEnvKey && apiKey) {
+          body.api_key = apiKey;
         }
 
         const response = await fetch("/api/models/ibm", {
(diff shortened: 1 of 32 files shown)


edwinjosechittilappilly:
@coderabbitai review this PR, will why sanitize_provider_error_content still be flagged by codeql ?

edwinjosechittilappilly:
@Wallgau  can you take a look in to the lint issue also ?

frontend/app/onboarding/_hooks/useUpdateSettings.ts
```diff
@@ -31,15 +31,14 @@ export function useUpdateSettings(
...
-          updatedSettings.watsonx_api_key = config.apiKey;
-        }
+      // Map provider-specific API keys (clear when empty so env-key mode
+      // does not keep a previously typed key in onboarding settings).
+      if (provider === "openai") {
+        updatedSettings.openai_api_key = config.apiKey || "";
+      } else if (provider === "anthropic") {
+        updatedSettings.anthropic_api_key = config.apiKey || "";
+      } else if (provider === "watsonx") {
+        updatedSettings.watsonx_api_key = config.apiKey || "";
```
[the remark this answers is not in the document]
edwinjosechittilappilly: @Wallgau

frontend/app/onboarding/_components/onboarding-card.tsx
```diff
@@ -408,18 +413,13 @@ const OnboardingCard = ({
...
 
-      // Set to final step to show "Done"
+      // Set to final step to show "Done", then delay onComplete via pendingComplete.
       setCurrentStep(totalSteps);
-      // Wait a bit before completing — stored in a ref so the timeout
-      // survives the re-render caused by setCurrentStep above.
-      completeTimeoutRef.current = setTimeout(() => {
-        onComplete();
-      }, 1000);
+      setPendingComplete(true);
```
Wallgau, line 418: this is intentional: show “Done”, then latch a delayed onComplete. Timer needs an effect; not a prop-driven chain

frontend/app/api/queries/useGetModelsQuery.ts
```diff
@@ -19,10 +20,14 @@ export interface ModelsResponse {
 
 export interface OpenAIModelsParams {
   apiKey?: string;
+  /** When true, omit api_key so the backend uses configured/env credentials. */
+  useEnvKey?: boolean;
```
lucaseduoli, line 24: is this needed? when apiKey is null isn't it implicit that it's using env?

lucaseduoli (changes requested):

frontend/app/api/queries/useGetModelsQuery.ts
Wallgau: on the API side you’re right, null/missing apiKey already means we fall back to env/config. We don’t need a backend flag for that.
useEnvKey is for the frontend. Empty key can mean two things: user hasn’t typed anything yet" (don’t call the models API), or "use the env key" (do call it, and don’t send a body key). Without that flag the toggle can’t tell those apart, so after you type a bad/disabled key and switch to env it can keep the old failed query / warning instead of actually validating the env key.
Screenshot 2026-07-24 at 8 52 28 AM

frontend/app/onboarding/_components/onboarding-card.tsx
```diff
@@ -376,7 +381,7 @@ const OnboardingCard = ({
         failed_files: taskWithFailure.failed_files,
       });
 
-      clearTimeout(completeTimeoutRef.current);
+      setPendingComplete(false);
```
Wallgau, line 384: Not a prop→state sync. This effect watches polled task results and cancels the Done→onComplete latch when ingest fails, then shows the error / rolls back. The render-time prev-prop pattern doesn’t fit this path, leaving as-is.

Wallgau, line 384: No change needed. Failure handling sets related state in one branch (React batches it); this isn’t an effect→effect chain. Trigger is polled task status, not a user event handler we can fold into.

frontend/app/onboarding/_components/onboarding-card.tsx
Wallgau, line 418: Intentional: show the "Done" step, then latch a delayed onComplete via pendingComplete (timer needs an effect). Not syncing local state from a parent prop, so the render-time prev-prop pattern doesn’t apply.

Wallgau, line 418: No change needed. Step + latch are set together (batched); the follow-on effect only runs the delayed onComplete timer. Not an effect→effect state cascade we can fold into a single event handler.

src/api/models.py
```diff
@@ -24,6 +46,51 @@ class IBMBody(BaseModel):
...
+            status_code=502,
+        )
+
+    error = sanitize_provider_error_content(exc)
+    redacted = _redact_credentials(error)
+    if is_provider_credential_error(exc) or is_provider_credential_error(error):
+        return JSONResponse(
+            {
+                "error": (
+                    "The configured API key is invalid or has been revoked. "
```
lucaseduoli: This message is the same that is used in the `error` variable, defined in `sanitize_provider_error_content`. In this case, as the error is redacted and sanitized, either pass all custom messages to this function, or to the `sanitize_provider_error_content` function

lucaseduoli (changes requested):
Error shown in console:

Image

Error shown in onboarding:

Image

And it was actually a wrong API key - I used another project's API keys and it gave me that error.

And when I use a disabled API key, this is what shows up:

Image

I think we can change the message since the API key was not configured yet

frontend/app/onboarding/_components/onboarding-card.tsx
Wallgau, line 384: No change needed. This isn’t syncing local state from a parent prop — the effect watches polled task results and cancels the Done→`onComplete` latch when ingest fails, then shows the error / rolls back. The render-time prev-prop pattern doesn’t apply here.

frontend/app/onboarding/_components/onboarding-card.tsx
Wallgau, line 418: No change needed. Intentional: show the "Done" step, then latch a delayed `onComplete` via `pendingComplete` (the timer needs an effect). Not syncing local state from a parent prop, so the render-time prev-prop pattern doesn’t apply.

frontend/app/onboarding/_components/onboarding-card.tsx
Wallgau, line 384: No change needed. Failure handling sets the related state together in one branch (React batches it). The trigger is polled task status, not an event-handler → effect → effect cascade we can fold away.

frontend/app/onboarding/_components/onboarding-card.tsx
Wallgau, line 418: No change needed. Step + latch are set together (batched); the follow-on effect only runs the delayed `onComplete` timer. Not a state-update chain we can collapse into a single event handler.

frontend/app/onboarding/_components/onboarding-card.tsx
Wallgau: The effect reacts to polled task status, not a parent prop. setPendingComplete(false/true) cancels or latches the Done → onComplete delay; the prev-prop render pattern doesn’t apply.

lucaseduoli (approved):
LGTM!!

PR #2165, opened 2026-07-28
feat: Knowledge table optimization

jason-dong03:
**Frontend**
_page.tsx_: 
- added currentPage / currentPageSize React States (as opposed to hardcoded values limiting to 100 files)
- sortBy / sortOrder state fed to the server with an onSortChanged handler that reads AG Grid column state, maps it to backend field names, resets to page 1, and clears the cursor cache on sort change 
- replaced all per-column client-side comparator functions on server-sorted columns (filename, size, mimetype, owner, chunkCount, embedding_model, embedding_dimensions) with comparator: () => 0 so AG Grid preserves the order the API returned rather than re-sorting the page locally 
- cursor cache (cursorCacheRef) stores after_key per page number for forward navigation; 
- custom pagination footer below the grid (page-size select, row summary, « / ‹ / › nav buttons)
_useListFiles_.ts: 
- added afterKey, sortBy, sortOrder to ListFilesParams
- after_key to ListFilesResponse; URL updated to /api/v2/files;
- full params object in the React Query key so any param change triggers a refetch

**Backend**
-Fixed the sort-key crash in file_service.py:241-245
_src/api/v2/files.py + src/api/v2/__init__.py_:
-new v2 endpoint package
-handlers accept after_key cursor, sort_by, sort_order
- routed at /v2/files and /v2/files/search
_src/services/file_service_v2.py_ : new service using OpenSearch composite aggregation; O(page_size) cost per request instead of returning all unique files on page switch; returns after_key for the next page (null on the last page); cardinality aggregation for the total count
_src/app/routes/internal.py_ : registers /v2/files/search and /v2/files; imports from api.v2
**Why v1 is kept**
/api/files (v1, terms-agg) remains unchanged for anything already calling it. The frontend exclusively uses v2. The two are fully independent.

## Summary by CodeRabbit

* **New Features**
  * Added faster server-side file listing and searching with cursor-based pagination.
  * Added sortable file listings with filters and configurable page sizes up to 500.
  * Added custom pagination controls for wildcard file views.

* **Bug Fixes**
  * Improved numeric file sorting.
  * Improved provider validation, connection status handling, and user-facing error messages.
  * Improved onboarding model-loading indicators and configuration feedback.

* **Tests**
  * Updated coverage for pagination, provider errors, and onboarding behavior.


frontend/app/knowledge/page.tsx
```diff
@@ -487,6 +499,32 @@ function SearchPage() {
...
+
+    const newSortBy = sortedCol
+      ? (AG_FIELD_TO_SORT_BY[sortedCol.colId] ?? sortedCol.colId)
+      : "filename";
+    const newSortOrder: "asc" | "desc" =
+      sortedCol?.sort === "desc" ? "desc" : "asc";
+
+    console.log(
+      `%c[sort]     sortBy=${newSortBy}  sortOrder=${newSortOrder}`,
+      "color: #e67e22; font-weight: bold",
```
mfortman11: Remove console.log

src/api/v2/files.py
```diff
@@ -0,0 +1,115 @@
...
+    from services.file_service_v2 import FileServiceV2
+
+    return FileServiceV2(session_manager=session_manager)
+
+
+async def list_files(
+    page: int = Query(1, ge=1, description="Page number (for display only; navigation uses after_key cursor)"),
+    page_size: int = Query(25, ge=1, le=500, description="Items per page"),
+    sort_by: str = Query("filename", description="Sort field"),
+    sort_order: str = Query("asc", regex="^(asc|desc)$", description="Sort order"),
```
mfortman11: regex is deprecated use `pattern` as done in the v1 endpoint

frontend/app/api/queries/useListFiles.ts
```diff
@@ -41,8 +43,12 @@ export const useListFiles = (
     if (params.mimetype) searchParams.set("mimetype", params.mimetype);
     if (params.owner) searchParams.set("owner", params.owner);
     if (params.search) searchParams.set("search", params.search);
+    if (params.afterKey)
+      searchParams.set("after_key", JSON.stringify(params.afterKey));
+
+    // const url = `/api/files?${searchParams.toString()}`;
```
mfortman11: remove commented code

src/services/file_service_v2.py
```diff
@@ -0,0 +1,288 @@
...
+                        "precision_threshold": 3000,
+                    }
+                }
+            },
+        }
+        try:
+            result = await opensearch_client.search(index=get_index_name(), body=body)
+            return result.get("aggregations", {}).get("file_count", {}).get("value", 0)
+        except Exception:
+            return 0
```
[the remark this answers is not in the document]
mfortman11: I agreed w/ this we'll need to handle and log these exceptions

frontend/app/knowledge/page.tsx
```diff
@@ -1036,73 +1065,183 @@ function SearchPage() {
...
                 </div>
-              </div>
-            )}
-          />
+              )}
+            />
+          </div>
+        )}
+
+        {isWildcardQuery && serverTotal > 0 && (
```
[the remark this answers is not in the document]
mfortman11: Agree here we should show pagination when a search is applied as well

mfortman11 (changes requested):

frontend/components/knowledge-pagination-footer.tsx
```diff
@@ -0,0 +1,149 @@
...
+        justifyContent: "flex-end",
+        gap: 16,
+        height: "var(--ag-pagination-panel-height, 48px)",
+        paddingInline: "calc(var(--ag-grid-size, 4px) * 3)",
+        borderTop: "1px solid var(--ag-border-color, hsl(var(--border)))",
+        backgroundColor: "var(--ag-background-color, hsl(var(--background)))",
+        color: "var(--ag-foreground-color, hsl(var(--muted-foreground)))",
+        fontSize: "var(--ag-font-size, 14px)",
+        fontFamily: "var(--ag-font-family, inherit)",
+      }}
```
mfortman11: Can you ask your agent to turn this into tailwind

frontend/components/knowledge-pagination-footer.tsx
```diff
@@ -0,0 +1,149 @@
...
+          style={{
+            background: "var(--ag-background-color, hsl(var(--background)))",
+            color: "var(--ag-foreground-color, hsl(var(--muted-foreground)))",
+            border: "1px solid var(--ag-border-color, hsl(var(--border)))",
+            borderRadius: "var(--ag-border-radius, 0px)",
+            fontSize: "var(--ag-font-size, 14px)",
+            fontFamily: "var(--ag-font-family, inherit)",
+            padding: "2px 4px",
+            cursor: "pointer",
+          }}
```
mfortman11: same here you can just instruct it to do so for the whole file

src/services/file_service_v2.py
```diff
@@ -0,0 +1,299 @@
...
+    "filename": "filename",
+    "file_size": "file_size",
+    "mimetype": "mimetype",
+    "indexed_time": "indexed_time",
+    "connector_type": "connector_type",
+    "embedding_model": "embedding_model",
+}
+
+# fall back to filename composite sort,then re-sort the returned page in python
+_PYTHON_SORT_FIELDS = {"chunk_count", "owner"}
```
mfortman11: _PYTHON_SORT_FIELDS (chunk_count, owner) fetches a page sorted by filename from OpenSearch, then re-sorts only that page in Python. This gives a locally sorted page, not a globally sorted result. If a user sorts by "Chunks" descending, page 1 shows the highest chunk counts among files whose filenames sort first alphabetically — not the highest chunk counts overall.

The v1 approach (fetch all, sort, slice) got global sort correct for these fields. The v2 approach silently breaks it.

frontend/app/knowledge/page.tsx
```diff
@@ -1041,72 +1048,84 @@ function SearchPage() {
...
+                  </div>
                 </div>
-              </div>
-            )}
+              )}
+            />
+          </div>
+        )}
+
+        {isWildcardQuery && serverTotal > 0 && (
```
mfortman11: This causes the footer to only be shown when using a wildcard query
So if a user does a search it disappears even if there are a lot of results
Might be good to always show it and just show 0 Pages or 0 results if that happens to be the case

mfortman11 (changes requested):

src/api/files.py
```diff
@@ -63,7 +63,7 @@ async def list_files(
 async def search_files(
     q: str = Query(..., min_length=1, description="Search query"),
     page: int = Query(1, ge=1, description="Page number"),
-    page_size: int = Query(25, ge=1, le=100, description="Items per page"),
+    page_size: int = Query(25, ge=1, description="Items per page"),
```
mfortman11: v1 search_files lost its page_size upper bound

 le=100 was removed with no replacement. Not blocking but we should probably leave that just in case

frontend/app/knowledge/page.tsx
```diff
@@ -1041,74 +1050,84 @@ function SearchPage() {
...
                 </div>
-              </div>
-            )}
-          />
+              )}
+            />
+          </div>
         )}
+
+        <KnowledgePaginationFooter
```
mfortman11, line 1124: Suggested fix prompt

In `frontend/app/knowledge/page.tsx`, the pagination footer is disconnected from search results. When the user types a non-wildcard query, the grid shows `searchFiles` from a separate search hook, but the pagination controls only affect `useListFiles` — so the footer displays page counts and navigation buttons that don't actually paginate the visible data. Fix this by routing all knowledge table data through `useListFiles`, which already hits `/api/v2/files` with the search parameter and supports server-side sort, filtering, and cursor pagination. The main change is in `page.tsx`: make effectiveData always use `listFilesData?.files` instead of branching between `listFilesData` and `searchFiles` based on `isWildcardQuery`. You'll also need to update `serverTotal` to always use `listFilesData?.total` rather than falling back to `searchFiles.length` for non-wildcard queries. Once this is done, the `isWildcardQuery` flag may no longer be needed for data-source selection (though it may still be used elsewhere in the component for UI logic — check before removing). The v2 `search_files` endpoint in `src/api/v2/files.py` and `src/services/file_service_v2.py` becomes unused by the frontend after this change but can remain for now. Verify that typing a search term in the knowledge table search bar still returns filtered results and that pagination, sort, and page-size controls all work correctly during search.

mfortman11 (changes requested):

frontend/app/knowledge/page.tsx
```diff
@@ -468,14 +470,30 @@ function SearchPage() {
...
+  const serverTotal = isWildcardQuery
+    ? (listFilesData?.total ?? 0)
+    : searchFiles.length;
+  const gridRows: File[] = fileResults;
+  const totalPages = Math.max(1, Math.ceil(serverTotal / currentPageSize));
+
+  useEffect(() => {
+    cursorCacheRef.current = new Map();
+    setCurrentPage(1);
+  }, [isWildcardQuery, searchFiles.length]);
```
mfortman11: We should trigger this on `searchFiles` instead of `searchFiles.length`

src/api/files.py
```diff
@@ -23,7 +23,7 @@ def _get_file_service(session_manager=Depends(get_session_manager)):
 
 async def list_files(
     page: int = Query(1, ge=1, description="Page number"),
-    page_size: int = Query(25, ge=1, le=100, description="Items per page"),
+    page_size: int = Query(25, ge=1, le=500, description="Items per page"),
```
mfortman11: let's just leave this at 100. If a revert is ever need best to leave as it was

mfortman11 (approved):

src/api/v2/files.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        return JSONResponse(result)
+    except Exception as e:
+        logger.error("Failed to search files (v2)", error=str(e))
+        from utils.opensearch_utils import AUTH_ERROR_MESSAGE, is_opensearch_auth_error
+
+        if is_opensearch_auth_error(e):
+            return JSONResponse({"error": AUTH_ERROR_MESSAGE}, status_code=401)
+        return JSONResponse(
+            {"error": "Failed to search files", "detail": str(e)},
```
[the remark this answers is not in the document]
VisveshJ: resolved

src/api/v2/files.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+        return JSONResponse(result)
+    except Exception as e:
+        logger.error("Failed to list files (v2)", error=str(e))
+        from utils.opensearch_utils import AUTH_ERROR_MESSAGE, is_opensearch_auth_error
+
+        if is_opensearch_auth_error(e):
+            return JSONResponse({"error": AUTH_ERROR_MESSAGE}, status_code=401)
+        return JSONResponse(
+            {"error": "Failed to list files", "detail": str(e)},
```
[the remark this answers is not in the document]
VisveshJ: resolved

PR #2172, opened 2026-07-29
fix: support images in non langflow ingestion

mfortman11:
The issue

1. The non-langflow parser never reads picture annotations. Grepping `document_processing.py `+ `processors.py` for annotation/pictures finds nothing — extract_relevant only pulls `texts[]` and `tables[]`. VLM-generated captions land in `pictures[].annotations`, which is discarded.
2. An image-only PDF yields texts: 0 (cat1.pdf → texts 0, pictures 1). Docling succeeds, but there's no text layer.

[cat2.pdf](https://github.com/user-attachments/files/30521799/cat2.pdf)
test file ^

The fix:

`src/utils/document_processing.py` — `extract_relevant` now emits a chunk per picture that has a VLM description annotation:
- Reads `pictures[].annotations`, keeps `kind == "description"` (skips classification, which has no text), joins multiple descriptions, groups by `prov[0].page_no` (defaults to page 1).
- Skips pictures with no/empty description so no blank chunks are created.

## Summary by CodeRabbit

* **New Features**
  * Image-based documents now produce searchable content from picture descriptions.
  * Multiple descriptions from the same picture are combined, while empty or unrelated annotations are ignored.
  * Picture content can be processed alongside text content, including documents without a text layer.
* **Bug Fixes**
  * Improved handling of missing page information by defaulting picture content to page 1.

### DIFF
--- a/src/utils/document_processing.py
+++ b/src/utils/document_processing.py
@@ -131,6 +131,34 @@ def extract_relevant(doc_dict: dict) -> dict:
             }
         )
 
+    # 3) process picture descriptions (VLM-generated annotations)
+    # Docling stores these under pictures[].annotations with kind == "description".
+    # Without this, image-only documents (no text layer) yield zero chunks on the
+    # non-Langflow path, which surfaces as a misleading "corrupted or invalid" error.
+    # The Langflow path already captures them via Docling's export_to_markdown.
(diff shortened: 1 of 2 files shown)


edwinjosechittilappilly (approved):
code LGTM

PR #2175, opened 2026-07-30
fix: overflows setup apiKey in settings

Wallgau:
issue: [](https://github.ibm.com/lakehouse/tracker/issues/78890)
**Summary**
- Prevent long provider validation/error messages (e.g. OpenAI echoing a masked invalid key as an unbroken string) from expanding the Setup dialog past "max-w-2xl.
- Add "break-all" on validation and mutation error text in OpenAI, Anthropic, watsonx, and Ollama settings forms/dialogs.
- Clamp dialog layout with "overflow-hidden" on "DialogContent" and "min-w-0" on the form grid so the input and Cancel/Save footer stay inside the modal.

**Before**
Screenshot 2026-07-30 at 8 22 01 AM

**After**
Screenshot 2026-07-30 at 8 18 59 AM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved settings dialogs for Anthropic, Ollama, OpenAI, and Watsonx to prevent long error messages from overflowing in narrow dialog windows.
  * Updated error text wrapping for API key, endpoint, project/model, and configuration/removal failures to wrap reliably instead of using prior overflow-prone styling.
  * Refined dialog/form layout constraints so content fits cleanly (including better sizing inside the dialog grid).

### DIFF
--- a/frontend/app/settings/_components/anthropic-settings-dialog.tsx
+++ b/frontend/app/settings/_components/anthropic-settings-dialog.tsx
@@ -173,7 +173,7 @@ const AnthropicSettingsDialog = ({
                   animate={{ opacity: 1, y: 0 }}
                   exit={{ opacity: 0, y: -10 }}
                 >
-                  <p className="rounded-lg border border-destructive p-4 break-all">
+                  <p className="rounded-lg border border-destructive p-4 min-w-0 [overflow-wrap:anywhere]">
                     {settingsMutation.error?.message}
                   </p>
                 </motion.div>
@@ -185,7 +185,7 @@ const AnthropicSettingsDialog = ({
                   animate={{ opacity: 1, y: 0 }}
                   exit={{ opacity: 0, y: -10 }}
                 >
-                  <p className="rounded-lg border border-destructive p-4 break-all">
+                  <p className="rounded-lg border border-destructive p-4 min-w-0 [overflow-wrap:anywhere]">
                     {removeMutation.error?.message}
                   </p>
                 </motion.div>
--- a/frontend/app/settings/_components/anthropic-settings-dialog.tsx
+++ b/frontend/app/settings/_components/anthropic-settings-dialog.tsx
@@ -145,9 +145,12 @@ const AnthropicSettingsDialog = ({
         setOpen(o);
       }}
     >
-      <DialogContent className="max-w-2xl">
+      <DialogContent className="max-w-2xl overflow-hidden">
         <FormProvider {...methods}>
-          <form onSubmit={handleSubmit(onSubmit)} className="grid gap-4">
+          <form
+            onSubmit={handleSubmit(onSubmit)}
+            className="grid min-w-0 gap-4"
+          >
             <DialogHeader className="mb-2">
               <DialogTitle className="flex items-center gap-3">
                 <div className="w-8 h-8 rounded flex items-center justify-center bg-white border">
@@ -170,7 +173,7 @@ const AnthropicSettingsDialog = ({
                   animate={{ opacity: 1, y: 0 }}
                   exit={{ opacity: 0, y: -10 }}
                 >
-                  <p className="rounded-lg border border-destructive p-4">
+                  <p className="rounded-lg border border-destructive p-4 break-all">
                     {settingsMutation.error?.message}
                   </p>
                 </motion.div>
@@ -182,7 +185,7 @@ const AnthropicSettingsDialog = ({
                   animate={{ opacity: 1, y: 0 }}
                   exit={{ opacity: 0, y: -10 }}
                 >
-                  <p className="rounded-lg border border-destructive p-4">
+                  <p className="rounded-lg border border-destructive p-4 break-all">
                     {removeMutation.error?.message}
                   </p>
                 </motion.div>
--- a/frontend/app/settings/_components/openai-settings-form.tsx
+++ b/frontend/app/settings/_components/openai-settings-form.tsx
@@ -41,7 +41,7 @@ export function OpenAISettingsForm({
           />
         </LabelWrapper>
         {apiKeyError && (
-          <p className="text-sm text-destructive">{apiKeyError}</p>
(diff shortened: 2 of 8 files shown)


frontend/app/settings/_components/anthropic-settings-dialog.tsx
```diff
@@ -145,9 +145,12 @@ const AnthropicSettingsDialog = ({
         setOpen(o);
       }}
     >
-      <DialogContent className="max-w-2xl">
+      <DialogContent className="max-w-2xl overflow-hidden">
```
mpawlow, line 148: ### (1a) [Minor] Fix not applied to the S3 connector dialog, which shares the identical overflow bug

#### Problem

- `s3-settings-dialog.tsx` and `s3-settings-form.tsx` have the exact same at-risk pattern this PR fixes elsewhere: a `DialogContent` capped at `max-w-2xl` and a destructive-styled `<p>` rendering an unbounded, backend/provider-supplied error string (`bucketsError`, `formError`) with no `break-all`/`break-words` and no `overflow-hidden` / `min-w-0` on the surrounding grid.
- S3 connection errors (bad access key/secret, boto3/S3-compatible endpoint errors, or an echoed masked secret similar to the OpenAI case described in the PR body) can just as easily be long, unbroken strings and blow out the dialog width the same way.
- This PR fixes 4 of 5 dialogs sharing this exact CSS shape, leaving the fix inconsistently applied across the settings surface.

#### Background Information

- Confirmed via `grep -rln "max-w-2xl" app/settings/_components/ | xargs grep -l DialogContent` → 5 matches: `openai`, `anthropic`, `ollama`, `watsonx` (all fixed) and `s3` (not touched).
- `s3-settings-dialog.tsx` already uses a different overflow strategy (`max-h-[90vh] overflow-y-auto`) rather than `overflow-hidden`, so the fix would need slight adaptation rather than a blind copy-paste, but the horizontal overflow risk on the error `<p>` is the same class of bug.

#### Code References

- `frontend/app/settings/_components/s3-settings-dialog.tsx:151` (`DialogContent` — no `overflow-hidden`, form has no `min-w-0`)
- `frontend/app/settings/_components/s3-settings-form.tsx:154-163` (`bucketsError` / `formError` `<p>` — no `break-all`/`break-words`)

#### Potential Solution

- Add `break-all` (or `break-words`, see next finding) to the `bucketsError` and `formError` paragraphs in `s3-settings-form.tsx`.
- Add `min-w-0` to the S3 dialog's `form` grid to match the pattern used in the other four dialogs.

#### Alternative Solutions

- If this is intentionally scoped to "model provider" dialogs only (S3 is a connector, not a model provider) and a follow-up PR is planned for connector dialogs, note that explicitly in the PR description so reviewers don't need to guess whether it was an oversight.

frontend/app/settings/_components/openai-settings-form.tsx
```diff
@@ -41,7 +41,7 @@ export function OpenAISettingsForm({
           />
         </LabelWrapper>
         {apiKeyError && (
-          <p className="text-sm text-destructive">{apiKeyError}</p>
+          <p className="text-sm text-destructive break-all">{apiKeyError}</p>
```
mpawlow: ### (1b) [Minor] `break-all` breaks words mid-character rather than only at natural boundaries, unlike existing codebase precedent

#### Problem

- `break-all` (`word-break: break-all`) will break *any* line that doesn't fit, including breaking ordinary English words in the middle, not just the long unbroken token (e.g., masked API key) that actually causes the overflow.
- The codebase already has precedent for a softer alternative elsewhere: `frontend/app/chat/_components/user-message.tsx` uses `break-words overflow-wrap-anywhere`, and `frontend/components/task-dialog/file-error-details.tsx` uses `break-words` for similar backend-error text. `break-words` (`overflow-wrap: break-word`) only breaks mid-word as a last resort, preserving readability of the normal sentence portions of the message while still wrapping an unbroken token like a masked key.
- Net effect: this PR's fix works, but on a wrapped multi-word error message it will look choppier than necessary compared to the pattern already used for the same kind of content (long/user-controlled text) elsewhere in the app.

#### Code References

- `frontend/app/settings/_components/openai-settings-form.tsx:44`
- `frontend/app/settings/_components/anthropic-settings-form.tsx:44`
- `frontend/app/settings/_components/ollama-settings-form.tsx:43`
- `frontend/app/settings/_components/watsonx-settings-form.tsx:83-146`
- Same pattern repeated in all four `*-settings-dialog.tsx` files on the `settingsMutation.error?.message` / `removeMutation.error?.message` paragraphs.

#### Potential Solution

- Swap `break-all` → `break-words` (optionally combined with `overflow-wrap-anywhere` as done in `user-message.tsx`) to match existing conventions and improve readability of wrapped error text.

#### Alternative Solutions

- Keep `break-all` if the team has a specific reason to guarantee no horizontal overflow under any circumstance (e.g., a pathological string with zero break opportunities even under `break-words`, such as `text-embedding-3-large-...` with no spaces) — `break-all` is a strictly stronger guarantee, just less pretty. This is a legitimate, defensible choice, just inconsistent with nearby code.

lucaseduoli (approved):

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
- See Minor PR comments: (1a), (1b) for optional considerations
   - (1a) can be considered out of scope
   - (1b) can be considered an optimization but maybe not desirable

frontend/app/settings/_components/anthropic-settings-dialog.tsx
Wallgau: out of scope for this PR

frontend/app/settings/_components/openai-settings-form.tsx
Wallgau: Good catch, will change it

frontend/app/settings/_components/anthropic-settings-dialog.tsx
mpawlow: Resolving accordingly.

PR #2184, opened 2026-07-30
fix: Fetch latest documentation and apply style fixes

edwinjosechittilappilly:
This pull request introduces permission-based access control for actions related to refreshing documentation in the knowledge management UI. By wrapping the "Fetch latest docs" buttons with the `RequirePermission` component and specifying the `config:write` permission, only authorized users can trigger these actions. Minor formatting improvements were also made for consistency.

**Permission Enforcement for Refresh Actions:**

* Wrapped the "Fetch latest docs" button in `frontend/app/knowledge/page.tsx` with the `RequirePermission` component, restricting access to users with `config:write` permission. [[1]](diffhunk://#diff-60146564fb7979730b4a32d5e921417e85149f7cd99bccceddbfe042345b505fR37) [[2]](diffhunk://#diff-60146564fb7979730b4a32d5e921417e85149f7cd99bccceddbfe042345b505fR957) [[3]](diffhunk://#diff-60146564fb7979730b4a32d5e921417e85149f7cd99bccceddbfe042345b505fR989)
* Applied the same permission check to the corresponding button in `frontend/components/knowledge-search-bar.tsx`. [[1]](diffhunk://#diff-229892fad705143238b083dbf547db55ae8a9983a0d1ae91c1d95918d8e50244R10) [[2]](diffhunk://#diff-229892fad705143238b083dbf547db55ae8a9983a0d1ae91c1d95918d8e50244R170) [[3]](diffhunk://#diff-229892fad705143238b083dbf547db55ae8a9983a0d1ae91c1d95918d8e50244R194)

**Code Formatting:**

* Improved formatting for the async mutation call in the refresh handler for better readability.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted the “Fetch latest docs” and “Refreshing docs…” actions to users with the required configuration permission.
  * Users without permission can no longer view or use the documentation refresh controls.

PR #2185, opened 2026-07-30
fix: gate fetch latest docs by permission

edwinjosechittilappilly:
## Summary
- backport the UI permission gating from #2184 to `main`
- hide **Fetch latest docs** unless the user has `config:write`
- apply the gate to both standard and cloud knowledge-page layouts
- exclude release-only Docker changes

## Test plan
- [x] Confirm the diff against `main` contains only the two knowledge UI files
- [x] Run editor lint diagnostics on both modified files
- [ ] Verify an admin can see and use **Fetch latest docs**
- [ ] Verify developer, user, and viewer roles cannot see the action
- [ ] Verify RBAC-disabled/no-auth behavior remains unchanged

## Summary by CodeRabbit

* **Access Control**
  * Restricted the “Fetch latest docs” action to users with the required configuration permission.
  * Users without permission no longer see or can trigger the document refresh action.
  * Kept the existing refresh behavior, notifications, and loading states unchanged for authorized users.

### DIFF
--- a/frontend/app/knowledge/page.tsx
+++ b/frontend/app/knowledge/page.tsx
@@ -973,7 +973,8 @@ function SearchPage() {
                   });
                   try {
                     toast.info("Refreshing OpenRAG docs...");
-                    const result = await refreshOpenragDocsMutation.mutateAsync();
+                    const result =
+                      await refreshOpenragDocsMutation.mutateAsync();
                     toast.success(result.message);
                   } catch (error) {
                     toast.error(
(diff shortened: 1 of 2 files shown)


frontend/app/knowledge/page.tsx
mfortman11, line 1: Not sure we need this change here (oss) but doesn't hurt

mfortman11 (approved):

PR #2167, opened 2026-07-28
refactor: Filter panel improvements

mfortman11:
Fixes how filters applied.
Previously source file filters didn't filter the knowledge table when selected.

UI improvements

Adds error state for filters attempted to be created w/o a name
Screenshot 2026-07-28 at 4 28 26 PM

Matches dropdown width to the input button and truncates long file names, as well as removing the number count (always 1)
Screenshot 2026-07-28 at 4 28 34 PM

Also shuffles the filter icon color

## Summary by CodeRabbit

- **New Features**
  - New filters now receive a varied default color when created.
  - Active source filters are listed alphabetically without displaying document counts.
- **Bug Fixes**
  - Filtered searches now correctly honor active filters, even when the search text is empty or uses a wildcard.
  - Facet bucket and source option counts are handled more defensively when data is missing.
- **UI Improvements**
  - Filter dropdowns now match their trigger width and truncate long labels cleanly.
  - “Created” timestamps are hidden while creating a filter or when validation errors are present.
  - Filter icon options use a shared icon/color palette consistently.

PR #2187, opened 2026-07-30
refactor: Filter panel improvements (#2167)

mfortman11:
* filtering, error state, and input changes

* design for icon color and error

* coderabbit cleanup

## Summary by CodeRabbit

- **New Features**
  - Knowledge filters now correctly apply when browsing without search text.
  - New filter color and icon options are shared consistently across the interface.
  - New filters receive a varied default color.

- **Improvements**
  - Filter options are sorted consistently and display cleaner labels.
  - Multi-select menus align with their controls and match their width.
  - Long filter labels are truncated to prevent layout overflow.
  - Filter creation and editing now clear name errors appropriately.
  - Creation screens no longer display an unavailable “Created” timestamp.

### DIFF
--- a/frontend/components/knowledge-filter-panel.tsx
+++ b/frontend/components/knowledge-filter-panel.tsx
@@ -10,11 +10,7 @@ import {
   EMPTY_SEARCH_RESULT,
   useGetSearchQuery,
 } from "@/app/api/queries/useGetSearchQuery";
-import {
-  type FilterColor,
-  FilterIconPopover,
-  type IconKey,
-} from "@/components/filter-icon-popover";
+import { FilterIconPopover } from "@/components/filter-icon-popover";
 import { Button } from "@/components/ui/button";
 import {
   Card,
@@ -32,6 +28,7 @@ import { useKnowledgeFilter } from "@/contexts/knowledge-filter-context";
 import { useTask } from "@/contexts/task-context";
 import { usePermissions } from "@/hooks/use-permissions";
 import { trackButton } from "@/lib/analytics";
+import type { FilterColor, IconKey } from "@/lib/filter-constants";
 import {
   buildActiveSourceOptions,
   buildKnowledgeTableRows,
@@ -39,7 +36,7 @@ import {
 
 interface FacetBucket {
   key: string;
-  count: number;
+  count?: number;
 }
 
 interface AvailableFacets {
@@ -118,6 +115,7 @@ export function KnowledgeFilterPanel() {
   // Load current filter data into controls when a filter is selected
   useEffect(() => {
     if (selectedFilter && parsedFilterData) {
+      setNameError(null);
       setQuery(parsedFilterData.query || "");
 
       // Set the actual filter selections from the saved knowledge filter
@@ -146,6 +144,7 @@ export function KnowledgeFilterPanel() {
   // Initialize defaults when entering create mode
   useEffect(() => {
     if (createMode && parsedFilterData) {
+      setNameError(null);
       setQuery(parsedFilterData.query || "");
       // Provide defaults for missing filter fields
       const filters = parsedFilterData.filters || {};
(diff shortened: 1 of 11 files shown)


frontend/components/knowledge-filter-panel.tsx
```diff
@@ -179,11 +178,13 @@ export function KnowledgeFilterPanel() {
...
   useEffect(() => {
     if (!aggregations) return;
+    const extractKeys = (buckets: { key: string }[] = []): FacetBucket[] =>
+      buckets.map((b) => ({ key: b.key }));
     const facets = {
-      data_sources: aggregations.data_sources?.buckets || [],
-      document_types: aggregations.document_types?.buckets || [],
-      owners: aggregations.owners?.buckets || [],
-      connector_types: aggregations.connector_types?.buckets || [],
+      data_sources: extractKeys(aggregations.data_sources?.buckets),
```
mpawlow, line 184: ### (1b) [Minor] `availableFacets.data_sources` is computed but never rendered

#### Problem

- `KnowledgeFilterPanel` fetches and stores `data_sources` buckets from `useGetSearchAggregations` into `availableFacets.data_sources`, but no JSX in the component reads that field — the "sources" `MultiSelect` is driven entirely by `buildActiveSourceOptions(tableRows)` instead.
- This is dead state: an extra aggregation dimension is parsed and stored on every `aggregations` change for no consumer.

#### Code References

- `frontend/components/knowledge-filter-panel.tsx:42-47` (`AvailableFacets` interface)
- `frontend/components/knowledge-filter-panel.tsx:179-190` (effect populating `data_sources`)
- Confirmed via grep: only `document_types`, `owners`, `connector_types` are read from `availableFacets` anywhere in the file (lines 393, 410, 423).

#### Potential Solution

- Drop `data_sources` from `AvailableFacets`/the `extractKeys` effect if the `sourceOptions`-from-`tableRows` approach is the intended long-term source of truth.

#### Alternative Solutions

- If `data_sources` from aggregations was meant to eventually replace `buildActiveSourceOptions`, wire it into the sources `MultiSelect` instead of deleting it — but as-is, one of the two code paths is unused and should not ship in both forms.

frontend/lib/knowledge-table-state.ts
```diff
@@ -4,7 +4,7 @@ import type { TaskFile } from "@/contexts/task-context";
 export interface KnowledgeSourceOption {
   value: string;
   label: string;
-  count: number;
+  count?: number;
```
mpawlow, line 7: ### (1c) [Minor] Per-option "count" badge support is now dead code

#### Problem

- `MultiSelect`'s `Option.count` rendering (`option.count !== undefined && <span>…count…</span>`) is unreachable: every caller building options for the four facet dropdowns explicitly stopped passing `count` in this PR (the `count: bucket.count` lines were removed for `document_types`, `owners`, `connector_types`; `buildActiveSourceOptions` no longer computes it either).
- `FacetBucket.count`, `KnowledgeSourceOption.count`, and `Option.count` are now optional fields that are always `undefined` in this codebase's only consumer.

#### Background Information

- This looks like a deliberate removal of facet counts (possibly because counts didn't reflect the currently-filtered subset, or a perf trade-off), but the removal was only done at the call sites — the display feature and its types were left in place instead of being deleted, per repo convention against half-finished/dead code paths.

#### Code References

- `frontend/components/ui/multi-select.tsx:21-25,160-164`
- `frontend/lib/knowledge-table-state.ts:4-8,277-292` (`KnowledgeSourceOption.count`, no longer populated)
- `frontend/components/knowledge-filter-panel.tsx:230-234,393-434` (`FacetBucket.count` optional, never populated)

#### Potential Solution

- If counts are intentionally gone, remove `count` from `Option`, `FacetBucket`, and `KnowledgeSourceOption`, and delete the now-unreachable rendering branch in `multi-select.tsx`.

#### Alternative Solutions

- If counts are meant to return later (e.g. pending a backend fix), leave a short comment noting why they were removed, so the dead branch reads as "intentionally deferred" rather than as leftover cruft.

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
- See Minor PR comments: (1b), (1c) for optional consideration

PR #2191, opened 2026-07-31
fix: resolve embedding-model fallbacks from deployment env, never guess

edwinjosechittilappilly:
## Summary
- `_default_embedding_model()` previously hardcoded a per-provider guess (e.g. `text-embedding-3-small` for `openai`) used whenever a provider removal left another provider as the embedding fallback. A provider name doesn't reliably say which models are actually deployed — it's often an internal OpenAI/watsonx-compatible gateway with a curated model subset. The hardcoded guess silently selected a model the gateway didn't serve, breaking every ingestion with no clear signal why (incident: gateway only served `text-embedding-3-large`; 25/25 files failed at the Langflow embedding step).
- Added `get_declared_default_embedding_model()`, which resolves the deployment's own declared default from `EMBEDDING_MODEL` / `EMBEDDING_PROVIDER` (Helm values / operator ConfigMap) instead of a hardcoded string, and returns `""` when the deployment hasn't declared one for that provider. Read directly rather than through `ConfigManager`'s env-override path, which is skipped once `config.edited` is `True` — a protection meant for a user's own prior choices, not a reason to block resolving an unmade choice.
- Wired this into every "guess an embedding model" call site, not just the settings-save path: `search_service.py`, `utils/embeddings.py`, and `langflow_file_service.py` all previously fell back to the same hardcoded constant when no model was configured; they now fall back to the deployment-declared default (or none) instead.

## Note
This is a necessary but not sufficient fix for the specific incident that motivated it — the affected cluster's `EMBEDDING_MODEL`/`EMBEDDING_PROVIDER` env vars are currently unset, so it also needs an infra-side change (outside this repo) to actually populate them from a correct value.

## Test plan
- [x] `uv run pytest tests/unit/test_settings_provider_removal_defaults.py tests/unit/test_embedding_fields.py tests/unit/test_skip_os_security_setup_init_index.py -q` — 48 passed; 3 pre-existing failures in unrelated `_default_llm_model` tests, confirmed present identically on unmodified `release-cpd-0.6.0` before this change.
- [x] Verified diff is empty-conflict cherry-pick from a `main`-based version of the same fix onto `release-cpd-0.6.0`.


edwinjosechittilappilly:
@coderabbitai mention the risk of merging the pr and does it need full regression testing of onboarding and ingestions?

ricofurtado (approved):
LGTM - It is working for me.

PR #2192, opened 2026-07-31
fix: resolve embedding-model fallbacks from deployment env, never guess (backport of #2191)

edwinjosechittilappilly:
Backport of #2191 to \`main\`. Clean cherry-pick, no conflicts.

## Summary
- \`_default_embedding_model()\` previously hardcoded a per-provider guess (e.g. \`text-embedding-3-small\` for \`openai\`) used whenever a provider removal left another provider as the embedding fallback. A provider name doesn't reliably say which models are actually deployed — it's often an internal OpenAI/watsonx-compatible gateway with a curated model subset. The hardcoded guess silently selected a model the gateway didn't serve, breaking every ingestion with no clear signal why (incident: gateway only served \`text-embedding-3-large\`; 25/25 files failed at the Langflow embedding step).
- Added \`get_declared_default_embedding_model()\`, which resolves the deployment's own declared default from \`EMBEDDING_MODEL\` / \`EMBEDDING_PROVIDER\` (Helm values / operator ConfigMap) instead of a hardcoded string, and returns \`""\` when the deployment hasn't declared one for that provider. Read directly rather than through \`ConfigManager\`'s env-override path, which is skipped once \`config.edited\` is \`True\` — a protection meant for a user's own prior choices, not a reason to block resolving an unmade choice.
- Wired this into every "guess an embedding model" call site, not just the settings-save path: \`search_service.py\`, \`utils/embeddings.py\`, and \`langflow_file_service.py\` all previously fell back to the same hardcoded constant when no model was configured; they now fall back to the deployment-declared default (or none) instead.

## Test plan
- [x] \`uv run pytest tests/unit/test_settings_provider_removal_defaults.py tests/unit/test_embedding_fields.py tests/unit/test_skip_os_security_setup_init_index.py -q\` — 48 passed; 3 pre-existing failures in unrelated \`_default_llm_model\` tests, confirmed present identically on unmodified \`main\` before this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Embedding model selection now respects the deployment’s declared provider and model configuration.
  * Removed automatic fallback to a hardcoded OpenAI embedding model when no model is declared.
  * Search, indexing, and ingestion workflows now use consistent provider-specific embedding settings.

* **Tests**
  * Added coverage for declared embedding defaults and provider-specific behavior.


ricofurtado (approved):
LGTM

PR #2216, opened 2026-08-06
fix: add better citation prompt, add upgrade prompt functionality

lucaseduoli:
This pull request improves the handling and consistency of the OpenRAG Agent's system prompt across the backend, frontend, and configuration files. The main enhancements include stricter citation requirements, a new "Restore Default" button for the system prompt in the UI, and centralization of the default prompt in backend code for maintainability.

**System Prompt and Citation Policy Updates:**

* The system prompt for the OpenRAG Agent was updated everywhere to clarify that citations are mandatory for every factual claim. Citations must be inline (e.g., `(Source: <chunk_id>)`), and bibliographies or descriptive source lists are explicitly disallowed. [[1]](diffhunk://#diff-c3017dc80bfe15c829a85817f843fd4d084615890e488bd7cdc80857436dcaecL1512-R1512) [[2]](diffhunk://#diff-8b88658adc69c3bf02cd711c6b17ec1f8f9bdc41a49b9e43c7df5c7e9bcd6126L7-R7)
* The prompt now includes a clear instruction to ignore any directives found within untrusted document data blocks, further preventing prompt injection. [[1]](diffhunk://#diff-c3017dc80bfe15c829a85817f843fd4d084615890e488bd7cdc80857436dcaecL1512-R1512) [[2]](diffhunk://#diff-8b88658adc69c3bf02cd711c6b17ec1f8f9bdc41a49b9e43c7df5c7e9bcd6126L7-R7)

**Frontend Enhancements:**

* Added `default_system_prompt` to the agent settings API and UI.
* Introduced a "Restore Default" button in the agent settings section, allowing users to easily revert the system prompt to its default value.

**Backend Improvements:**

* Centralized the default system prompt in a single location (`DEFAULT_SYSTEM_PROMPT`), which is now imported and used in the agent logic, improving maintainability and consistency. [[1]](diffhunk://#diff-3f21ad1752eea52d46c1770fecd891a347976b44142baa2a02235d239cc6e016R5) [[2]](diffhunk://#diff-3f21ad1752eea52d46c1770fecd891a347976b44142baa2a02235d239cc6e016L142-R143)

These changes collectively ensure that the agent's behavior is more predictable, easier to update, and that factual claims are always properly sourced.

## Summary by CodeRabbit

- **New Features**
  - Added a **Restore Default** option for agent instructions when standard instructions are available.
  - New conversations automatically use the latest default agent instructions.

- **Improvements**
  - Agent responses now require precise inline citations for factual claims.
  - Retrieved content is treated as untrusted data, preventing embedded instructions from affecting responses or tool use.
  - Improved guidance for context, files, URLs, retrieval, and calculator use.
  - Default instructions synchronize automatically while preserving customized prompts.
  - Improved reliability when retrieving or updating agent instructions.


src/services/startup_orchestrator.py
```diff
@@ -175,6 +175,10 @@ async def startup_tasks(services) -> None:
     try:
         flows_service = services["flows_service"]
         await flows_service.ensure_flows_exist()
+        
+        # Ensure the system prompt is synced to Langflow (in case of legacy prompt auto-upgrades)
+        await flows_service.update_chat_flow_system_prompt(get_openrag_config().agent.system_prompt)
```
ricofurtado: startup now unconditionally patches the existing Langflow retrieval flow’s system_prompt from config. That breaks the existing create-only/preserve-edits contract around ensure_flows_exist(), and can wipe a prompt edited directly in Langflow on every backend restart, even when it is not a legacy prompt. I’d gate this to actual legacy-prompt migration, or read the current flow prompt and only replace it when it matches LEGACY_SYSTEM_PROMPTS.

ricofurtado (approved):
Hi Lucas, what is the intended source of truth for the Agent system_prompt after this change?
ensure_flows_exist() is documented as create-only and preserving existing Langflow edits, but startup now always calls update_chat_flow_system_prompt(get_openrag_config().agent.system_prompt) for the retrieval flow. That means any prompt edited directly in Langflow can be overwritten on backend restart by the config value, even if the Langflow prompt is not one of the legacy prompts.
Is that overwrite intentional because config should now be canonical, or should this startup sync be limited to migration cases, for example only replacing the Langflow prompt when the current flow value matches one of LEGACY_SYSTEM_PROMPTS?

PR #2223, opened 2026-08-11
feat: v2 public facing endpoint

jason-dong03:
-2 new endpoints: v2/files and v2/files/search?q=
-implemented file ingestion endpoint based on a v2 endpoint for consistency / future roadmaps
-currently calls the same function from v1's file ingestion endpoint 

**Future Note**:
-right now, v1 and v2 share the same composite agg endpoint call (both public endpoint calls from v2's file ingestion); therefore, the code right now is just duplicated. If we were to keep it this way, we can create shared helpers to optimize code flow and reduce duplications.
-since the code duplicated, integration testing, sdk testing, mcp testing all yield same results as v1/files/ and v1/files/search

## Summary by CodeRabbit

* **New Features**
  * Added v2 file APIs for listing and searching files.
  * Added support for pagination, sorting, filtering, search terms, and cursor-based navigation.
  * Added API-key authentication for v2 file operations.
  * Exposed v2 file listing and search through MCP tools with dedicated names and descriptions.

### DIFF
--- a/src/app/routes/__init__.py
+++ b/src/app/routes/__init__.py
@@ -9,6 +9,7 @@ from fastapi import FastAPI
 from app.routes.internal import register_internal_routes
 from app.routes.mcp import mount_mcp
 from app.routes.public_v1 import register_public_v1_routes
+from app.routes.public_v2 import register_public_v2_routes
 
 
 def register_all_routes(app: FastAPI):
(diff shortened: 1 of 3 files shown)


Wallgau (approved):
LGTM

PR #2231, opened 2026-08-12
feat: expose GET /v1/files on public API surface with SDK support and…

jason-dong03:
* feat: public ingested files endpoint
-Added public file listing endpoints (`GET /v1/files` and `/v1/files/search`). 
-Exposed file listing for API key, SDK, and MCP clients. 
-Added public v1 file routes and delegated to the v2 file service. 
-Registered new MCP file listing and search endpoints. 
-Fixed case-insensitive filename search.
-Added file listing models to the Python SDK.
-Added `list_files()` to the Python SDK.
-Exported new file listing models in the Python SDK. -Added file listing examples to the Python SDK docs.
 -Added file listing types to the TypeScript SDK.
-Added `listFiles()` to the TypeScript SDK.
-Exported new file listing types in the TypeScript SDK. -Added file listing examples to the TypeScript SDK docs. 
-Added unit tests for the new v1 file endpoints.
-Added Python SDK integration tests for file listing. -Added TypeScript SDK integration tests for file listing. -Updated API documentation with file listing examples and pagination.
-case-insensitive filename search fix


Wallgau (approved):

src/app/routes/public_v2.py
```diff
@@ -0,0 +1,22 @@
+"""Public /v2/* route registrations (API-key auth)."""
+
```
Wallgau, line 2: is dropping the _list_files / _search_files API-key wrappers intentional here? 
Latest main still uses them with require_api_key_permission("knowledge:read:own"), while this registers the cookie-auth handlers directly.

tests/unit/api/test_v1_files.py
```diff
@@ -0,0 +1,350 @@
...
+- 500 returned on unexpected errors
+"""
+
+import inspect
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from fastapi import params as fastapi_params
+
+from api.v1.files import list_files, search_files
```
Wallgau, line 19: This still imports list_files and search_files from api.v1.files, but that module only defines get_all_files now. Should these tests be rewritten for getAll, or will collection fail with ImportError?

src/app/routes/public_v2.py
```diff
@@ -0,0 +1,22 @@
...
+    # Files endpoints (composite-agg pagination)
+    app.add_api_route(
+        "/v2/files/search",
+        v2_files.search_files,
+        methods=["GET"],
+        tags=["public"],
+    )
+    app.add_api_route(
+        "/v2/files",
+        v2_files.list_files,
```
Wallgau, line 19: internal.py already registers GET /v2/files with cookie auth, and register_all_routes registers the internal routes before the public ones. Since Starlette uses first-match-wins routing, wouldn’t an SDK request to GET /api/v2/files hit the internal cookie-auth route even if we restore the API-key wrappers here?

main doesn’t appear to have this collision because the SDK uses /v1/files, whereas this PR switches it to /v2/files.

I think the same issue applies to /v2/files/search here vs. internal.py:136.

Wallgau (commented):
just need few clarifications @edwinjosechittilappilly can you take a look to my comments please

tests/unit/api/test_v1_files.py
edwinjosechittilappilly: Agreed. The suggetsed changes as disscussed earlier to be merged to main first and then as cheerry pick to release branch

PR #2238, opened 2026-08-12
fix: onboarding progress bar is centered again

Vchen7629:
## Summary
fixes a regression introduced by pr https://github.com/langflow-ai/openrag/pull/2208. The feature adds a new unconditional relative to chat-renderer which shifted the progress bar that appears during onboarding to the right and non-centered. I made this relative conditional on isSelectingChats being true so the progress bar during onboarding is centered again

## Before
regression-before

## After
regression-after

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat layout behavior by applying positioning styles only when chat selection mode is active.

### DIFF
--- a/frontend/components/chat-renderer.tsx
+++ b/frontend/components/chat-renderer.tsx
@@ -302,7 +302,12 @@ export function ChatRenderer({
       </div>
 
       {/* Main Content */}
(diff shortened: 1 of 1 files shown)


VisveshJ:
functionality works great, good feature and fix!

VisveshJ (approved):

PR #2242, opened 2026-08-13
feat: Knowledge table filtering

jason-dong03:
- changed filter params to accept repeated params
- added data_source to filter params as an option from knowledge filter
-safe guarded sentinel value * to not be taken as a literal
-listFilesFilterParam no longer reduces array to one string; replaced with repeated-param serialisation
-Fix: knowledge filter facet counts and option limits

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting multiple connector types, document types, owners, and data sources when browsing or searching files.
  * Added data-source filtering for file searches and listings.
  * Filter options now display matching file counts.

* **Bug Fixes**
  * Improved facet counts and filtering accuracy for exact-match searches.
  * Ensured counts are based on files rather than duplicate content segments.
  * Improved wildcard data-source handling and multi-value filter behavior.

### DIFF
--- a/frontend/app/knowledge/page.tsx
+++ b/frontend/app/knowledge/page.tsx
@@ -137,6 +137,22 @@ function listFilesFilterValues(values?: string[]) {
   return filtered && filtered.length > 0 ? filtered : undefined;
 }
 
+function buildFilterPageResetKey(
+  parsedFilterData: ReturnType<typeof useKnowledgeFilter>["parsedFilterData"],
+) {
+  if (!parsedFilterData) {
+    return "";
+  }
+
+  return JSON.stringify({
+    query: parsedFilterData.query,
+    connector_types: parsedFilterData.filters.connector_types,
+    document_types: parsedFilterData.filters.document_types,
+    owners: parsedFilterData.filters.owners,
+    data_sources: parsedFilterData.filters.data_sources,
+  });
+}
+
 function SearchPage() {
   const isCloudBrand = useIsCloudBrand();
   const queryClient = useQueryClient();
@@ -337,6 +353,7 @@ function SearchPage() {
   const isWildcardQuery =
     (effectiveSearchText === "" || effectiveSearchText === "*") &&
     !hasActiveFilters;
+  const filterPageResetKey = buildFilterPageResetKey(parsedFilterData);
 
   const {
     data: listFilesData,
@@ -494,7 +511,7 @@ function SearchPage() {
   useEffect(() => {
     cursorCacheRef.current = new Map();
     setCurrentPage(1);
-  }, [effectiveSearchText]);
+  }, [effectiveSearchText, filterPageResetKey]);
 
   // when the server responds with an after_key for page N, cache it as the cursor for page N+1
   useEffect(() => {
--- a/frontend/app/knowledge/page.tsx
+++ b/frontend/app/knowledge/page.tsx
@@ -132,6 +132,11 @@ const AG_FIELD_TO_SORT_BY: Record<string, string> = {
   status: "status",
 };
 
+function listFilesFilterValues(values?: string[]) {
+  const filtered = values?.filter((value) => value !== "*");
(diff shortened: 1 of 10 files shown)


frontend/app/knowledge/page.tsx
```diff
@@ -355,11 +346,18 @@ function SearchPage() {
...
       sortOrder,
       afterKey: cursorCacheRef.current.get(currentPage) ?? null,
-      connectorType: listFilesFilterParam(
-        parsedFilterData?.filters?.connector_types,
-      ),
-      mimetype: listFilesFilterParam(parsedFilterData?.filters?.document_types),
-      owner: listFilesFilterParam(parsedFilterData?.filters?.owners),
+      connectorType:
+        parsedFilterData?.filters?.connector_types?.filter((v) => v !== "*") ||
+        undefined,
```
mfortman11: nit: `.filter((v) => v !== "*") || undefined` never yields `undefined` because empty arrays are truthy in JavaScript; the `|| undefined` fallback is dead code

frontend/components/knowledge-filter-panel.tsx
```diff
@@ -393,7 +381,10 @@ export function KnowledgeFilterPanel() {
                 options={(availableFacets.document_types || []).map(
                   (bucket) => ({
                     value: bucket.key,
-                    label: bucket.key,
+                    label:
+                      typeof bucket.count === "number"
+                        ? `${bucket.key} (${bucket.count})`
+                        : bucket.key,
```
mfortman11: nit: this is small but duplicated across each dropdown could be made resuable

src/services/file_service_v2.py
```diff
@@ -217,25 +221,49 @@ async def search_files(
...
+            filter_clauses.append(clause)
+
+        effective_sources = [s for s in (data_sources or []) if s != "*"]
+        if effective_sources:
+            clause = (
+                {"term": {"filename": effective_sources[0]}}
+                if len(effective_sources) == 1
+                else {"terms": {"filename": effective_sources}}
+            )
+            filter_clauses.append(clause)
```
mfortman11: Only `data_sources` strips the `"*"` wildcard sentinel; `connector_type`, `mimetype`, and `owner` pass `"*"` through as a literal OpenSearch term filter that matches zero documents.

src/services/search_service.py
```diff
@@ -37,6 +37,70 @@ def _is_exact_token_query(query: str) -> bool:
...
+                    **(
+                        {}
+                        if facet_name == "data_sources"
+                        else {
+                            "doc_count": bucket.get("files", {}).get(
+                                "value", bucket.get("doc_count", 0)
+                            )
+                        }
+                    ),
+                }
```
mfortman11: The normalizer at line 88-89 deliberately strips `doc_count` from `data_sources` buckets. But when a token-like query triggers `_apply_exact_match_file_filter`, line 163 rebuilds all facets with `doc_count`. So the frontend will show "`config.yaml (1)`" for token queries but just "`config.yaml`" for prose queries — the count flickers on/off based on query shape.

mfortman11: On second though this isn't very important since the UI doesn't show these counts

mfortman11 (commented):
Functional review

Pagination doesn't reset on filter change
- create a couple filters
- select a filter and go to page 2
- switch filters if the new filter is small you end up on an empty page

Plural vs Singular for filter selections (small cleanup would be nice to add to this PR)
Image

Show Owner Name instead of ID
Image

VisveshJ (commented):
the knowledge table already showed readable owner names, but the Owner dropdown in the filter panel was still displaying the raw owner ID. I updated the search aggregation response to keep the owner key for filtering while adding a label that can show owner_name / owner_email. The dropdown should now shows owner names with file counts

mfortman11 (approved):
Lookin good 😎

PR #2253, opened 2026-08-17
fix: CORS misconfiguration: replace wildcard Access-Control-Allow-Origin with explicit origin allowlist

mfortman11:
Issue https://github.ibm.com/lakehouse/tracker/issues/79665

## Summary by CodeRabbit

* **New Features**
  * Added configurable cross-origin access for frontend applications.
  * Supports multiple comma-separated allowed origins.
  * Defaults to local frontend access during development.
  * Configuration is available across local and Kubernetes deployments.

* **Security**
  * Cross-origin access can be disabled when no origins are configured.
  * Requests from unapproved origins are not granted cross-origin access.
  * Consistent permissions apply across supported API responses.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,7 +64,7 @@ OPENRAG_BACKEND_PORT=8000
 # --- CORS -------------------------------------------------------------------
 # Comma-separated list of origins allowed to make cross-origin requests
 # (e.g. "https://app.example.com,https://admin.example.com").
-# Leave empty or unset to forbid all cross-origin requests.
+# Unset → defaults to http://localhost:3000.  Set to "" to disable CORS entirely.
 CORS_ALLOWED_ORIGINS=http://localhost:3000
 
 # --- Backend ingestion-callback proxy router --------------------------------
--- a/.env.example
+++ b/.env.example
@@ -65,7 +65,7 @@ OPENRAG_BACKEND_PORT=8000
 # Comma-separated list of origins allowed to make cross-origin requests
 # (e.g. "https://app.example.com,https://admin.example.com").
 # Leave empty or unset to forbid all cross-origin requests.
-# CORS_ALLOWED_ORIGINS=
+CORS_ALLOWED_ORIGINS=http://localhost:3000
 
 # --- Backend ingestion-callback proxy router --------------------------------
 # When enabled, a tiny standalone uvicorn app is started in the same process as
--- a/.env.example
+++ b/.env.example
@@ -61,6 +61,12 @@ OPENRAG_BACKEND_PORT=8000
 # if you need to override the auto-derived value (e.g., custom hostname).
 # OPENRAG_BACKEND_INTERNAL_URL=http://openrag-backend:8000
 
+# --- CORS -------------------------------------------------------------------
+# Comma-separated list of origins allowed to make cross-origin requests
+# (e.g. "https://app.example.com,https://admin.example.com").
+# Leave empty or unset to forbid all cross-origin requests.
+# CORS_ALLOWED_ORIGINS=
+
 # --- Backend ingestion-callback proxy router --------------------------------
 # When enabled, a tiny standalone uvicorn app is started in the same process as
 # the backend (its own port) that proxies ONLY the Langflow ingest callback
--- a/src/app/factory.py
+++ b/src/app/factory.py
@@ -5,13 +5,14 @@ into a ready-to-serve FastAPI app.
 """
 
 from fastapi import FastAPI
+from fastapi.middleware.cors import CORSMiddleware
 from prometheus_fastapi_instrumentator import Instrumentator
 
 from app.container import initialize_services
 from app.lifespan import run_shutdown, run_startup
 from app.middleware import RequestLoggingMiddleware
 from app.routes import register_all_routes
-from config.settings import FASTAPI_DEBUG
+from config.settings import CORS_ALLOWED_ORIGINS, FASTAPI_DEBUG
 from utils.logging_config import get_logger
 from utils.version_utils import OPENRAG_VERSION
 
@@ -25,6 +26,15 @@ async def create_app():
     app = FastAPI(title="OpenRAG API", version=OPENRAG_VERSION, debug=FASTAPI_DEBUG)
     app.state.services = services
     app.state.background_tasks = set()
+
+    if CORS_ALLOWED_ORIGINS:
+        app.add_middleware(
+            CORSMiddleware,
+            allow_origins=CORS_ALLOWED_ORIGINS,
+            allow_credentials=False,
+            allow_methods=["*"],
+            allow_headers=["*"],
+        )
 
     # Wire up ASGI request logging middleware (pure ASGI, not BaseHTTPMiddleware)
     app.add_middleware(RequestLoggingMiddleware)
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -110,7 +110,7 @@ DEFAULT_CHUNK_OVERLAP = 200
 OPENRAG_BACKEND_PORT = get_env_int("OPENRAG_BACKEND_PORT", 8000)
 
 # CORS – comma-separated list of allowed origins (e.g. "https://app.example.com,https://admin.example.com").
-# Leave empty or unset to forbid all cross-origin requests.
+# Unset → defaults to http://localhost:3000.  Set to "" to disable CORS entirely.
 _raw_cors_origins = os.getenv("CORS_ALLOWED_ORIGINS", "http://localhost:3000")
 CORS_ALLOWED_ORIGINS: list[str] = [o.strip() for o in _raw_cors_origins.split(",") if o.strip()]
 OPENRAG_BACKEND_INTERNAL_URL = os.getenv(
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -108,6 +108,13 @@ LANGFLOW_URL_INGEST_FLOW_ID = (
 DEFAULT_CHUNK_SIZE = 1000
 DEFAULT_CHUNK_OVERLAP = 200
 OPENRAG_BACKEND_PORT = get_env_int("OPENRAG_BACKEND_PORT", 8000)
+
+# CORS – comma-separated list of allowed origins (e.g. "https://app.example.com,https://admin.example.com").
+# Leave empty or unset to forbid all cross-origin requests.
+_raw_cors_origins = os.getenv("CORS_ALLOWED_ORIGINS", "")
+CORS_ALLOWED_ORIGINS: list[str] = [
+    o.strip() for o in _raw_cors_origins.split(",") if o.strip()
+]
 OPENRAG_BACKEND_INTERNAL_URL = os.getenv(
     "OPENRAG_BACKEND_INTERNAL_URL",
     f"http://openrag-backend:{OPENRAG_BACKEND_PORT}",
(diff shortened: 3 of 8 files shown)


mpawlow (commented):
## Functional Review 1

- ✅ Passed on **Ubuntu Linux** using IP address value for CORS

```
CORS_ALLOWED_ORIGINS=http://9.46.110.49:3000
```

### Manual Test Scenarios 

1. Golden path — normal chat still works (regression check)
   - Open http://9.46.110.49:3000/chat, send a message, confirm the streamed response renders normally.
   - Open DevTools → Console: confirm no CORS errors logged.
   - This proves removing the manual Access-Control-Allow-Origin: * headers in chat.py didn't break the app's actual (same-origin, proxied) path.

2. Allowed-origin direct request succeeds
   - With the app open at http://9.46.110.49:3000, open DevTools Console and run:
fetch("http://9.46.110.49:8000/health").then(r => console.log(r.status, r.headers.get("access-control-allow-origin")))
   - Expect 200 and access-control-allow-origin: http://9.46.110.49:3000. This is the simplest unauthenticated endpoint (/health, no session/API key needed) to confirm the allowlist grants the configured origin.

3. Disallowed-origin request is blocked (the actual vulnerability check)
   - In a separate terminal: cd /tmp && python3 -m http.server 5500, open http://9.46.110.49:5500 in the browser, open its Console and run the same fetch("http://9.46.110.49:8000/health").
   - Expect a CORS error in the console (No 'Access-Control-Allow-Origin' header...) and the response body inaccessible to JS. This is the scenario that was previously exploitable with the wildcard — it's the key proof the fix works, not just that nothing broke.

4. Preflight inspection (Network tab)
   - From either origin above, trigger a request with a custom header/method that forces a preflight (e.g. fetch(url, {method: "POST", headers: {"Content-Type": "application/json"}})).
   - In DevTools → Network, inspect the OPTIONS request: allowed origin gets Access-Control-Allow-Origin matching its exact origin; disallowed origin gets no such header (browser blocks before the actual request fires).

5. Streaming endpoint specifically (the code path that changed in chat.py)
   - During scenario 1's chat send, open Network tab, find the text/event-stream request to /api/chat (or /api/langflow), inspect Response Headers.
   - Confirm exactly one Access-Control-Allow-Origin (from CORSMiddleware now, not the old manually-set header — no duplicates), and that streaming still completes token-by-token without stalling.

src/config/settings.py
```diff
@@ -108,6 +108,11 @@ def get_opensearch_password() -> str | None:
 DEFAULT_CHUNK_SIZE = 1000
 DEFAULT_CHUNK_OVERLAP = 200
 OPENRAG_BACKEND_PORT = get_env_int("OPENRAG_BACKEND_PORT", 8000)
+
+# CORS – comma-separated list of allowed origins (e.g. "https://app.example.com,https://admin.example.com").
+# Unset → defaults to http://localhost:3000.  Set to "" to disable CORS entirely.
+_raw_cors_origins = os.getenv("CORS_ALLOWED_ORIGINS", "http://localhost:3000")
+CORS_ALLOWED_ORIGINS: list[str] = [o.strip() for o in _raw_cors_origins.split(",") if o.strip()]
```
mpawlow: ### (1a) [Normal] `CORS_ALLOWED_ORIGINS` is not wired into `docker-compose.yml` or the Helm chart — the new control is unconfigurable in the primary supported deployment paths

#### Problem

- `src/config/settings.py` reads `CORS_ALLOWED_ORIGINS` via `os.getenv(...)`, and `.env.example` documents it as a user-facing setting that `make setup` copies into `.env`.
- `docker-compose.yml`'s `openrag-backend` service does **not** use `env_file: .env` — it whitelists ~78 individual env vars in an explicit `environment:` block (lines 57–134), each written as `- VAR=${VAR:-default}`. `CORS_ALLOWED_ORIGINS` is absent from that list.
- Because of this, any value a user sets for `CORS_ALLOWED_ORIGINS` in `.env` never reaches the container under `make dev`, `make dev-cpu`, `make dev-local*`, or `make dev-branch*` — the backend process always falls back to the code-level default (`http://localhost:3000`), regardless of `.env`.
- The Kubernetes Helm chart has the same gap: `kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml` renders a `.env` file for the backend pod from `values.yaml` (see the existing pattern at line 107: `OPENRAG_AZURE_BLOB_ENABLED={{ .Values.global.azureBlob.enabled }}`), but there is no equivalent `CORS_ALLOWED_ORIGINS` entry or `values.yaml` field.
- Net effect: this security control can only actually be configured when the backend runs on the host directly (`make backend`, which sources `.env` via the process environment) — not via Docker or Kubernetes, which are the deployment paths this repo documents and ships images for.

#### Background Information

- This is exactly the failure mode the codebase's own convention exists to prevent — every other configurable backend env var (`OPENRAG_AZURE_BLOB_ENABLED`, `IBM_AUTH_ENABLED`, `DOCLING_SERVE_VERIFY_SSL`, etc.) is explicitly passed through in both `docker-compose.yml` and the Helm `backend-dotenv.yaml` template.
- Because the code default and the `.env.example` default happen to coincide (`http://localhost:3000`), local dev "works" and masks the gap — a production operator setting a real origin (e.g. `https://app.example.com`) in `.env` would silently get no CORS access for that origin under Docker/Helm, with no error or warning.

#### Code References

- `docker-compose.yml:57-134` (missing `- CORS_ALLOWED_ORIGINS=${CORS_ALLOWED_ORIGINS}` in the `openrag-backend` `environment:` block)
- `kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml` (no `CORS_ALLOWED_ORIGINS` line)
- `kubernetes/helm/openrag/values.yaml` (no corresponding value)
- `src/config/settings.py:112-115`

#### Potential Solution

- Add `- CORS_ALLOWED_ORIGINS=${CORS_ALLOWED_ORIGINS}` to the `openrag-backend` service's `environment:` block in `docker-compose.yml`.
- Add a corresponding `CORS_ALLOWED_ORIGINS={{ .Values.backend.cors.allowedOrigins | ... }}` line to `backend-dotenv.yaml` and a `backend.cors.allowedOrigins` field in `values.yaml`, following the existing `azureBlob.enabled` pattern.

src/config/settings.py
```diff
@@ -108,6 +108,11 @@ def get_opensearch_password() -> str | None:
 DEFAULT_CHUNK_SIZE = 1000
 DEFAULT_CHUNK_OVERLAP = 200
 OPENRAG_BACKEND_PORT = get_env_int("OPENRAG_BACKEND_PORT", 8000)
+
+# CORS – comma-separated list of allowed origins (e.g. "https://app.example.com,https://admin.example.com").
+# Unset → defaults to http://localhost:3000.  Set to "" to disable CORS entirely.
```
mpawlow, line 113: ### (1b) [Minor] Misleading comment: "unset" does not disable CORS, it defaults to `http://localhost:3000`

#### Problem

- `src/config/settings.py:113` states: *"Leave empty or unset to forbid all cross-origin requests."*
- The actual implementation: `os.getenv("CORS_ALLOWED_ORIGINS", "http://localhost:3000")` — only an **explicitly empty string** (`CORS_ALLOWED_ORIGINS=`) disables CORS; leaving the variable **unset** falls back to the default and CORS stays enabled for `http://localhost:3000`.
- `.env.example:68` reinforces the confusion by shipping the line uncommented with a non-empty value, so a reader following the comment's "leave unset" guidance by deleting the line entirely would not get the described "forbid all" behavior.

#### Code References

- `src/config/settings.py:112-115`
- `.env.example:64-68`

#### Potential Solution

- Reword the comment to reflect actual behavior, e.g.: *"Defaults to `http://localhost:3000` if unset. Set to an explicit empty string (`CORS_ALLOWED_ORIGINS=`) to forbid all cross-origin requests."*

src/config/settings.py
```diff
@@ -108,6 +108,11 @@ def get_opensearch_password() -> str | None:
 DEFAULT_CHUNK_SIZE = 1000
 DEFAULT_CHUNK_OVERLAP = 200
 OPENRAG_BACKEND_PORT = get_env_int("OPENRAG_BACKEND_PORT", 8000)
+
+# CORS – comma-separated list of allowed origins (e.g. "https://app.example.com,https://admin.example.com").
```
mpawlow, line 112: ### (1c) [Minor] No documentation update for the new env var / feature

#### Problem

- `docs/docs/reference/configuration.mdx` maintains a table of environment variables (model provider, document processing, Langflow settings, etc.) but was not updated with `CORS_ALLOWED_ORIGINS`.

#### Code References

- `docs/docs/reference/configuration.mdx`
- `src/config/settings.py:112-115`

#### Potential Solution

- Add a row/section documenting `CORS_ALLOWED_ORIGINS` (default, format, and the "empty disables" behavior once (1b) is resolved) to `docs/docs/reference/configuration.mdx`.

src/app/factory.py
```diff
@@ -26,6 +27,15 @@ async def create_app():
     app.state.services = services
     app.state.background_tasks = set()
 
+    if CORS_ALLOWED_ORIGINS:
```
mpawlow, line 30: ### (1d) [Minor] No test coverage for the new CORS behavior

#### Problem

- This PR changes a security control (replacing wildcard CORS with an allowlist) but adds no unit tests — neither for `CORS_ALLOWED_ORIGINS` parsing in `config/settings.py` (comma-splitting, whitespace-trimming, empty-entry filtering) nor for the conditional middleware registration in `app/factory.py`.
- A regression (e.g. accidentally reintroducing `allow_origins=["*"]`, or the middleware being skipped due to a truthiness bug) would not be caught by `make test-unit`.

#### Code References

- `src/config/settings.py:112-115`
- `src/app/factory.py:30-37`

#### Potential Solution

- Add a unit test asserting `CORS_ALLOWED_ORIGINS` parses `"a, b ,,c"` → `["a", "b", "c"]` and empty string → `[]`.
- Add an app-factory test (or extend an existing one) asserting `CORSMiddleware` is present/absent based on the parsed list, and that `allow_origins` never becomes `["*"]`.

.env.example
```diff
@@ -61,6 +61,12 @@ OPENRAG_BACKEND_PORT=8000
 # if you need to override the auto-derived value (e.g., custom hostname).
 # OPENRAG_BACKEND_INTERNAL_URL=http://openrag-backend:8000
 
+# --- CORS -------------------------------------------------------------------
+# Comma-separated list of origins allowed to make cross-origin requests
+# (e.g. "https://app.example.com,https://admin.example.com").
+# Unset → defaults to http://localhost:3000.  Set to "" to disable CORS entirely.
+CORS_ALLOWED_ORIGINS=http://localhost:3000
```
mpawlow, line 68: ### (1e) [Minor] Consolidate NEXT_ALLOWED_DEV_ORIGINS with CORS_ALLOWED_ORIGINS ?

#### Problem

- Slight redundancy between the two env vars: `CORS_ALLOWED_ORIGINS` and `NEXT_ALLOWED_DEV_ORIGINS`

```
# OPTIONAL: A development-mode only configuration value that allows additional origins (domains/hosts) to make requests to the Next.js dev server (next dev).
# - A list of comma-separate hostnames may be specified
# - e.g. 9.46.110.49,10.21.103.227,openrag.example.com
NEXT_ALLOWED_DEV_ORIGINS=9.46.110.49
```

#### Code References

- Line 326

#### Potential Solution

- Probably safest route is to not combine both variables together; but, worth investigating
- Simply grouping both env vars together in the `.env.example` file is probably more than sufficient

mpawlow (approved):
## Code Review 1

- See PR comment (1a) for consideration
- See Minor PR comments (1b) to (1e) for optional consideration

PR #2263, opened 2026-08-19
feat: Litellm OpenAI compatible router

edwinjosechittilappilly:
## Summary

- OpenRAG now proxies chat and embeddings through OpenAI-compatible `/v1` (`/models`, `/model-catalog`, `/chat/completions`, `/embeddings`). LiteLLM stays on the backend; Langflow and other clients never hold upstream keys. Auth is a short-lived Langflow hop token (`OPENRAG_LLM_TOKEN`), a user JWT, or an `orag_` API key.
- Classic four providers (`openai`, `anthropic`, `watsonx`, `ollama`) stay first-class. Everything else is a generic LiteLLM provider: catalog-driven credential fields, encrypted persistence, health checks, model listing, and gateway routing by model prefix. Secrets are never returned on GET.
- Settings and onboarding use one catalog picker with provider logos (real mark or monogram), provider-aware model matching when IDs collide, credential reuse, and a full-catalogue card grid. Saving a generic provider runs a real LiteLLM call and surfaces the vendor error in the dialog instead of failing later in chat/ingest.
- Bundled flows switch to **OpenRAG LLM** / **OpenRAG Embeddings**, bound to `OPENRAG_LLM_BASE_URL`, `OPENRAG_LLM_TOKEN`, and `SELECTED_*_MODEL`. Langflow sidebar gets an OpenRAG bundle; Docker/Helm seed a non-empty `OPENRAG_LLM_TOKEN=None` placeholder so Credential globals resolve.
- Onboarding ingest no longer dies under Langflow 1.11 restricted mode: OpenSearch (Multi-Model Multi-Embedding) is registered in `flows/component_index.json`, referenced as `ext:openrag:…@extra` so the Elastic rewriter cannot steal the node, index SHA-256 matches Langflow’s UTF-8 digest, and ingest/URL node IDs used for sample-doc tweaks match the re-exported graphs.

## Test plan

- [ ] Configure OpenAI (or another classic provider), pick chat + embedding models, complete onboarding, and confirm a document ingest plus a chat turn succeed.
- [ ] Add a generic LiteLLM provider from the catalogue cards (e.g. Azure/Groq). Confirm the save-time validation call fails inline on a bad endpoint/key, then succeeds with good credentials, and that the provider can be used for chat and ingest.
- [ ] Confirm Settings → Providers still shows the four built-in cards first, then configured custom providers, then the rest of the catalogue; logos match between built-in and catalogue tiles.
- [ ] Hit `GET /v1/models` and `GET /v1/model-catalog` with a JWT/`orag_` key; confirm provider secrets are absent from settings GET.
- [ ] Restore bundled flows in Langflow Settings (startup is create-only). Confirm OpenRAG LLM / Embeddings / OpenSearch nodes are present and ingest is not blocked with “custom components are not allowed”.
- [ ] Local compose: Langflow should load `LANGFLOW_COMPONENTS_INDEX_PATH=/app/flows/component_index.json` (mounted). Helm/image deploys still need a Langflow image rebuild so the baked index includes OpenSearch.
- [ ] Provider health banner on Settings should report a real check, not `cache_key() got an unexpected keyword argument 'credentials'`.
Screenshot 2026-08-18 at 7 41 58 PM
Screenshot 2026-08-18 at 7 41 33 PM

## Summary by CodeRabbit

* **New Features**
  * Added a unified model catalog with provider grouping, capabilities, pricing, context limits, and deprecation details.
  * Added custom provider configuration with dynamic credentials and secret reuse.
  * Added OpenAI-compatible chat and embedding access through the OpenRAG proxy, including streaming.
  * Redesigned onboarding and settings with shared model selection, provider search, credential management, and verification.
  * Updated ingestion flows to support configurable OpenRAG-compatible embedding models.

* **Bug Fixes**
  * Improved provider validation, credential caching, model fallback selection, and secure credential handling.


mjobrien05:
Hi — I've been running a temporary local stopgap for this exact need (routing OpenRAG's `openai` provider through LiteLLM to self-hosted NVIDIA NIM models) ahead of this PR landing, and hit a finding along the way that's worth flagging regardless of the final routing/credential approach.

**NVIDIA's asymmetric embedding NIMs (e.g. `llama-3.2-nv-embedqa-1b-v2`) require an `input_type` field (`"passage"` for indexed content, `"query"` for search queries) that isn't part of the standard OpenAI embeddings API.** None of the `embeddings.create()` call sites in the codebase pass it, so any embedding call to one of these models fails with:

```
'input_type' parameter is required for asymmetric models
```

I hit this at all three real call sites:
- `src/models/processors.py` — the actual document-chunk ingestion embedding call
- `src/services/langflow_file_service.py` — the dimension-detection probe
- `src/services/search_service.py` — query-time embedding at retrieval

This is independent of how the provider/credentials get configured — it'll affect this PR's routing too once someone points it at an NVIDIA NIM embedding endpoint, not just my stopgap. My fix was a small conditional `extra_body={"input_type": ...}` at each site (only when a non-default endpoint is configured, so standard OpenAI usage is unaffected) — happy to share the exact diff if useful, or this may already be a natural fit for wherever this PR ends up modeling provider-specific request shaping.

Separately, and less pressing: I noticed `config.yaml` isn't the actual source of truth once `WorkspaceConfigService`'s "db" storage mode is active — settings written outside the app's own save path can get silently shadowed by the DB-mirrored config. Not related to this PR, just a trap worth knowing about when testing provider changes by hand.

PR #2265, opened 2026-08-19
feat: Instrument Backend app via IBM Instana for APM and observability

### LINKED ISSUE #2264: [Feature]: Instrument Backend application via IBM Instana for APM and full-stack observability
### Feature Area

Backend/API (Python/FastAPI)

### Problem Description

- Need a tool that automatically provides application performance monitoring (APM) and full-stack observability.

### Proposed Solution

- Instrument Backend application via IBM Instana for APM and full-stack observability
- Primary Deployment Target: Private Cloud (CPD)
- Secondary Deployment Target: Public Cloud (SaaS)

### Use Case

- From an engineering, SRE, and DevOps teams perspective, I need a tool that helps track, analyze, and troubleshoot complex cloud-native applications, microservices, and container environments in real time.

### Alternatives Considered

- Prometheus

### Priority

Would improve my workflow

### Additional Context

_No response_

### Contribution

### Checklist

mpawlow:
### ℹ️ Issue

- https://github.com/langflow-ai/openrag/issues/2264

### ℹ️ Summary

- Added opt-in `IBM Instana` auto-instrumentation to the backend, giving engineering, SRE, and DevOps teams `APM` and full-stack observability.
- Wired the feature through every deployment path: local `uv`/Docker dev, the Docker Compose stack, the Helm chart, and the Kubernetes operator — disabled by default in all of them.

---

### ✅ Backend Instrumentation

- Added `src/observability/instana_boot.py`, gating tracer activation behind `INSTANA_ENABLED` (accepts `true`/`1`/`yes`).
- Called `boot_instana()` from `bootstrap.load_env()` immediately after `.env` load and structlog configuration — the earliest point where the flag was readable, and before `httpx`, `urllib3`, `sqlalchemy`, `starlette`, `fastapi`, and `logging` were imported elsewhere, since `instana` monkey-patches those at import time.
- Documented `bootstrap.py` as the sanctioned exception to "`config/settings.py` is the only module that reads `os.environ` directly," since importing `config.settings` would pull in the very libraries the tracer needs to patch first.
- Downgraded a missing `instana` package to a warning rather than a startup failure, since the tracer ships as an optional extra.
- Added `instana` as an optional dependency (`apm` extra, `pyproject.toml`, `uv.lock`) rather than a hard dependency, so OSS installs that never enable tracing don't pay for it.
- Updated `Dockerfile.backend` to `uv sync --extra apm`, so published images can turn on `APM` via `INSTANA_ENABLED` alone.

### ✅ Local Dev Tooling

- Added `make instana-agent-up` / `make instana-agent-down` targets (plus `help_local` text) to run a local `Instana` host agent via a new `instana` Docker Compose profile.
- Added the `instana-agent` service to `docker-compose.yml`: pinned image tag, host networking, and forwarding to the `Instana` SaaS ingress named by `INSTANA_AGENT_ENDPOINT`.
- Flagged the agent container as local-dev-only in Make output and compose comments, since it runs privileged, in the host PID namespace, with `Docker` socket access.
- Forwarded `INSTANA_ENABLED`, `INSTANA_AGENT_HOST`/`_PORT`, `INSTANA_TRACING_DISABLE`, `INSTANA_STACK_TRACE`, `INSTANA_SERVICE_NAME`, `INSTANA_LOG_LEVEL`, and `INSTANA_ZONE` into the backend container, with `INSTANA_TRACING_DISABLE` defaulted to `logging` and `INSTANA_STACK_TRACE` defaulted to `error` to avoid unbounded span/event growth and full stack captures on every exit span.

### ✅ Kubernetes (Helm + Operator)

- Added a `backend.instana` value block to `kubernetes/helm/openrag/values.yaml` (`enabled`, `agentHost`, `agentPort`, `serviceName`, `logLevel`, `zone`), defaulting to disabled.
- Rendered `INSTANA_ENABLED` / `INSTANA_AGENT_PORT` into the backend `.env` `Secret`, emitting `INSTANA_SERVICE_NAME`, `INSTANA_LOG_LEVEL`, and `INSTANA_ZONE` only when set, to avoid a blank service name or an "Unknown `INSTANA_LOG_LEVEL`" warning at boot.
- Injected `INSTANA_AGENT_HOST` into the backend `Deployment` via the Downward API (`status.hostIP`), since the agent address is the pod's own node IP and can't be rendered into the mounted `.env` file.
- Added `INSTANA_ENABLED` (`false`) and `INSTANA_AGENT_PORT` (`42699`) defaults to `DefaultOpenRagBEEnvVars`, plus `InstanaAgentHostEnvVar` to build the Downward API env var.
- Added `lookupBackendEnvValue` to resolve the flag across the three precedence levels for literal values, and `isTruthyEnvValue` to mirror the backend's truthy-value parsing so the operator and the application agree on "enabled."
- Suppressed the Downward-API injection whenever an explicit agent host was configured at any precedence level.
- Added `kubernetes/operator/internal/controller/instana_test.go`, covering default-off behavior, truthiness parsing, `status.hostIP` injection, explicit-host suppression, operator-prefix precedence, `valueFrom` fallback, and the guard against empty defaults.
- Corrected `ENV_OVERRIDE.md`'s `valueFrom` section, which had claimed CR `valueFrom` vars pass through to the pod spec; they're resolved during reconcile, and `fieldRef`/`resourceFieldRef` are rejected.

### ✅ Configuration & Documentation

- Documented the full `INSTANA_*` environment variable set in `.env.example` (enable flag, agent host/port, service/process name, log level, tracing/stack-trace guardrails, agent key/endpoint/port).
- Added `docs/docs/reference/observability.mdx` (deployment topologies, tenant endpoints, security considerations, startup requirements, configuration options) and linked it from `docs/sidebars.js`.
- Extended `docs/docs/reference/configuration.mdx` with the `Instana` env var reference.

### ✅ Test Coverage

- Added `tests/unit/observability/test_instana_boot.py` for the enable-flag parsing and tracer-load gate.
- Updated `tests/conftest.py` to force `INSTANA_ENABLED=false` for the test process unless a developer set it explicitly on the shell, avoiding an instrumented test run (and the tracer's per-warning event-leak behavior in `instana` 3.17.0) whenever `INSTANA_ENABLED=true` was left in `.env`.

### ✅ Unrelated Fix Bundled In

- Fixed `ensure_required_langflow_global_variables()` (`src/api/settings/langflow_sync.py`) to stop pushing empty values for unconfigured optional providers (`OLLAMA_BASE_URL`, `WATSONX_PROJECT_ID`, `WATSONX_URL`) to Langflow, which previously triggered a `400 "Variable value cannot be empty"` and, during the type-migration path, could drop the variable entirely after its `DELETE` had already landed.
- Extended `tests/unit/test_langflow_global_variables.py` to cover the new empty-value handling.

## New Features
- Added opt-in IBM Instana APM and distributed tracing for backend requests, external calls, databases, and logs.
- Added configurable agent connectivity, service metadata, logging, zones, credentials, endpoints, and image tags.
- Added a local Instana agent for Docker with start and stop commands.
- Added Kubernetes and operator support with automatic node-based agent host detection.

## Bug Fixes
- Prevented empty optional provider values from overwriting existing Langflow settings.

## Documentation
- Added Instana setup, deployment, security, and configuration guidance.


.env.example
```diff
@@ -420,6 +420,37 @@ SERVICE_NAME=openrag
...
+#
+# OPTIONAL: Instana tracer's own log verbosity
+# - Default: info
+# INSTANA_LOG_LEVEL=info
+#
+# Instana host agent key
+# INSTANA_AGENT_KEY=
+# OPTIONAL: Instana host agent endpoint
+# - Default: ingress-red-saas.instana.io
+# INSTANA_AGENT_ENDPOINT=ingress-coral-saas.instana.io
```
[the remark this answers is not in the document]
mpawlow: Agree 👍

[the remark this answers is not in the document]
mpawlow: Fixed.

src/main.py
```diff
@@ -20,6 +20,15 @@
...
 
+import os
+
+# Must run before config.settings (and everything it pulls in — httpx,
+# opensearch-py, sqlalchemy) is imported, so Instana can patch those libraries.
+# Exception to "config/settings.py is the only place that reads os.environ":
+# config.settings can't be imported yet at this point in the file.
+if os.getenv("INSTANA_ENABLED", "false").lower() in ("true", "1", "yes"):
+    import instana  # noqa: F401  — must precede httpx/opensearch-py/sqlalchemy imports below
+
```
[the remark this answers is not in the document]
mpawlow: Fixed.

kubernetes/operator/internal/controller/env.go
```diff
@@ -315,6 +332,68 @@ func resolveEnvVarValue(ctx context.Context, c client.Client, namespace string,
...
+func (m *EnvVarManager) lookupBackendEnvValue(crEnvVars []corev1.EnvVar, key string) string {
+	value := m.DefaultOpenRagBEEnvVars[key]
+	if v, ok := os.LookupEnv(OPENRAGBE_ENV_PREFIX + key); ok {
+		value = v
+	}
+	for _, envVar := range crEnvVars {
+		if envVar.Name == key && envVar.ValueFrom == nil {
+			value = envVar.Value
+		}
+	}
```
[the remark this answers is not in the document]
mpawlow: Fixed.

docs/docs/reference/observability.mdx
```diff
@@ -0,0 +1,227 @@
...
+   ```bash
+   INSTANA_ENABLED=true
+   INSTANA_AGENT_KEY=YOUR_AGENT_KEY
+   INSTANA_AGENT_ENDPOINT=ingress-coral-saas.instana.io
+   INSTANA_AGENT_ENDPOINT_PORT=443
+   INSTANA_SERVICE_NAME="OpenRAG Backend"
+   INSTANA_ZONE="openrag-dev"
+   ```
+
+   `INSTANA_SERVICE_NAME` and `INSTANA_ZONE` are optional, but they make OpenRAG easier to find in the Instana UI.
```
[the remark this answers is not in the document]
mpawlow: Fixed.

src/observability/instana_boot.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def is_instana_enabled(value: str | None = None) -> bool:
+    """Return True when ``INSTANA_ENABLED`` asks for tracing.
+
+    Reads the process environment when ``value`` is not supplied, which is what
+    the boot path does; passing it explicitly keeps the parsing testable.
+    """
+    if value is None:
+        value = os.environ.get(INSTANA_ENABLED_ENV_VAR, "")
```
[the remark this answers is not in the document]
mpawlow: Invalid. Current behavior is by design and a documented exception.

mpawlow:
@coderabbitai summary

mpawlow:
@coderabbitai full review

kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
```diff
@@ -247,4 +247,29 @@ stringData:
...
+    INSTANA_SERVICE_NAME={{ .Values.backend.instana.serviceName | quote }}
+    {{- end }}
+    {{- if .Values.backend.instana.logLevel }}
+    INSTANA_LOG_LEVEL={{ .Values.backend.instana.logLevel | quote }}
+    {{- end }}
+    {{- if .Values.backend.instana.zone }}
+    INSTANA_ZONE={{ .Values.backend.instana.zone | quote }}
+    {{- end }}
+    {{- if .Values.backend.instana.tracingDisable }}
+    INSTANA_TRACING_DISABLE={{ .Values.backend.instana.tracingDisable | quote }}
```
rodageve, line 269: From Bob IDE analysis regarding the defaults behaviour - (operator must set values to "" in order to set to default behaviour)

The backend-dotenv.yaml block writes INSTANA_AGENT_PORT unconditionally inside the {{- if .Values.backend.instana.enabled }} block, but INSTANA_TRACING_DISABLE and INSTANA_STACK_TRACE are guarded by {{- if .Values.backend.instana.tracingDisable }} / {{- if .Values.backend.instana.stackTrace }}. Since both those values default to non-empty strings ("logging" and "error"), the guards will always pass when enabled: true — so in practice the defaults are always written. The guard pattern is consistent with the other conditional vars; it just means an operator who wants to revert to tracer-default behaviour for those two specifically must set them to "" in their override.

rodageve (approved):

docker-compose.yml
```diff
@@ -125,6 +125,35 @@ services:
...
+      # level because the tracer rejects an empty string, logging "Invalid
+      # stack-trace value" on every boot and silently reverting to `all` — the
+      # expensive setting this exists to avoid. Opt out with `all`, not ``.
+      - INSTANA_TRACING_DISABLE=${INSTANA_TRACING_DISABLE-logging}
+      - INSTANA_STACK_TRACE=${INSTANA_STACK_TRACE:-error}
+      # Listed bare rather than as `VAR=${VAR}` on purpose: the tracer tests these
+      # for *presence*, not truthiness, so forwarding an empty string would set a
+      # blank service name and log an "Unknown INSTANA_LOG_LEVEL" warning on every
+      # boot. The bare form is omitted from the container entirely when unset.
+      - INSTANA_SERVICE_NAME
```
mfortman11, line 166: Forward `INSTANA_SECRETS` to the backend container

`.env.example` recommends setting `INSTANA_SECRETS` to redact OpenRAG’s free-text query parameters, including `q`, `search`, and `filename`. However, this environment list does not forward that variable into `openrag-backend`.

An operator can therefore configure the documented redaction policy and believe it is active while the tracer continues using Instana’s credential-only default. Requests such as `GET /v2/files/search?q=<usertext>` may export user-supplied search text in trace data. Because this silently defeats an explicit privacy control, I consider this blocking.

Please add the variable alongside the other optional Instana settings:

  - `INSTANA_SECRETS`

The bare form preserves Instana’s default when the variable is unset and forwards the configured value unchanged when present.

kubernetes/helm/openrag/values.yaml
```diff
@@ -381,6 +381,42 @@ backend:
...
+    #
+    # tracingDisable: span categories/types not to trace. `logging` avoids an
+    # unbounded event-list leak (and O(n^2) CPU) on every WARNING/ERROR logged
+    # inside a request trace, at no loss of information — the same events are
+    # already in OpenRAG's structured logs.
+    tracingDisable: "logging"
+    # stackTrace: when to capture a Python stack for EXIT spans (all|error|none).
+    # `all` means a full traceback on every httpx / urllib3 / sqlalchemy call.
+    # Anything other than all|error|none logs a warning and falls back to `all`.
+    stackTrace: "error"
```
mfortman11, line 420: **Expose query-parameter redaction in Helm deployments**

The chart exposes the other supported Instana settings, but it provides neither a value for `INSTANA_SECRETS` nor a generic backend environment-variable passthrough. Helm users therefore have no way to apply the query-redaction policy documented in `.env.example` and the configuration reference.

With Instana enabled, the tracer consequently retains its credential-only default. Free-text parameters such as `q`, `search`, and `filename` may be included in exported traces even when an operator intends to redact them. This is a blocking privacy concern for Helm deployments because the documented control cannot currently be configured at all.

  Please add a chart value, for example:

```
  backend:
    instana:
      secrets: ""
```

Then render it conditionally in `templates/backend/backend-dotenv.yaml`:
```
  {{- if .Values.backend.instana.secrets }}
  INSTANA_SECRETS={{ .Values.backend.instana.secrets | quote }}
  {{- end }}
```
Please also document the value in the Helm example and add a render test confirming that it appears in the backend dotenv when configured and remains absent when unset.

tests/conftest.py
```diff
@@ -14,9 +14,27 @@
...
+# explicitly on the command line (`INSTANA_ENABLED=true uv run pytest`).
+#
+# `from main import ...` below pulls in `bootstrap`, which boots the tracer at
+# import time, so a developer with INSTANA_ENABLED=true in .env would otherwise
+# run the entire suite instrumented. That costs ~1.1s of extra import time per
+# process and monkey-patches logging, asyncio.create_task, httpx, urllib3 and
+# sqlalchemy underneath every test — and in instana 3.17.0 each WARNING/ERROR
+# logged inside a trace leaks an event into a process-global list that every
+# later log span re-walks. None of that belongs in a test run.
+if _INSTANA_FROM_SHELL is None:
```
mfortman11: **Lower Priority: Do not treat Make-exported `.env` values as an explicit test opt-in**

This check cannot reliably distinguish a value supplied explicitly to pytest from one loaded by the Makefile. The Makefile includes .env and exports all of its variables before starting pytest, so with
`INSTANA_ENABLED=true` in `.env`, `_INSTANA_FROM_SHELL` is already non-None under the standard make test and make test-unit workflows.

Those test runs therefore remain instrumented despite this isolation guard. Depending on the installed extras and agent configuration, that can monkey-patch shared libraries, trigger the documented Instana event leak, slow or destabilize the suite, and potentially send test traces externally. This is a P2 because it affects test isolation rather than production behavior, but it undermines the guard’s stated purpose on the repository’s standard test path.

Please make test instrumentation an unambiguous, test-specific opt-in. For example, conftest.py could always disable Instana unless a dedicated flag is set:

```
  if os.environ.get("OPENRAG_TEST_INSTANA", "").lower() not in {"true", "1", "yes"}:
      os.environ["INSTANA_ENABLED"] = "false"
```
A developer who intentionally wants an instrumented test run could then use:
```
  OPENRAG_TEST_INSTANA=true INSTANA_ENABLED=true uv run pytest
```
Alternatively, the Make test targets could explicitly invoke `pytest` with `INSTANA_ENABLED=false`, but a dedicated opt-in flag makes the isolation rule independent of how `pytest` is launched. Please add coverage for both `.env-derived` values and an explicit test opt-in.

mfortman11 (changes requested):

.env.example
```diff
@@ -426,6 +426,113 @@ SERVICE_NAME=openrag
...
+#
+# OPTIONAL: Query-string parameters the tracer redacts before spans leave the
+# host. Format: <matcher>:<name>[,<name>], matcher one of equals,
+# equals-ignore-case, contains, contains-ignore-case, regex.
+# - Default: contains-ignore-case:key,pass,secret
+# The default covers credential-shaped names but NOT free-text values, so
+# GET /v2/files/search?q=<user query> is exported verbatim. The value below
+# keeps the default credential coverage and adds OpenRAG's search parameters.
+# Spans never carry request bodies, so chat prompts and document text stay local.
+# INSTANA_SECRETS=regex:.*key.*,.*pass.*,.*secret.*,.*token.*,^q$,^search$,^filename$
```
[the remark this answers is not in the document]
mpawlow: Addressed locally. Commit push pending to upstream branch...

docs/docs/reference/observability.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+OpenRAG Backend can be instrumented with [IBM Instana](https://www.ibm.com/products/instana) for application performance monitoring (APM) and distributed tracing.
+
+Instrumentation is off by default. When you set `INSTANA_ENABLED=true`, the backend loads Instana's Python tracer at startup, before the libraries the tracer instruments, and traces are collected automatically for:
+
+* Incoming API requests to the backend (FastAPI and Starlette).
+* Outgoing HTTP calls to Langflow, Docling, and model providers.
+* OpenSearch queries.
+* Database queries issued through SQLAlchemy, for both the default SQLite database and PostgreSQL.
+* Backend log records, which the tracer attaches to the trace that was active when they were emitted.
```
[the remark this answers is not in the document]
mpawlow: Addressed locally. Commit push pending to upstream branch...

src/api/settings/langflow_sync.py
```diff
@@ -158,6 +158,16 @@ async def ensure_required_langflow_global_variables(config=None):
...
+            # Langflow rejects an empty value with 400 "Variable value cannot be
+            # empty". Optional providers (Ollama, watsonx) resolve to "" when
+            # unconfigured, so keep whatever Langflow already holds instead of
+            # pushing a value it will refuse. This matters most in the type
+            # migration below: the DELETE has already landed by then, so a
+            # refused recreate would drop the variable entirely.
+            sync_value = is_generic and bool(target_val)
+            if not sync_value:
+                target_val = curr_val
+
```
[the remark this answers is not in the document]
mpawlow: Addressed locally. Commit push pending to upstream branch...

kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
mpawlow: - The Helm guard pattern for tracingDisable/stackTrace (non-empty defaults that always render unless explicitly overridden to ""/"all") is intentional
- Missing a Kubernetes documentation section in `docs/docs/reference/observability.mdx` 
- Documented only the opposite rule (presence-based omission for serviceName/zone/logLevel) and never called out that tracingDisable/stackTrace behave differently.

docker-compose.yml
mpawlow: Valid concern.
Resolved upstream in a commit addressing the following code review comment: https://github.com/langflow-ai/openrag/pull/2265#discussion_r3853556310

.env.example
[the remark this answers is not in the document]
mpawlow: Commit pushed upstream.
Fixed.

docs/docs/reference/observability.mdx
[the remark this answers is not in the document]
mpawlow: Commit pushed upstream.
Fixed.

src/api/settings/langflow_sync.py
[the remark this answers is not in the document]
mpawlow: Commit pushed upstream.
Fixed.

kubernetes/helm/openrag/templates/backend/backend-dotenv.yaml
mpawlow: Fixed.

docker-compose.yml
mpawlow: Fixed.

kubernetes/helm/openrag/values.yaml
mpawlow: Fixed by commit 34881f62
- Updated doc example to include the missing secrets field accordingly: `docs/docs/reference/observability.mdx`

tests/conftest.py
mpawlow: Fixed.

- Added `is_instana_test_opt_in()` / TEST_OPT_IN_ENV_VAR = "OPENRAG_TEST_INSTANA", a dedicated test-only flag decoupled from `INSTANA_ENABLED`
- Added 4 new tests

.env.example
```diff
@@ -426,6 +426,122 @@ SERVICE_NAME=openrag
...
+# WHY: the tracer default only catches credential-shaped names. It does NOT
+# redact free-text values, so GET /v2/files/search?q=<user query> is exported
+# to Instana verbatim, along with the `search` and `filename` parameters used
+# by file listing. Spans never carry request or response bodies, so chat
+# prompts and document text stay local regardless of this setting.
+#
+# Set this to an empty value (not commented out) to fall back to the tracer's
+# own weaker default; it logs a "Couldn't parse INSTANA_SECRETS" warning each
+# time you do.
+INSTANA_SECRETS=regex:.*key.*,.*pass.*,.*secret.*,.*token.*,^q$,^search$,^filename$
```
mfortman11: When `.env.example` is copied to `.env` and the documented `make backend` or Compose targets are used, the Makefile directly includes and exports this assignment, so GNU Make interprets each `$` as a variable reference. The value becomes `regex:.*key.*,.*pass.*,.*secret.*,.*token.*,^q^search^filename`, causing the promised `q`, `search`, and `filename` redaction to fail and exporting user search text to Instana. Represent the anchors in a Make-safe form or avoid parsing this dotenv value through Make.

mfortman11 (approved):

.env.example
mpawlow: Valid, reproducible, and worth fixing.

mpawlow: Fixed.

PR #2266, opened 2026-08-19
feat: Add http metrics to prometheus

mfortman11:
## Summary by CodeRabbit

* **New Features**
  * Added HTTP request monitoring with duration and request-count metrics.
  * Metrics now include request method, normalized route, and response status.
  * Added route normalization for framework assets, known application routes, and unmatched paths.
* **Tests**
  * Expanded coverage for route normalization, including framework paths, root routes, settings, health, API, chat, and favicon requests.


frontend/proxy.ts
```diff
@@ -0,0 +1,37 @@
...
+  // `backend_proxy_*`.
+  const durationSeconds = (performance.now() - start) / 1000;
+  const labels = {
+    method: request.method,
+    route: normalizeRoute(request.nextUrl.pathname),
+    status_code: response.status.toString(),
+  };
+
+  httpRequestDuration.observe(labels, durationSeconds);
+  httpRequestsTotal.inc(labels);
```
mpawlow: ### (1a) [Major] Inbound HTTP metrics always record status `200` and near-zero duration

- 👍  Concur with CodeRabbit AI 

#### Problem

- `proxy()` calls `NextResponse.next()`, which returns synchronously **before** the matched route handler runs.
- `response.status` at that point is always the `NextResponse.next()` default (`200`) — it is never the actual status the client ultimately receives.
- `durationSeconds` is measured between two synchronous statements (`performance.now()` → `NextResponse.next()`), so it captures only the proxy hop, not route execution — this reliably yields sub-millisecond durations regardless of real request latency.
- The result: `http_request_duration_seconds` and `http_requests_total` are populated with data that does not reflect real HTTP status or real latency for **any** request, defeating the stated purpose of the PR ("Add http metrics to prometheus" for monitoring/alerting).
- The code's own comment (`proxy.ts:17-20`) acknowledges this limitation but ships the metric anyway rather than fixing the measurement point.

#### Background Information

- The existing `backend_proxy_*` metrics (`frontend/app/api/[...path]/route.ts:125-144`) already demonstrate the correct pattern in this codebase: measure `performance.now()` before `await fetch(...)`, then observe duration/status **after** the awaited response returns.
- Next.js 16's Proxy convention (formerly `middleware.ts`) runs as a separate execution layer ahead of route handlers and explicitly does not have access to the final response — this is a known platform constraint (confirmed by CodeRabbit's own research citing Next.js 16 docs), not something fixable by tweaking this function in place.
- For the same physical request to `/api/*`, this makes `http_requests_total{status_code="200"}` and `backend_proxy_requests_total{status_code="500"}` disagree — an operator using the new metric to gauge overall site health would see a misleadingly healthy signal even during backend outages.

#### Code References

- `frontend/proxy.ts:12-31`
- `frontend/lib/metrics.ts:50-62` (metric definitions whose docstrings claim accuracy the implementation can't deliver)
- Compare with correct pattern: `frontend/app/api/[...path]/route.ts:125-144`

#### Potential Solution

- Move inbound HTTP metrics recording out of `proxy.ts` to a point where the final response is known, e.g.:
  - Wrap `app/api/[...path]/route.ts`'s `proxyRequest` (and any other route handlers that should be measured) directly, following the existing `backend_proxy_*` pattern — observe after `await fetch(...)`/render completes.
  - Or, if inbound-edge instrumentation for **all** routes (not just `/api/*`) is required, use a `res.on("finish")`-style hook via a custom server, or Next's `after()` API / response streaming tap that fires once the response is fully sent — not the proxy hop.
- If per-request measurement of arbitrary App Router pages truly isn't feasible in Next 16's Proxy model, scope this metric down to what `proxy.ts` can honestly measure (e.g. rename to reflect "proxy dispatch overhead" rather than "inbound HTTP duration/status"), or drop the `status_code` label entirely from proxy-level metrics since it's always fabricated.

#### Alternative Solutions

- Instrument the small number of App Router route handlers explicitly (most idiomatic, matches existing `backend_proxy_*` precedent) rather than trying to force whole-app inbound coverage through Proxy.
- If broad edge visibility is the real goal, get it from the reverse proxy/ingress layer (e.g., existing infra metrics) instead of application code, since Proxy is architecturally unable to observe final status.

frontend/proxy.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  return response;
+}
+
+export const config = {
+  // Every inbound path except the dev-only HMR socket.
+  matcher: ["/((?!_next/webpack-hmr).*)"],
```
mpawlow: ### (1b) [Normal] Unbounded `route` label cardinality from unsanitized inbound paths

#### Problem

- `proxy.ts`'s matcher (`/((?!_next/webpack-hmr).*)`) applies to **every** inbound request to the site, including paths never routed by the app (404s, bot/scanner probes like `/wp-login.php`, `/.env`, `/phpmyinfo.php`, etc.).
- `normalizeRoute` → `normalizePath` only collapses UUID, 24-hex, and all-numeric path segments to `:id`; any other arbitrary segment passes through unchanged as a distinct label value.
- Prometheus label cardinality is a well-known operational hazard: since inbound requests are attacker/bot-influenced (unlike the existing `backend_proxy_*` metrics, which only see known application-issued `/api/*` calls), a public deployment is exposed to unbounded `route` label growth over the life of the process, inflating registry memory and Prometheus TSDB series count.

#### Code References

- `frontend/lib/metrics.ts:77-101` (`normalizePath` / `normalizeRoute`)

#### Potential Solution

- Cap cardinality by bucketing unmatched/unknown paths into a single `other` (or `404`) label value instead of passing raw segments through, e.g. only allow-list known top-level route prefixes and normalize everything else.

#### Alternative Solutions

- Narrow the proxy `matcher` to only the known first-class route prefixes (`/api`, `/chat`, `/settings`, …) plus a catch-all bucketed as `other`, rather than matching literally everything.

frontend/lib/metrics.test.ts
```diff
@@ -0,0 +1,40 @@
+import assert from "node:assert/strict";
```
mpawlow, line 1: ### (1c) [Minor] New unit test has no wired-up runner (mirrors a pre-existing gap)

#### Problem

- `frontend/lib/metrics.test.ts` uses Node's built-in `node:test` / `node:assert` APIs, but there is no `npm test` script, no Make target, and no CI workflow (`lint-frontend.yml`, `test-ci.yml`) that executes `node --test` anywhere in the repo.
- The test therefore never runs automatically and provides no real regression protection, despite reading as verified coverage.

#### Background Information

- This is not a new pattern introduced by this PR — `frontend/lib/chat-stream-errors.test.ts` has the identical problem already. This PR just adds a second orphaned file to the same gap rather than fixing it.

#### Code References

- `frontend/package.json` (no `test` script)
- `frontend/lib/chat-stream-errors.test.ts` (pre-existing instance of the same gap)

#### Potential Solution

- Out of scope to fix fully in this PR, but worth a follow-up: add a `"test": "node --test lib/**/*.test.ts"` (or similar) script to `frontend/package.json` and wire it into `lint-frontend.yml` or a new CI job, covering both this file and `chat-stream-errors.test.ts`.

mpawlow (changes requested):
## Code Review 1

- See Major PR comment (1a) 

rodageve (approved):
Review Code Rabbit suggestions and @mpawlow 's

frontend/lib/metrics.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { describe, it } from "node:test";
+import { normalizePath, normalizeRoute } from "./metrics";
+
+describe("normalizeRoute", () => {
+  it("collapses per-chunk static asset paths", () => {
```
mpawlow: ### (2a) [Minor] `metrics.test.ts` assertions are stale against the reworked `normalizeRoute` — 4/4 tests fail

#### Problem

- `normalizeRoute` was reworked (`frontend/lib/metrics.ts:108-116`) to collapse **all** `/_next/*` paths to `/_next/*`, and to bucket any path whose first segment isn't in `KNOWN_PREFIXES` to `"unmatched"`.
- `frontend/lib/metrics.test.ts` was not updated to match and still asserts the old contract (`/_next/static/*`, `/_next/data/*`, `/_next/image` preserved distinctly, `/favicon.ico` passed through unchanged).
- The suite was ran directly (`node --test` via `tsx`, after installing the missing `prom-client` dep locally) and confirmed **4 of 4** assertions in the `"collapses..."`/`"leaves..."`/`"delegates..."` tests fail against current `normalizeRoute` output (e.g. expects `/_next/static/*`, gets `/_next/*`; expects `/favicon.ico`, gets `unmatched`).
- This is currently masked only because — per review #1 finding (1c) — nothing in CI or `package.json` actually executes `node:test` files, so the breakage is silent today.

#### Code References

- `frontend/lib/metrics.test.ts:6-39` (stale assertions)
- `frontend/lib/metrics.ts:108-116` (current `normalizeRoute`)

#### Potential Solution

- Update `metrics.test.ts` to assert the current contract: `/_next/anything` → `/_next/*`, and add a case for the new `"unmatched"` bucket (e.g. `normalizeRoute("/favicon.ico")` → `"unmatched"`, `normalizeRoute("/wp-login.php")` → `"unmatched"`).

frontend/proxy.ts
```diff
@@ -0,0 +1,10 @@
+import { type NextRequest, NextResponse } from "next/server";
```
mpawlow: ### (2b) [Minor] `frontend/proxy.ts` is now dead code (a pure no-op)

#### Problem

- After moving metrics recording into `instrumentation.ts`'s `diagnostics_channel` subscription, `proxy()` does nothing but `return NextResponse.next()` for every non-HMR request — it has no remaining side effects, and its `config.matcher` no longer gates anything meaningful.
- `frontend/proxy.ts` did not exist before this PR (confirmed via `git log --all -- frontend/proxy.ts`) — it was introduced and then fully hollowed out within this same PR's commits, so it's pure leftover scaffolding, not a file serving some other pre-existing purpose.
- Keeping an empty Proxy file still costs a middleware invocation on every request in production for zero benefit, and invites future confusion about why it exists.

#### Potential Solution

- Delete `frontend/proxy.ts` entirely — Next.js does not require a Proxy/middleware file to be present.

mpawlow (approved):
## Code Review 2

- ✅  Approved / LGTM  🚀 
- See Minor PR comments (2a) (2b) for optional consideration

PR #2275, opened 2026-08-20
fix: remove local storage for toggles ingest

Wallgau:
Summary
Ignore those removed flags on read and rewrite openrag.ingest-preview.settings to persist only autoOpen, so users who previously saved false get the preview chrome back without clearing localStorage.

## Summary by CodeRabbit

* **Changes**
  * Ingest preview display toggles are always enabled and are no longer user-configurable.
  * Existing saved settings are automatically cleaned up, while the “auto-open” preference remains available.
  * Settings updates now focus exclusively on the auto-open option.
  * Improved settings handling preserves valid preferences and fallback behavior, even when cleanup cannot be completed.


mfortman11 (approved):
LGTM

PR #2276, opened 2026-08-21
fix: Code review for status page

jason-dong03:
-idiomatic design: TanStack Query V5 uses
-helper function for shared error api responses across 3 status hooks
- refactoring into helper functions, breaking up components into smaller components for better maintainability, etc

## Summary by CodeRabbit

* **New Features**
  * Added a console status panel showing component and provider health, details, diagnostics, synchronization, and logs.
  * Added expandable component cards with refresh, diagnosis, and log-viewing actions.

* **Bug Fixes**
  * Improved error messages for synchronization, status checks, and log-loading failures.
  * Unknown component log requests now provide clearer not-found details.

* **Improvements**
  * Standardized status colors, icons, badges, and relative-time labels.
  * Improved task notification indicators and menu coordination.
  * Enhanced health logging, loading states, diagnostics, and log display behavior.


src/services/component_logs.py
```diff
@@ -119,25 +119,33 @@ def record(component: str, level: str, message: str, detail: str | None = None)
 
 def record_check_result(
     component: str,
-    ok: bool,
+    state: str,
```
mfortman11: I believe this should be `ComponentState`

The parameter accepts any string but every caller passes a `ComponentState enum` value. The defensive str(state) cast on line 134 is a workaround for the imprecise type. Violates `CONTRIBUTING.md`: "Use type hints." Typing as `ComponentState` catches misuse at type-check time and removes the cast.

mfortman11 (changes requested):

frontend/instrumentation.ts
mfortman11, line 1: The runtime guard is correct at execution time, but Next.js also analyzes `instrumentation.ts` while building the Edge bundle. Because the dynamic imports use literal Node module names, Next still sees node:http and node:diagnostics_channel during that analysis and emits Edge compatibility warnings.

`export const runtime = "nodejs"` does not fix this because Next.js does not apply route-segment runtime configuration to `instrumentation.ts`.

Please keep `instrumentation.ts` as a small runtime dispatcher and move all Node-specific code into a separate file, such as `instrumentation.node.ts`:

```
  // instrumentation.ts
  export async function register() {
    if (process.env.NEXT_RUNTIME === "nodejs") {
      const { registerNodeInstrumentation } =
        await import("./instrumentation.node");
      await registerNodeInstrumentation();
    }
  }
```

```
  // instrumentation.node.ts
  import { createServer } from "node:http";
  import { subscribe } from "node:diagnostics_channel";

  export async function registerNodeInstrumentation() {
    // Existing metrics server and subscription setup
  }
```
This gives Next.js a clean Edge-compatible entry point while loading the Node-only implementation exclusively for the Node runtime. Please also remove the runtime export and its comment, then run next build
and confirm the Edge-runtime warnings are gone.

The key distinction is that the existing guard prevents Node code from running on Edge, but it does not prevent the bundler from seeing and analyzing that Node code. Moving it into a runtime-only module
solves the build-time problem.

mfortman11 (approved):
LGTM

PR #2291, opened 2026-08-26
feat: task notification overhaul — persistent delete, count badge, click-to-expand, light/dark mode fixes

jason-dong03:
UI Changes:

Console button in sync with light/dark mode 
Removed console status from task notification menu (redundant)
Added ability to clear/delete tasks from notification -> updating notification number accordingly 
clear all button introduced 
aligned indivudal task ui to include "x" buttons,

Quick Fixes:
-ensured localStorage doesnt exist on server, but next js runs 'use client' components on the server first during SSR. -Checked 'typeof window !== 'undefined' before accessing localStorage so the server gets a safe fallback and the real value is only read in browser 
-Console Status button and "Active" connector badge used hardcoded dark-mode zinc/foreground colors that broke in light mode

Screenshot 2026-09-14 at 8 19 35 PM

## Summary by CodeRabbit

- **New Features**
  - Added options to delete individual completed or failed tasks, or clear all terminal tasks.
  - Task notifications now show task counts, expandable progress details, and deletion controls.
  - Added a “Clear all” action when terminal tasks are available.
  - The header notification badge displays the number of tasks, up to 99+.
  - Deletion now distinguishes unavailable tasks from tasks still in progress.

- **Bug Fixes**
  - Improved keyboard and click interactions for expanding task error details.

- **Style**
  - Updated connector and console status styling to better support themes and muted states.


src/services/task_service.py
```diff
@@ -1542,6 +1549,55 @@ async def cancel_task(self, user_id: str, task_id: str) -> bool:
 
         return True
 
+    def delete_task(self, user_id: str, task_id: str) -> TaskDeleteResult:
```
mfortman11, line 1621: Lets add some tests for this function

frontend/app/api/mutations/useDeleteTaskMutation.ts
mfortman11, line 1: lets make sure we're type safe here you can look at `task-context.tsx` as an example

frontend/components/task-error-content.tsx
```diff
@@ -83,60 +83,36 @@ export function TaskErrorContent({
...
-        {accordionSummary}
-      </AccordionPrimitive.Trigger>
-      {openTaskDialogButton}
-    </AccordionPrimitive.Header>
-  );
+  const toggleAccordion = () =>
+    setAccordionValue((v) => (v === "failed-files" ? "" : "failed-files"));
 
   return (
     <div
```
mfortman11: A focusable `div role="button"` contains a real button, producing nested interactive controls with ambiguous focus and activation semantics. Restore a native accordion trigger or make a non-overlapping native button the row trigger.

mfortman11 (changes requested):

frontend/components/header.tsx
```diff
@@ -106,7 +94,11 @@ export function Header() {
                 isCloudBrand ? "text-foreground" : "text-muted-foreground"
               }
             />
-            {showNotificationDot && <div className="header-notifications" />}
+            {tasks.length > 0 && (
+              <span className="absolute -right-1 -top-1 flex min-w-[16px] h-4 items-center justify-center rounded-full bg-destructive px-1 text-[10px] font-semibold leading-none text-white">
```
Wallgau: The badge uses bg-destructive only. A successful completed upload keeps a red, not sure we want that for complete successful ones?

frontend/components/header.tsx
```diff
@@ -23,7 +23,6 @@ export function Header() {
   const { runMode } = useAuth();
 
   const {
-    hasProblem,
```
Wallgau, line 26: the bell no longer signals console health. On SaaS the Console Status button is hidden (runMode === "oss" only), so those problems disappear from the header entirely.

Wallgau (changes requested):

frontend/app/api/mutations/useDeleteTaskMutation.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export function useDeleteAllTerminalTasksMutation() {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: async (): Promise<string[]> => {
+      const res = await fetch("/api/tasks", { method: "DELETE" });
+      if (!res.ok) throw new Error("Failed to delete tasks");
+      const body = await res.json();
+      return (body.deleted_ids as string[]) ?? [];
```
Wallgau: body.deleted_ids as string[] is an unsafe cast. If the payload is missing or mistyped, Clear all will treat deletedIds as [] and leave the cache unchanged with no error. Validate the JSON (or use a typed parser) before updating the query cache.

src/services/task_service.py
```diff
@@ -1549,6 +1556,55 @@ async def cancel_task(self, user_id: str, task_id: str) -> bool:
...
+        """Remove a terminal (completed/failed) task from memory.
+
+        Returns:
+            TaskDeleteResult.DELETED      – task found and removed.
+            TaskDeleteResult.NOT_FOUND    – task ID does not exist.
+            TaskDeleteResult.IN_PROGRESS  – task exists but is not yet terminal.
+        """
+        resolved = self._resolve_upload_task_store(user_id, task_id)
```
Wallgau, line 1632: delete_task should be scoped strictly to the authenticated caller, matching the behavior of Clear all.

The fix is:
def delete_task(self, user_id, task_id):
    user_store = self.task_store.get(user_id)

    if user_store is None or task_id not in user_store:
        return NOT_FOUND

    del user_store[task_id]

delete_task should not call _resolve_upload_task_store or fall back to AnonymousUser().user_id. If the task exists only in the anonymous bucket, deletion by a named user must return NOT_FOUND.

That removes the cross-user deletion path while leaving the anonymous/shared lookup behavior available for any flows that genuinely need it.

Wallgau (changes requested):

tests/unit/test_task_service.py
```diff
@@ -324,3 +324,140 @@ def test_enhanced_keeps_completed_files_for_preview_tasks(task_service):
...
+
+def test_delete_task_resolves_anonymous_shared_task(task_service):
+    """delete_task must find tasks stored under the anonymous key."""
+    from models.tasks import TaskDeleteResult
+    from session_manager import AnonymousUser
+
+    anon_id = AnonymousUser().user_id
+    task = _make_terminal_task("shared1", TaskStatus.COMPLETED)
+    task_service.task_store[anon_id] = {"shared1": task}
+    assert task_service.delete_task("real_user", "shared1") is TaskDeleteResult.DELETED
```
Wallgau, line 404: asserts DELETED for an anonymous-store task; delete_task only looks at user_id and this test fails (NOT_FOUND).

src/services/task_service.py
```diff
@@ -1549,6 +1556,60 @@ async def cancel_task(self, user_id: str, task_id: str) -> bool:
...
+        different bucket (e.g. the anonymous key) are not visible here; callers
+        that need cross-user lookup should use _resolve_upload_task_store
+        directly.
+
+        Returns:
+            TaskDeleteResult.DELETED      – task found and removed.
+            TaskDeleteResult.NOT_FOUND    – task ID does not exist for this user.
+            TaskDeleteResult.IN_PROGRESS  – task exists but is not yet terminal.
+        """
+        user_store = self.task_store.get(user_id)
```
Wallgau: single delete cannot see shared anonymous tasks that get_all_tasks still lists, so the panel X 404s and the row stays.

frontend/components/task-notification-menu.tsx
```diff
@@ -451,94 +433,150 @@ export function TaskNotificationMenu() {
...
+                          defaultExpanded={shouldExpandDetails}
+                        />
+                        <button
+                          type="button"
+                          aria-label="Delete task"
+                          onClick={(e) => {
+                            e.stopPropagation();
+                            deleteTaskMutation.mutate(task.task_id);
+                          }}
+                          className="absolute top-2 right-2 flex h-5 w-5 items-center justify-center rounded text-muted-foreground opacity-0 group-hover/row:opacity-100 hover:bg-muted hover:text-foreground transition-opacity focus-visible:opacity-100 focus-visible:ring-1 focus-visible:ring-ring"
```
Wallgau: The X is right-2 (8px) × 20px wide → occupies ~8–28px from the right. Details is right-4 (16px) × 16px → occupies 16–32px. Same corner. The X is later in the DOM, so on hover it paints on top of "Open task details". A click meant to open the dialog deletes the task instead.

frontend/components/task-error-content.tsx
```diff
@@ -160,16 +122,60 @@ export function TaskErrorContent({
...
               </div>
-              <p className="min-h-4 text-xxs leading-4 text-muted-foreground whitespace-nowrap">
-                {formatTaskTimestamp(timestamp, mode, nowMs)}
-              </p>
             </div>
           </div>
+        </button>
+      )}
+      {showHeader && (
+        <div className="absolute right-4 top-1/2 -translate-y-1/2 flex items-center gap-1.5">
```
Wallgau: top-1/2 is 50% of the relative parent. That parent now includes the header and the accordion (px-4 pb-mmd + file list). Collapsed, extra bottom padding pulls the pill off the title row (the invisible spacer at 127–135 is on the title line; the real pill is not). Expanded, the details icon sits in the middle of the failed-file list.

frontend/app/api/mutations/useDeleteTaskMutation.ts
```diff
@@ -0,0 +1,43 @@
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { TASKS_QUERY_KEY, type Task } from "@/app/api/queries/useGetTasksQuery";
+
+export function useDeleteTaskMutation() {
+  const queryClient = useQueryClient();
+  return useMutation({
+    mutationFn: async (taskId: string) => {
+      const res = await fetch(`/api/tasks/${taskId}`, { method: "DELETE" });
+      if (!res.ok) throw new Error("Failed to delete task");
```
Wallgau: !res.ok is true for 404 and 409. throw stops onSuccess, so the row is not removed from the cache. React Query stores the error on the mutation.
Nobody reads that error. There is no onError, no toast, and the caller is fire-and-forget

Wallgau (changes requested):

frontend/app/api/mutations/useDeleteTaskMutation.ts
jason-dong03: wait so what should be addressed here? am I removing the throw Error so onSuccess remains in cache?

Wallgau: fix suggestion:
if (res.status === 404) return; // already gone and still drop from cache
  if (!res.ok) {
    const body = await res.json().catch(() => ({}));
    throw new Error(body.error || "Failed to delete task");
  }

res.json() parses the HTTP body. That can fail when the body is empty, not JSON, or already consumed. .catch(() => ({})) turns that into an empty object so the next line can still do body.error || "Failed to delete task" without crashing.

frontend/components/task-error-content.tsx
```diff
@@ -160,16 +120,64 @@ export function TaskErrorContent({
...
+                </span>
               </div>
-              <p className="min-h-4 text-xxs leading-4 text-muted-foreground whitespace-nowrap">
-                {formatTaskTimestamp(timestamp, mode, nowMs)}
-              </p>
             </div>
           </div>
+        </button>
+      )}
+      {showHeader && (
```
Wallgau: The two adjacent showHeader && fragments split the expand <button> from the details overlay, but they hang off div.w-full, which is not relative. The overlay is absolute right-4 top-0 h-full, so it does not attach to the header button (the one that is relative). It attaches to group/row in the notification menu, or to a transformed motion.div in FailedTasksInfo.

Collapsed, accordion pb-mmd pulls the pill off the title row. Expanded, the details icon sits in the middle of the failed-file list. In the knowledge ingestion-status panel there is no relative wrapper at all, so the control can pin to the top-right of the whole 360px panel.

The invisible spacer (lines 123–133) is a workaround for that overlay: it duplicates Failed / Warning / Complete in the DOM. getByText("Failed", { exact: true }) matches both copies, so tasks-unified-panel.spec.ts (toHaveCount(2) for two tasks) will see four.

One showHeader branch is enough: a relative header wrapper that is only as tall as the header, with the expand button and details button as siblings. That also removes the invisible duplicate pill spacer (lines 123–133), which is only there to leave a hole for this overlay.

Suggested fix:
{showHeader && (
  <div
    className={cn(
      "flex w-full items-start",
      isCloudBrand ? "border-t border-muted" : "rounded-mmd border border-muted",
    )}
  >
    <button
      type="button"
      onClick={toggleAccordion}
      aria-expanded={isExpanded}
      className="min-w-0 flex-1 text-left py-mmd px-4 transition-colors hover:bg-muted/60"
    >
      {/* icon + title + timestamp + summary + chevron — no spacer, no pill */}
    </button>
    <div className="flex shrink-0 items-center gap-1.5 py-mmd pr-4">
      {!isExpanded && (
        <p className={statusPillClassName}>{statusLabel}</p>
      )}
      <button
        type="button"
        aria-label="Open task details"
        className="inline-flex shrink-0 items-center justify-center text-muted-foreground hover:text-foreground focus-visible:ring-1 focus-visible:ring-ring rounded"
        onClick={() => openTaskDialog(task.task_id)}
      >
        <IncidentReporterIcon className="size-4" />
      </button>
    </div>
  </div>
)}

frontend/components/task-notification-menu.tsx
```diff
@@ -451,94 +433,150 @@ export function TaskNotificationMenu() {
...
+                          defaultExpanded={shouldExpandDetails}
+                        />
+                        <button
+                          type="button"
+                          aria-label="Delete task"
+                          onClick={(e) => {
+                            e.stopPropagation();
+                            deleteTaskMutation.mutate(task.task_id);
+                          }}
+                          className="absolute top-2 right-10 flex h-5 w-5 items-center justify-center rounded text-muted-foreground opacity-0 group-hover/row:opacity-100 hover:bg-muted hover:text-foreground transition-opacity focus-visible:opacity-100 focus-visible:ring-1 focus-visible:ring-ring"
```
Wallgau: right-10 no longer overlaps the details icon at right-4, but it sits on the status pill immediately left of that icon. On hover, clicking Failed/Complete can delete the task. Success rows collide the same way with the badge/chevron (pr-8 vs right-10).

Do not position the X with a magic right-*. Put it in normal document flow next to the other header actions, and remove both absolute X buttons.

This file has two row types:

Failed / partial rows (this X at 476–486) go through TaskErrorContent. After the header-flex change on task-error-content.tsx, pass the X in as a slot so it sits in the same flex as the pill and details icon.
Success rows (the X at 549–559) do not use TaskErrorContent. Those need a flex sibling of the expand button.
Suggested fix for failed rows (depends on adding headerEnd to TaskErrorContent and rendering it in the header actions flex, before the details button):
<TaskErrorContent
  ...
  headerEnd={
    <button
      type="button"
      aria-label="Delete task"
      onClick={() => deleteTaskMutation.mutate(task.task_id)}
      className="flex h-5 w-5 items-center justify-center rounded text-muted-foreground opacity-0 group-hover/row:opacity-100 hover:bg-muted hover:text-foreground transition-opacity focus-visible:opacity-100 focus-visible:ring-1 focus-visible:ring-ring"
    >
      <X className="size-3" />
    </button>
  }
/>

frontend/components/task-notification-menu.tsx
```diff
@@ -451,94 +433,150 @@ export function TaskNotificationMenu() {
...
+                        </div>
+                      </button>
+                      <button
+                        type="button"
+                        aria-label="Delete task"
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          deleteTaskMutation.mutate(task.task_id);
+                        }}
+                        className="absolute top-2 right-10 flex h-5 w-5 items-center justify-center rounded text-muted-foreground opacity-0 group-hover/row:opacity-100 hover:bg-muted hover:text-foreground transition-opacity focus-visible:opacity-100 focus-visible:ring-1 focus-visible:ring-ring"
```
Wallgau: Same absolute top-2 right-10 overlay as the failed-row X.

That row’s layout is:

Expand <button> with pr-8 (32px of padding on the right)
Inside it, on the right: status badge + chevron
Sibling X at right-10 (40px from the right edge), 20px wide → occupies 40–60px from the right
The badge and chevron sit in that same band. On hover the X appears on top of them. Clicking Complete / the chevron can delete the task instead of expanding it.

Failed rows hit the pill. Success rows hit the badge/chevron. Same cause: magic right-10 instead of putting the X in the row’s flex. This row does not use TaskErrorContent, so headerEnd does not apply here.
Suggested fix:

<div key={task.task_id} className="relative group/row flex items-start">
  <button
    type="button"
    onClick={toggleExpand}
    className={cn(pastTaskRowClass, "min-w-0 flex-1 text-left")}
  >
    {/* existing row content */}
  </button>
  <button
    type="button"
    aria-label="Delete task"
    onClick={() => deleteTaskMutation.mutate(task.task_id)}
    className="mt-2 mr-2 flex h-5 w-5 shrink-0 items-center justify-center rounded text-muted-foreground opacity-0 group-hover/row:opacity-100 hover:bg-muted hover:text-foreground transition-opacity focus-visible:opacity-100 focus-visible:ring-1 focus-visible:ring-ring"
  >
    <X className="size-3" />
  </button>
</div>

src/services/task_service.py
```diff
@@ -1762,6 +1769,58 @@ async def cancel_file(self, user_id: str, task_id: str, file_path: str) -> bool:
...
+            TaskDeleteResult.NOT_FOUND    – task ID does not exist for this user.
+            TaskDeleteResult.IN_PROGRESS  – task exists but is not yet terminal.
+        """
+        resolved = self._resolve_upload_task_store(user_id, task_id)
+        if resolved is None:
+            return TaskDeleteResult.NOT_FOUND
+        store_user_id, upload_task = resolved
+        if upload_task.status not in [TaskStatus.COMPLETED, TaskStatus.FAILED]:
+            return TaskDeleteResult.IN_PROGRESS
+        self._cleanup_upload_temp_files(upload_task, force=True)
```
mfortman11: ### [Medium] Let users delete cancelled tasks

Cancelled tasks appear under Past Tasks, but users cannot delete them. The Clear all button also skips them.

Update the deletion logic so TaskStatus.CANCELLED is treated like completed and failed tasks.

src/services/task_service.py
```diff
@@ -1762,6 +1769,58 @@ async def cancel_file(self, user_id: str, task_id: str, file_path: str) -> bool:
...
+
+        Only touches the calling user's own store. Shared tasks stored under
+        the anonymous key are intentionally excluded: they are visible to all
+        authenticated users, so a bulk clear by one user must not remove them
+        for everyone else. Those tasks are aged out by cleanup_old_tasks.
+
+        Returns the list of deleted task IDs.
+        """
+        if user_id not in self.task_store:
+            return []
```
mfortman11, line 1819: ### [Medium] Make “Clear all” handle shared tasks

Signed-in users can see some shared anonymous tasks. However, Clear all only deletes tasks owned by that user, so shared tasks remain visible.

Either:

  - Let users hide shared tasks for their own account, or
  - Do not include shared tasks when deciding whether to show or run Clear all.

mfortman11 (changes requested):

mfortman11 (approved):

Wallgau (approved):
Nice work!

PR #2308, opened 2026-08-28
fix: Match handling of ingest of image only doc w/o langflow

mfortman11:
If 
```
Disabled Langflow Ingestion: True
Picture Descriptions: True
```
Image only pdf upload works as expected

```
Disabled Langflow Ingestion: True
Picture Descriptions: False
```
Upload fails with The file appears corrupted or invalid and cannot be processed. Upload a valid file.

if `Disable Langflow Ingestion` is off it processes the file w/ langflow and returns an empty `<!-- image -->` tag this PR matches this expected behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * Uncaptioned images no longer add placeholder content when document text is available.
  * Image-only documents now return a clear no-text error instead of triggering unnecessary processing.
  * Image-only documents retain a placeholder for the first image when no text can be extracted.
  * Failed image processing now recommends enabling OCR in Settings > Knowledge Base and retrying ingestion.

* **Tests**
  * Added coverage for image-only documents, uncaptioned images, and OCR guidance.

### DIFF
--- a/src/utils/document_processing.py
+++ b/src/utils/document_processing.py
@@ -147,15 +147,12 @@ def extract_relevant(doc_dict: dict) -> dict:
             for ann in picture.get("annotations", [])
             if ann.get("kind") == "description" and ann.get("text", "").strip()
         ]
-        if not descriptions:
-            continue
-
         chunks.append(
             {
                 "page": page_no,
                 "type": "picture",
                 "picture_index": p_idx,
-                "text": "\n".join(descriptions),
+                "text": "\n".join(descriptions) if descriptions else "<!-- image -->",
             }
         )
 
--- a/src/utils/document_processing.py
+++ b/src/utils/document_processing.py
@@ -5,6 +5,8 @@ from utils.logging_config import get_logger
 
 logger = get_logger(__name__)
 
+IMAGE_PLACEHOLDER = "<!-- image -->"
+
 
 def process_text_file(file_path: str) -> dict:
     """
@@ -82,6 +84,8 @@ def extract_relevant(doc_dict: dict) -> dict:
       - Finds every text fragment in `texts`, groups them by page_no
       - Flattens tables in `tables` into tab-separated text, grouping by row
       - Concatenates each page's fragments and each table into its own chunk
+      - Emits a picture chunk for each picture with description annotations
+      - Emits one image placeholder only when pictures exist and no other chunks were produced
     Returns a slimmed dict ready for indexing, with each chunk under "text".
     """
     origin = doc_dict.get("origin", {})
@@ -136,7 +140,8 @@ def extract_relevant(doc_dict: dict) -> dict:
     # Without this, image-only documents (no text layer) yield zero chunks on the
     # non-Langflow path, which surfaces as a misleading "corrupted or invalid" error.
     # The Langflow path already captures them via Docling's export_to_markdown.
-    for p_idx, picture in enumerate(doc_dict.get("pictures", [])):
+    pictures = doc_dict.get("pictures", [])
+    for p_idx, picture in enumerate(pictures):
         prov = picture.get("prov", [])
         page_no = prov[0].get("page_no") if prov else None
         if page_no is None:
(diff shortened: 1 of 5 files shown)


src/utils/document_processing.py
```diff
@@ -147,15 +147,12 @@ def extract_relevant(doc_dict: dict) -> dict:
...
-        if not descriptions:
-            continue
-
         chunks.append(
             {
                 "page": page_no,
                 "type": "picture",
                 "picture_index": p_idx,
-                "text": "\n".join(descriptions),
+                "text": "\n".join(descriptions) if descriptions else "<!-- image -->",
```
mpawlow: ### (1a) [Normal] Placeholder emitted per description-less picture inflates the index with duplicate `<!-- image -->` chunks/embeddings

#### Problem

- Before this PR (`#2172`), `extract_relevant()` deliberately skipped pictures with no/empty description "so no blank chunks are created". This PR reverses that for **all** documents, not just image-only ones.
- Every picture without a VLM description now yields a standalone chunk whose text is the literal `<!-- image -->`.
  - In `src/models/processors.py`, these chunks are non-empty and survive both whitespace filters (lines ~532 and ~553), so each one is embedded (`patched_embedding_client.embeddings.create`) and indexed as its own OpenSearch document.
- For an image-heavy document with picture descriptions disabled (a scanned catalog, a slide deck exported to PDF, etc.) this produces **N near-identical `<!-- image -->` chunks / embeddings**:
  - Wasted embedding API calls and vector storage.
  - Retrieval pollution — many identical zero-information chunks can crowd out useful hits in the results window (the same failure mode CLAUDE.md calls out for `utils/opensearch_filenames.py`: "a many-chunk document crowds others out of the hits window").
- This also diverges from the Langflow path it claims to match: Langflow's `export_to_markdown` emits `<!-- image -->` **inline inside the page's text chunk**, not as a separate embedded chunk per image.

#### Background Information

- CodeRabbit flagged no actionable comments and rated merge risk minimal; it did not consider the fan-out / retrieval-noise effect on multi-image documents.
- `src/services/document_service.py:212-220` also consumes these chunks: `pages = len(slim_doc["chunks"])` and each chunk is rendered as `Page N:\n<!-- image -->` into extracted context, inflating both the page count and the context payload.

#### Code References

- `src/utils/document_processing.py:150-157`
- `src/models/processors.py:529-556`
- `src/services/document_service.py:210-221`

#### Potential Solution

- Only rescue genuinely empty documents: emit the placeholder **only when the document would otherwise yield zero chunks** (e.g. after building `chunks`, if `not chunks` and `doc_dict.get("pictures")`, append a single `<!-- image -->` chunk).
- This preserves the fix's intent (image-only docs no longer error) without adding placeholder chunks to text documents that already index fine.

#### Alternative Solutions

- Emit at most **one** `<!-- image -->` chunk per document regardless of picture count (dedupe), keeping the "document is non-empty" signal while avoiding the N-duplicate blow-up.
- Keep per-picture placeholders but mark them non-embeddable / exclude `type == "picture"` chunks with placeholder text from the embedding batch while still writing them to the index for provenance (more invasive).

mpawlow: ### (1b) [Normal] Silent "success" with non-searchable content masks the actionable "enable OCR" guidance

#### Problem

- Previously, an image file (`.png`/`.jpeg`/…) or image-only PDF ingested with **both** OCR and picture descriptions disabled failed with `"No text content could be extracted from document"`, which `services/task_service.py:1016-1027` turns into the actionable message *"Enable OCR in Settings > Knowledge Base and retry ingestion."*
- After this PR the same upload "succeeds": a `<!-- image -->` placeholder chunk is produced, embeddings are non-empty, and the `No embeddings generated` guard at `src/models/processors.py:567-573` no longer trips.
- Net effect: the user believes the document is ingested and searchable, but the index contains only an HTML-comment placeholder with zero retrievable content, and the "turn on OCR" hint is never surfaced.

#### Code References

- `src/models/processors.py:567-573`
- `src/services/task_service.py:1011-1027`
- `src/utils/document_processing.py:150-157`

#### Potential Solution

- Combine with (1a): if the only chunks produced are `<!-- image -->` placeholders (no text/table chunks and no real picture descriptions), keep returning the `"No text content could be extracted from document"` error so the OCR tip still fires — i.e. the placeholder should unblock *mixed* documents, not make pure-image uploads report false success.
- If product intent really is "pure-image upload should succeed silently to match Langflow", document that decision in the code comment and confirm the Langflow-on path behaves identically for a zero-text image (it likewise yields only `<!-- image -->`).

#### Alternative Solutions

- Leave behavior as-is but have the ingest result flag documents whose chunks are all placeholders (e.g. a `warning` on the task) so the UI can still prompt the user to enable OCR.

src/utils/document_processing.py
mpawlow, line 82: ### (1c) [Minor] `extract_relevant()` docstring not updated for picture / placeholder behavior

#### Problem

- The docstring (`src/utils/document_processing.py:79-86`) still only describes `texts` and `tables` handling; it never mentions the picture-annotation pass, and now omits the new placeholder semantics.
- CodeRabbit's pre-merge "Docstring Coverage" check is already at 40% (threshold 80%) for functions touched by this diff.

#### Potential Solution

- Add a bullet: "Emits one `picture` chunk per picture — joined `description` annotations when present, otherwise a `<!-- image -->` placeholder."

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
   - See Normal PR comments (1a), (1b) for consideration

PR #2312, opened 2026-08-31
fix: Creates task for fetch docs w/ langflow ingest disabled

mfortman11:
The Issue/Previous Behavior:
- When Langflow ingestion is used a task shows in the task panel when fetching the latest OpenRAG docs
- When Langflow Ingest is disabled there is no task created. Docs are still properly fetched but the status isn't well communicated to users

How to test
- Turn on `Disable Langflow Ingest`
- Click the `Fetch latest docs` button in the KT
- Confirm there was a task created in the task panel

## Summary by CodeRabbit

* **New Features**
  * Added scheduled URL ingestion with configurable crawl depth, ownership details, connector settings, sample-data handling, and task tracking.
  * Added clear completion and failure statuses for URL processing, including error details and progress counters.
  * Temporary files are automatically removed after processing.

* **Improvements**
  * Default documents now start in a separate task instead of being appended to URL ingestion tasks.

### DIFF
--- a/src/models/url.py
+++ b/src/models/url.py
@@ -62,7 +62,7 @@ class UrlProcessor(TaskProcessor):
                 is_sample_data=self.is_sample_data,
                 connector_type=self.connector_type,
             )
-            await processor.process_document_standard(
+            result = await processor.process_document_standard(
                 file_path=temp_file_path,
                 file_hash=hash_id(temp_file_path),
                 owner_user_id=self.owner_user_id,
@@ -75,9 +75,16 @@ class UrlProcessor(TaskProcessor):
                 is_sample_data=self.is_sample_data,
             )
 
-            file_task.status = TaskStatus.COMPLETED
-            file_task.updated_at = time.time()
-            upload_task.successful_files += 1
+            if result.get("status") == "error":
+                file_task.status = TaskStatus.FAILED
+                file_task.error = result.get("error") or "Failed to process document"
+                file_task.updated_at = time.time()
+                upload_task.failed_files += 1
+            else:
+                file_task.status = TaskStatus.COMPLETED
+                file_task.result = result
+                file_task.updated_at = time.time()
+                upload_task.successful_files += 1
 
         except Exception as e:
             file_task.status = TaskStatus.FAILED
--- a/src/models/url.py
+++ b/src/models/url.py
@@ -1,3 +1,4 @@
+import os
 import time
 from typing import Any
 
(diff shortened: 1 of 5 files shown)


src/models/url.py
```diff
@@ -9,6 +10,95 @@
...
+                self.document_service,
+                models_service=self.models_service,
+                owner_user_id=self.owner_user_id,
+                jwt_token=self.jwt_token,
+                owner_name=self.owner_name,
+                owner_email=self.owner_email,
+                is_sample_data=self.is_sample_data,
+                connector_type=self.connector_type,
+            )
+            await processor.process_document_standard(
```
mpawlow: ### (1a) [Normal] `UrlProcessor` ignores error result from `process_document_standard`

- ℹ️ Concur with CodeRabbit finding

#### Problem

- `DocumentFileProcessor.process_document_standard` signals the "no text content could be extracted" failure by **returning** `{"status": "error", "error": ...}` rather than raising (`src/models/processors.py:573`).
- Every other caller checks for this: `DocumentFileProcessor.process_item` (`processors.py:828`), `processors.py:1329`, `processors.py:1434` all branch to `TaskStatus.FAILED` / `failed_files += 1` on `result.get("status") == "error"`.
- `UrlProcessor.process_item` discards the return value entirely and unconditionally sets `file_task.status = TaskStatus.COMPLETED` and `upload_task.successful_files += 1`.
- Net effect: a docs fetch that produced no usable content is reported to the user as a **successful** task — directly contradicting the PR's stated goal of communicating fetch status accurately.

#### Background Information

- The exception path is handled correctly (`except Exception` re-raises after marking `FAILED`); only the non-raising error-return path is missed.

#### Code References

- `src/models/url.py:65-80`
- `src/models/processors.py:573`, `src/models/processors.py:828-836`

#### Potential Solution

- Capture the result and mirror the sibling processors:
  ```python
  result = await processor.process_document_standard(...)
  if result.get("status") == "error":
      file_task.status = TaskStatus.FAILED
      file_task.error = result.get("error") or "Failed to process document"
      upload_task.failed_files += 1
      return
  file_task.status = TaskStatus.COMPLETED
  file_task.result = result
  upload_task.successful_files += 1
  ```

src/services/default_docs_service.py
```diff
@@ -311,55 +308,33 @@ async def _ingest_default_documents_url(
...
+    anonymous_user = AnonymousUser()
+
     logger.info(
         "Running default URL docs ingestion with OpenRAG processor",
         docs_url=docs_url,
         crawl_depth=crawl_depth,
     )
-    temp_file_path, title = await materialize_url_as_text_file(
+
+    task_id = await task_service.create_url_upload_task(
```
mpawlow, line 317: ### (1b) [Normal] URL upload task is reused for local-file ingestion (wrong processor on retry; new concurrency race)

- ℹ️ Concur with CodeRabbit finding

#### Problem

- In `ingest_default_documents_when_ready` the task id returned by `_ingest_default_documents_url` (whose `upload_task.processor` is a `UrlProcessor`) is passed as `existing_task_id` to `_ingest_default_documents_openrag`, which appends the bundled local files to the **same** `UploadTask`.
- `create_custom_task`'s existing-task branch (`task_service.py:448-456`) merges `file_tasks` and bumps `total_files` but never updates `upload_task.processor`. It stays `UrlProcessor`.
- `retry_failed_files` uses `upload_task.processor` for all retried items (`task_service.py:868`; docstring: "This reuses the task's original processor"). If a local bundled-doc file fails and the user clicks retry in the (now visible) task panel, `UrlProcessor.process_item` runs for that local path: it **ignores the item**, re-fetches `docs_url`, and marks the local file task `COMPLETED`.
- Additional regression specific to this PR: `_ingest_default_documents_url` previously **awaited** ingestion inline, so local-file ingestion started only after the URL doc was fully indexed. It now returns immediately after spawning the background task, so `_ingest_default_documents_openrag` spawns a **second** `background_custom_processor` on the same `task_id` concurrently. `total_files` is incremented outside the task lock while the first run evaluates `processed_files >= total_files`, so the task can briefly flip to `COMPLETED` (running `_cleanup_upload_temp_files`) and back to `RUNNING`, with duplicated completion side effects.

#### Background Information

- The processor-reuse pattern is **pre-existing** in the Langflow path (`_ingest_default_documents_url_langflow` + `_ingest_default_documents_langflow` share a task the same way), so this is a replication of an existing latent bug rather than a wholly new one. The concurrency race, however, is newly introduced for the non-Langflow path.

#### Code References

- `src/services/default_docs_service.py:144-154`, `src/services/default_docs_service.py:321-330`
- `src/services/task_service.py:448-456`, `src/services/task_service.py:486-494`, `src/services/task_service.py:835-868`

#### Potential Solution

- Give URL and local-file ingestion **separate** tasks (drop `existing_task_id` threading between the two), so each task carries a processor that matches its items.

#### Alternative Solutions

- Persist the processor per `FileTask` (or a processor map on the `UploadTask`) and have `retry_failed_files` / `background_custom_processor` dispatch per item. Larger change; also fixes the pre-existing Langflow-path variant.
- At minimum, if the single-task UX is desired, await the URL task to completion before appending local files (restores prior ordering and removes the race) and refuse retry on mixed-processor tasks.

src/models/url.py
```diff
@@ -9,6 +10,95 @@
...
+                original_filename=title,
+                jwt_token=self.jwt_token,
+                owner_name=self.owner_name,
+                owner_email=self.owner_email,
+                file_size=os.path.getsize(temp_file_path),
+                connector_type=self.connector_type,
+                is_sample_data=self.is_sample_data,
+            )
+
+            file_task.status = TaskStatus.COMPLETED
```
mpawlow: ### (1c) [Minor] `UrlProcessor` does not record `file_task.result` on success

#### Problem

- Sibling processors set `file_task.result = result` on success (`processors.py:835`, `processors.py:240` in `LangflowUrlProcessor`).
- `UrlProcessor.process_item` never assigns `file_task.result`, so downstream consumers (index-proof lookups, UI result display) see an empty result for these tasks.

#### Code References

- `src/models/url.py:78-80`

#### Potential Solution

- Capture the `process_document_standard` return value (already needed for 1a) and assign `file_task.result = result` before marking `COMPLETED`.

src/models/url.py
```diff
@@ -9,6 +10,95 @@
 logger = get_logger(__name__)
 
 
+class UrlProcessor(TaskProcessor):
```
mpawlow, line 13: ### (1d) [Minor] No test coverage for `UrlProcessor` or the rerouted default-docs URL path

#### Problem

- The PR adds a new processor class and changes the control flow of `_ingest_default_documents_url` with no accompanying tests.
- Per repo convention (`CLAUDE.md`: "`tests/unit/` mirrors `src/` … Put a new test where its module lives"), a `tests/unit/models/test_url.py` (or extension of `tests/unit/services/test_default_docs_jwt.py`) is expected.
- The error-result gap (1a) and processor-reuse gap (1b) would both be caught by straightforward unit tests.

#### Code References

- `src/models/url.py:13-99`
- `tests/unit/services/test_default_docs_jwt.py` (covers only the Langflow variant)

#### Potential Solution

- Add a unit test asserting: (a) an `{"status": "error"}` result marks the file task `FAILED` and increments `failed_files`; (b) the temp file is unlinked in the `finally` block; (c) `create_url_upload_task` wires `UrlProcessor` with the expected owner/connector/sample-data fields.

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
- See Normal PR comments for consideration: (1a), (1b)
- See Minor PR comments for optional consideration: (1c), (1d)

## Functional Review 1

- ✅ Successfully verified fix

PR #2318, opened 2026-09-01
fix: Improve handling of no LF image processing

mfortman11:
If 
```
Disabled Langflow Ingestion: True
Picture Descriptions: True
```
Image only pdf upload works as expected

```
Disabled Langflow Ingestion: True
Picture Descriptions: False
```
Upload fails with The file appears corrupted or invalid and cannot be processed. Upload a valid file.

if `Disable Langflow Ingestion` is off it processes the file w/ langflow and returns an empty `<!-- image -->` tag this PR matches this expected behavior

Fix non-Langflow ingestion for image-only documents when picture descriptions are disabled.

Previously, Docling detected the image and produced , but the processor explicitly rejected placeholder-only documents as empty. This change allows the single placeholder to be embedded and indexed, matching Langflow behavior, while genuinely empty documents still fail. It also prevents custom chunk sizes from splitting the placeholder.

Tests cover successful image-only ingestion, exact placeholder indexing, small chunk sizes, and truly empty documents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Documents containing only images can now be processed and indexed successfully.
  * Image placeholders remain intact during document chunking, even with small chunk sizes.
  * Empty documents without images continue to return an appropriate error.
* **Tests**
  * Added coverage for image-only documents and placeholder preservation across supported chunking configurations.

### DIFF
--- a/src/models/processors.py
+++ b/src/models/processors.py
@@ -7,7 +7,6 @@ from typing import TYPE_CHECKING, Any, Literal
 from config.settings import clients, get_embedding_model, get_index_name, get_openrag_config
 from session_manager import AnonymousUser
 from utils.document_processing import (
-    IMAGE_PLACEHOLDER,
     extract_relevant,
     process_text_file,
     resplit_chunks_character_windows,
@@ -531,18 +530,6 @@ class TaskProcessor:
         # This ensures the length of chunks matches the length of the embeddings array,
         # since chunk_texts_for_embeddings also drops empty texts.
         slim_doc["chunks"] = [c for c in slim_doc["chunks"] if c.get("text") and c["text"].strip()]
-
-        # A placeholder preserves picture provenance for otherwise empty documents,
-        # but it is not searchable content and must not turn ingestion into a false
-        # success or consume an embedding request.
-        if slim_doc["chunks"] and all(
-            chunk["text"].strip() == IMAGE_PLACEHOLDER for chunk in slim_doc["chunks"]
-        ):
-            logger.error(
-                "No searchable content extracted — document contains only image placeholders",
-                file_hash=file_hash,
-            )
-            return {"status": "error", "error": "No text content could be extracted from document"}
 
         litellm_embedding_model = (
             await self.models_service.get_litellm_model_name(embedding_model)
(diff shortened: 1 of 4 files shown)


src/models/processors.py
```diff
@@ -532,18 +531,6 @@ async def process_document_standard(
         # since chunk_texts_for_embeddings also drops empty texts.
         slim_doc["chunks"] = [c for c in slim_doc["chunks"] if c.get("text") and c["text"].strip()]
 
-        # A placeholder preserves picture provenance for otherwise empty documents,
```
mpawlow, line 535: ### (1a) [Minor] Literal `<!-- image -->` is embedded and becomes retrievable content

#### Problem

- Removing the guard means every uncaptioned image-only document now contributes exactly one indexed chunk whose `text` is the literal string `<!-- image -->`.
- That chunk is embedded, stored in the `documents` index, and is eligible for hybrid/vector retrieval like any other chunk.
- On retrieval it is wrapped in the untrusted-doc fence and passed to the agent, and `parse_knowledge_chunks` will build a citation whose visible body is `<!-- image -->` — a citation with no informational value for the end user.
- Because the string is identical across all such documents, their placeholder chunks share a near-identical embedding; a query with low lexical/semantic specificity could pull several of them into the top-k, displacing real content.

#### Background Information

- This is a deliberate parity decision: the Langflow path already indexes Docling's `<!-- image -->` markdown, and the PR explicitly aims to match it. So the behavior is defensible and consistent.
- The residual concern is retrieval quality, not correctness. It is strictly better than the `#2308` state for the user who just wants their scan to ingest without an error.
- The removed comment in `processors.py` captured the original rationale ("not searchable content … must not … consume an embedding request"); that trade-off has now been consciously inverted.

#### Code References

- `src/models/processors.py:531-533` (placeholder now flows through the empty-text filter unharmed)
- `src/models/processors.py:554-565` (placeholder embedded)
- `src/utils/document_processing.py:165-178` (placeholder chunk origin)

#### Potential Solution

- Accept as-is for parity, but consider indexing the placeholder chunk with an attribute that keeps it out of retrieval scoring (e.g. index the chunk for provenance/file-listing but mark `type: "picture"` placeholders as non-retrievable, or give them an empty/So-scored embedding), so an image-only file still "exists" in Knowledge without polluting RAG results.
- Alternatively, substitute a slightly more meaningful body (e.g. `"[image]"` or the filename) so a surfaced citation is less confusing — though this then diverges from the Langflow path.

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
- See Minor PR comment (1a) for optional consideration

PR #2319, opened 2026-09-01
chore: worker upgrades for workflows

mfortman11:
Upgrades our actions to use non deprecated node versions and uses valid `setup-uv`

## Summary by CodeRabbit

* **Chores**
  * Updated automation workflows to newer versions of their underlying actions.
  * Refreshed actions used across builds, tests, security scans, deployments, publishing, and release processes.
  * Workflow logic, job structure, and configuration remain unchanged.
  * Deployment workflows now use more restrictive credential persistence settings during checkout.

### DIFF
--- a/.github/workflows/autofix.ci.yml
+++ b/.github/workflows/autofix.ci.yml
@@ -33,12 +33,12 @@ jobs:
 
     steps:
       - name: Checkout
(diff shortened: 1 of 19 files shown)


jason-dong03 (approved):
lgtm

PR #2321, opened 2026-09-02
fix: preserve provider identity across ingestion and retrieval

edwinjosechittilappilly:
## Summary

- Fix generic LiteLLM recovery probes so credentials-based providers preserve actionable deployment and authentication errors.
- Persist exact embedding provider/space metadata for new documents while keeping legacy model fields backward compatible.
- Restore multimodel retrieval across every resolvable indexed embedding space, with paginated discovery, strict provider routing, legacy OpenAI inference for `text-embedding-3-small`, and operator mappings for other legacy models.
- Keep keyword retrieval available when a provider is removed, unconfigured, or a legacy model cannot be mapped.

## How to test

```bash
uv run pytest -q \
  tests/unit/test_provider_error_formatting.py \
  tests/unit/test_task_service_get_task_status2.py \
  tests/unit/services/test_llm_gateway.py \
  tests/unit/test_embedding_fields.py \
  tests/unit/test_document_index_writer_ownership.py \
  tests/unit/test_opensearch_embedding_spaces.py \
  tests/unit/test_search_service_embedding_spaces.py \
  tests/unit/test_openai_compatible_langflow_components.py \
  tests/unit/test_flow_opensearch_outputs.py \
  tests/unit/config/test_legacy_embedding_provider_map.py
```

## Test criteria

- Azure and generic provider credentials are used by recovery probes; missing deployments reach the UI as actionable errors.
- New ingestion stores separate provider-qualified spaces for identical model names; selected provider controls ingestion only.
- Backend and Langflow retrieval query all paginated, configured provider spaces—not only the selected model.
- Legacy `text-embedding-3-small` routes to OpenAI; configured legacy mappings route other models to the specified provider.
- Unknown or removed providers never fall back to the selected provider; vector search is skipped for that space and keyword search still works.

## Summary by CodeRabbit

* **New Features**
  * Added provider-aware embedding indexing and search for Azure and other configured providers.
  * Preserved compatibility with legacy embeddings through configurable provider mappings and keyword-search fallback.
  * Added embedding metadata to ingestion previews, file results, and search responses.
  * Added support for generic provider credentials in provider health and validation checks.
  * Langflow now receives the selected embedding provider automatically.
* **Bug Fixes**
  * Missing Langflow deployments are now identified as user-actionable errors.
* **Documentation**
  * Documented legacy embedding-provider configuration and compatibility behavior.


src/utils/langflow_headers.py
```diff
@@ -131,11 +131,13 @@ def build_model_provider_headers(config, embedding_model: str | None = None) ->
...
         getattr(config, "knowledge", None), "embedding_model", None
     )
+    emb_provider = getattr(getattr(config, "knowledge", None), "embedding_provider", None)
     agent = getattr(config, "agent", None)
     llm_model = getattr(agent, "llm_model", None)
 
     return {
         "X-LANGFLOW-GLOBAL-VAR-SELECTED_EMBEDDING_MODEL": str(emb_model or ""),
+        "X-LANGFLOW-GLOBAL-VAR-SELECTED_EMBEDDING_PROVIDER": str(emb_provider or "openai"),
         "X-LANGFLOW-GLOBAL-VAR-SELECTED_EMBEDDING_MODEL_PROVIDER": "OpenAI",
```
lucaseduoli, line 144: I dont know why this is hardcoded to openai but it should be the same as above, no need for the above one, its the same thing

lucaseduoli (approved):
LGTM otherwise

src/utils/langflow_headers.py
edwinjosechittilappilly: "openai" is kept as a backward-compatible fallback for legacy flows and configurations where no embedding provider was stored.

  For legacy unqualified model names, such as text-embedding-3-small, we resolve the missing provider to openai rather than leaving it empty. This preserves the previous OpenRAG behavior and prevents an existing OpenAI embedding space from being incorrectly routed to an Azure provider.

  This assumption is safe for the legacy OpenRAG model catalog: text-embedding-3-small was previously exposed as an OpenAI model before provider-aware model identity was introduced. New configurations must store and send the explicit provider, including Azure and Watsonx.

PR #2323, opened 2026-09-02
fix: enabling-disabling langflow with model provider

ricofurtado:
Introduce a new setting to disable chat functionality with Langflow, allowing users to run chat against the language model directly. This change includes updates to the settings interface, environment variables, and UI components to support the new feature.

## Summary by CodeRabbit

* **New Features**
  * Added settings and environment options to disable Langflow chat and nudges.
  * Chat, history, and suggestions use OpenRAG services when Langflow is disabled.
  * Improved conversation continuity and sidebar thread handling.
  * Added Azure AI Foundry support across available deployment modes.
  * Improved provider alias handling and custom provider configuration.
  * Improved embedding processing and optional exclusion of sample documents from search results.

* **Bug Fixes**
  * Limited empty nudge polling retries and improved error handling.
  * Added compatibility for newer OpenAI response events.

* **Tests**
  * Expanded coverage across settings, providers, chat, embeddings, nudges, search, and compatibility behavior.


frontend/app/api/queries/useGetNudgesQuery.ts
lucaseduoli: This should be handled by the backend, query until it gets the response. The retry shouldn't happen on the frontend

edwinjosechittilappilly (changes requested):
@coderabbitai instead of disable chat with langflow can we have a generic non langflow mode  in settings that uses  the new litellm /models /completrions and /embeddings end points directly with the token just like langflow./ enev without it internally too.

@coderabbitai can draft a plan with the above thoughts.

PR #2326, opened 2026-09-03
feat: cancel ingestion button for knowledge table

### LINKED ISSUE #1528: [Feature]: Stop ingestion from the table
### Feature Area

### Problem Description

Users should be able to cancel an ongoing ingestion process directly from the knowledge table.

Problem
Once ingestion starts, users cannot interrupt it, even if:

wrong file
wrong embedding model
system issue

Proposed solution

Add “Cancel” action in table for processing items
Reflect state change immediately (e.g. “Cancelled”)

Acceptance criteria

Cancel button visible for in-progress ingestion
Backend ingestion job is stopped
UI reflects cancellation clearly
No partial corrupted state

### Proposed Solution

Users should be able to cancel an ongoing ingestion process directly from the knowledge table.

### Use Case

As a user I should be able to cancel ingesting a file from the table

### Alternatives Considered

_No response_

### Priority

Nice to have

### Additional Context

_No response_

### Contribution

- [ ] I would be willing to help implement this feature.
- [ ] I can help test this feature once implemented.

### Checklist

VisveshJ:
image

image

PR for #1528 

Cancel Button to in-progress rows in the knowledge table. Clicking it stops the ingestion job, updates the row status to Cancelled, and moves the task to Past Tasks in the notification panel.

**Changes**
CancelIngestionButton component renders in the actions column for processing rows
Cancelled rows show "Cancelled" in muted text instead of red "Failed"

**Bug fixes**
Fixed a deadlock in cancel_task() — the previous implementation held _task_lock while awaiting the background task, which itself needed the same lock, causing the task to stay stuck as RUNNING forever
Fixed CancelledError bypassing Langflow chunk cleanup — run_ingestion_flow caught only Exception, leaving partial chunks and ingest tokens orphaned on cancellation
Fixed task panel spinner persisting after cancel — optimistic cache update now transitions the task in-place rather than removing it, so the terminal-state cleanup fires correctly
Added CANCELLED as a distinct task status throughout the backend enum, API types, and frontend status handling

## Summary by CodeRabbit

* **New Features**
  * Added the ability to cancel file ingestion and active tasks.
  * Added cancellation controls with progress feedback while cancellation is processing.
  * Added “Cancelling…” and “Cancelled” indicators in task notifications and knowledge tables.
  * Cancelled tasks and files now display a distinct status and are treated as terminal states.

* **Bug Fixes**
  * Improved cancellation handling to stop background processing reliably and update affected files consistently.
  * Cancelled tasks are now cleaned up alongside other completed task states.


Wallgau (changes requested):
The X currently cancels the whole task, rather than just that specific row. I think it would be better if it canceled only the row, that’s part of the value of having the action in the row table vs in the task panel

src/services/langflow_file_service.py
```diff
@@ -650,7 +650,11 @@ async def run_ingestion_flow(
 
                 raise
             return resp_json
+        except asyncio.CancelledError:
+            # Re-raise cancellation immediately without cleanup to propagate task cancellation
+            raise
```
mfortman11, line 664: Clean up partial ingestion before propagating cancellation — 
When cancellation occurs after Langflow has written one or more callback batches, this branch skips `_cleanup_failed_callback_ingest`, leaving partial chunks searchable and the ingest token valid even though the file is reported as cancelled. Run the cleanup, protected from cancellation if necessary, before re-raising CancelledError.

src/services/task_service.py
```diff
@@ -666,10 +707,26 @@ async def process_with_semaphore(item, item_key: str):
...
+                    self._file_tasks[(task_id, file_task.file_path)] = task
 
             await asyncio.gather(*tasks, return_exceptions=True)
 
+            # Clean up file task references
+            for item in items:
+                item_key = str(item)
+                file_task = upload_task.file_tasks.get(item_key)
+                if file_task:
+                    self._file_tasks.pop((task_id, file_task.file_path), None)
```
mfortman11: Medium - Remove tracked file tasks when the parent is cancelled — 
When a whole upload task is cancelled while awaiting `gather`, control jumps directly to the outer `CancelledError` handler, bypassing this cleanup loop. Consequently `_file_tasks` permanently retains every child task for that upload in the singleton service, causing memory to grow with each cancelled task; move this cleanup into a `finally` block or use done callbacks.

frontend/contexts/task-context.tsx
```diff
@@ -699,6 +763,13 @@ export function TaskProvider({ children }: { children: React.ReactNode }) {
     [cancelTaskMutation],
   );
 
+  const cancelFile = useCallback(
+    async (taskId: string, filePath: string) => {
+      cancelFileMutation.mutate({ taskId, filePath });
+    },
+    [cancelFileMutation],
+  );
```
mfortman11, line 799: Medium - Await cancellation mutations before resolving —
When the cancellation request is slow or fails, this async wrapper resolves immediately because `mutate` returns void. The file button therefore re-enables before the request completes, permitting duplicate
requests, and the newly added task-menu `try/catch` can never clear its “Cancelling...” state on failure. Use `mutateAsync` so callers receive the actual request promise.

mfortman11 (changes requested):

frontend/lib/knowledge-table-state.ts
mfortman11, line 1: P1: `frontend/lib/knowledge-table-state.ts:224` discards the task’s actual file key when merging re-ingestion overlays. `frontend/app/knowledge/page.tsx:267` may consequently hide cancellation or submit the wrong key and receive a 400.

mfortman11 (changes requested):

src/services/task_service.py
```diff
@@ -1514,41 +1576,122 @@ async def cancel_task(self, user_id: str, task_id: str) -> bool:
...
+
+        # Lock to prevent race conditions
+        async with self._get_task_lock(task_id):
+            # Can only cancel pending or running files
+            if file_task.status not in [TaskStatus.PENDING, TaskStatus.RUNNING]:
+                return False
+
+            # Mark file as failed (cancelled)
+            was_running = file_task.status == TaskStatus.RUNNING
+            file_task.status = TaskStatus.FAILED
```
Wallgau: Depends on the case: if this is one file inside a folder / multi-file task, we should **not** flip the parent to `CANCELLED`. Sibling files may still be running or already completed, so the task is at most partially cancelled — which we do not currently model.

Leaving the parent as running/completed and this file as failed-with-a-cancellation-error is the right split for now. A dedicated file `CANCELLED` status (or a parent `partially_cancelled`) would collapse the magic-string checks, but it is a product decision, not a must-fix on this PR.

Wallgau (commented):
The functional has been tested and works fine, please address @mfortman11 comments

mfortman11 (approved):

Wallgau (approved):
Great work!

PR #2327, opened 2026-09-03
fix: resolve OPENSEARCH_INDEX_NAME consistently across config loaders

mfortman11:
### Problem

Connector ingests intermittently fail with `NotFoundError(404, 'index_not_found_exception', 'no such index [documents]')`, marking the file as FAILED in the task panel. Reloading the page and re-syncing makes the same file succeed.

Root cause: `OPENSEARCH_INDEX_NAME` is only applied by the file-based config loader (ConfigManager._load_env_overrides). In db/hybrid storage mode (the SaaS default), config is built straight from the workspace DB by `WorkspaceConfigService.load_config()`, which never applies env overrides — so knowledge.index_name silently falls back to the "documents" default whenever the stored config lacks the key or is marked edited. Additionally, even in file mode the override was applied after the edited-flag gate, so an onboarded workspace lost it.

The result is a split: the Langflow ingest path writes chunks to the configured index (e.g. `orag-documents`) while the connector post-ingest enrichment (`_update_connector_metadata` → `update_by_query`) resolves documents, which was never `created. _update_connector_metadata` then re-raises the 404, failing the file. A page reload triggers `_ensure_index_exists()` / a consistent config reload, so the retry lines up on one index name.

### Fix

- Apply `OPENSEARCH_INDEX_NAME` as an unconditional infra override in both config loaders, before the edited gate, so every storage mode / process / cache state resolves the same index name.
- `_update_connector_metadata` now treats `index_not_found_exception` as non-fatal (warn + skip enrichment) like the sibling connector paths already do — the chunks are already indexed and enrichment re-runs on the next sync.

### How to Reproduce

Repro on main:

Part 1: (see the state issue)

1. `git checkout main`
2. .env:
```
OPENSEARCH_INDEX_NAME=orag-documents
OPENRAG_STORAGE_MODE=hybrid
```
3. `make factory-reset`
4. Create `config/config.yaml` with just:
```
edited: true
knowledge:
  index_name: documents
```
5. `make dev-local` + start the host backend.
6. Complete onboarding, then sync any connector
7. Observe the bug:
```
curl -s localhost:8000/api/settings | jq '.knowledge.index_name'   # => "documents"
curl -sku admin:admin https://localhost:9200/_cat/indices | grep documents #   => a `documents` index exists; no `orag-documents`
```
   The `OPENSEARCH_INDEX_NAME=orag-documents` setting was ignored.
   
Part 2: (functional reproduction)

Add this shim
```
 src/config/settings.py
def get_index_name() -> str:
    """Return the currently configured index name."""
    import inspect  # TEST AID — simulate the SaaS cross-process index-name split
    for _f in inspect.stack():
        if _f.function in ("_update_connector_metadata", "should_update_acl", "update_document_acl"):
            return "documents"
    return get_openrag_config().knowledge.index_name
```

1. Apply the patch above.
2. `.env`: `OPENSEARCH_INDEX_NAME=orag-documents` (storage mode can be default db).
3. `make factory-reset`, `make dev-local`, onboard.
4. Delete both indices so the enrichment target truly doesn't exist:
```
curl -sku admin:admin -XDELETE https://localhost:9200/orag-documents
curl -sku admin:admin -XDELETE https://localhost:9200/documents
```
5. Sync a connector → file shows FAILED, `NotFoundError(404, 'index_not_found_exception', 'no such index [documents]')`. Refreshing the Documents page + re-sync recovers it.

### Verify the fix

**Part 1:**

1. git checkout this branch and leave `.env` and `config/config.yaml` as above.
2. `make factory-reset`, `make dev-local`, restart backend, onboard, sync the same connector.
3. Observe the fix:
```
curl -s localhost:8000/api/settings | jq '.knowledge.index_name'   # => "orag-documents"
curl -sku admin:admin https://localhost:9200/_cat/indices #   => `orag-documents` exists; NO `documents` index
```
   Backend log shows: Stored index_name 'documents' overridden by OPENSEARCH_INDEX_NAME='orag-documents'
4. File syncs successfully.

**Part 2:**

1. Checkout this branch, keep the patch, same `.env`, `make factory-reset`, onboard, delete both indices, sync.
2. File completes. Backend log shows: Skipping connector metadata enrichment — index not found.
   (The resolution fix can't help here — the patch hard-codes the mismatch — so this proves the hardening: a missing index no longer fails the file.)
3. Remove the patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the configured OpenSearch index name consistently honors the permitted environment setting, including edited and database-backed configurations.
  * Prevented connector enrichment from failing when the target OpenSearch index is missing; a warning is recorded instead.
  * Continued to surface other OpenSearch errors normally.
  * Applied the configured index name consistently when loading settings from empty or unavailable database configurations.


src/connectors/service.py
```diff
@@ -180,6 +182,23 @@ async def _update_connector_metadata(
...
+            # them (e.g. a residual index-name mismatch, issue 81583). The
+            # document is already indexed; metadata enrichment is best-effort
+            # and re-runs on the next sync, so don't fail the file over it —
+            # matching get_synced_file_ids_for_connector / should_update_acl.
+            if (
+                isinstance(e, NotFoundError)
+                and e.status_code == 404
+                and e.error == "index_not_found_exception"
+            ):
+                logger.warning(
```
edwinjosechittilappilly, line 195: @mfortman11  should we ensure this on the operator side as a check in an init container.
@zzzming

edwinjosechittilappilly:
@coderabbitai thoughts on wasnt the index name read from the env at start up earlier and what caused this issue?

edwinjosechittilappilly:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== PR diff ==='
> gh pr diff 2327 -- src/config/config_manager.py src/config/settings.py src/services/workspace_config_service.py
> 
> echo '=== Relevant symbols ==='
> ast-grep outline src/config/config_manager.py --items all --match 'apply_index_name_env_override|_load_env_overrides'
> ast-grep outline src/config/settings.py --items all --match 'get_opensearch_index_name_override'
> ast-grep outline src/services/workspace_config_service.py --items all --match 'load_config|_config_from_db_dict'
> 
> echo '=== Configuration load and override call sites ==='
> rg -n -C 5 --glob '*.py' \
>   'apply_index_name_env_override|get_opensearch_index_name_override|_load_env_overrides|WorkspaceConfigService|ConfigManager' \
>   src tests/unit/config/test_config_manager_env_overrides.py tests/unit/services/test_workspace_config_service.py
> 
> echo '=== Relevant test cases ==='
> sed -n '1,220p' tests/unit/config/test_config_manager_env_overrides.py
> sed -n '1,280p' tests/unit/services/test_workspace_config_service.py
> ```
> 
> Length of output: 1391
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== PR diff ==='
> gh pr diff 2327
> 
> echo '=== Configuration manager map ==='
> ast-grep outline src/config/config_manager.py --items all --match 'apply_index_name_env_override|_load_env_overrides'
> echo '=== Settings map ==='
> ast-grep outline src/config/settings.py --items all --match 'get_opensearch_index_name_override'
> echo '=== Workspace config service map ==='
> ast-grep outline src/services/workspace_config_service.py --items all --match 'load_config|_config_from_db_dict'
> 
> echo '=== Loader and override implementations/call sites ==='
> rg -n -C 6 --glob '*.py' \
>   'apply_index_name_env_override|get_opensearch_index_name_override|def _load_env_overrides|def load_config|def _config_from_db_dict' \
>   src tests/unit/config/test_config_manager_env_overrides.py tests/unit/services/test_workspace_config_service.py
> 
> echo '=== Focused regression tests ==='
> sed -n '1,180p' tests/unit/config/test_config_manager_env_overrides.py
> rg -n -C 8 --glob '*.py' \
>   'OPENSEARCH_INDEX_NAME|test_.*db.*index|test_.*index.*db' \
>   tests/unit/services/test_workspace_config_service.py
> ```
> 
> Length of output: 41327
> 
> `@edwinjosechittilappilly` Yes, but only on one configuration path.
> 
> Before this PR, `ConfigManager._load_env_overrides` read `OPENSEARCH_INDEX_NAME` during YAML-backed configuration loading. It did not apply the value when `edited` was `true`, because the method returned before it reached the index-name override.
> 
> The DB-backed path had a separate problem. `WorkspaceConfigService.load_config` built `OpenRAGConfig` directly from DB rows. It did not call `_load_env_overrides` at all. In DB mode, or DB-first hybrid mode, `index_name` therefore came from the DB value or the default `documents`.
> 
> This caused inconsistent resolution after startup:
> 
> 1. An ingest path could use the environment-configured index.
> 2. A connector metadata-enrichment path could reload configuration through the DB path.
> 3. That path could resolve `documents` instead.
> 4. OpenSearch then returned `index_not_found_exception` when enrichment targeted the wrong index.
> 
> This PR makes `OPENSEARCH_INDEX_NAME` an unconditional infrastructure override in both loaders. It also applies the override before the `edited` check.
> 
> _You are interacting with an AI system._

@coderabbitai but isnt suppose to be same values read from env how did the user got different values for index name

edwinjosechittilappilly:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== Configuration-related PR diff ==='
> gh pr diff 2327 -- \
>   src/config/config_manager.py \
>   src/config/settings.py \
>   src/services/workspace_config_service.py \
>   src/connectors/service.py
> 
> echo '=== Configuration and connector symbols ==='
> ast-grep outline src/config/config_manager.py --items all --match 'apply_index_name_env_override|_load_env_overrides|load'
> ast-grep outline src/services/workspace_config_service.py --items all --match 'load_config|_config_from_db_dict'
> ast-grep outline src/connectors/service.py --items all --match '_update_connector_metadata'
> 
> echo '=== Index-name reads and configuration loader call sites ==='
> rg -n -C 6 --glob '*.py' \
>   'OPENSEARCH_INDEX_NAME|get_opensearch_index_name_override|apply_index_name_env_override|_load_env_overrides|WorkspaceConfigService|ConfigManager|index_name' \
>   src
> 
> echo '=== Base-version configuration-related diff context ==='
> git diff HEAD^ HEAD -- \
>   src/config/config_manager.py \
>   src/config/settings.py \
>   src/services/workspace_config_service.py \
>   src/connectors/service.py
> ```
> 
> Length of output: 1477
> 
> 🏁 Script executed:
> 
> ```shell
> #!/bin/bash
> set -euo pipefail
> 
> echo '=== Configuration-related PR diff ==='
> git diff HEAD^ HEAD -- \
>   src/config/config_manager.py \
>   src/config/settings.py \
>   src/services/workspace_config_service.py \
>   src/connectors/service.py
> 
> echo '=== Configuration and connector symbols ==='
> ast-grep outline src/config/config_manager.py --items all --match 'apply_index_name_env_override|_load_env_overrides|load' || true
> ast-grep outline src/services/workspace_config_service.py --items all --match 'load_config|_config_from_db_dict' || true
> ast-grep outline src/connectors/service.py --items all --match '_update_connector_metadata' || true
> 
> echo '=== Index-name reads and configuration loader call sites ==='
> rg -n -C 6 --glob '*.py' \
>   'OPENSEARCH_INDEX_NAME|get_opensearch_index_name_override|apply_index_name_env_override|_load_env_overrides|WorkspaceConfigService|ConfigManager|index_name' \
>   src || true
> 
> echo '=== Base versions of changed loader code ==='
> git show HEAD^:src/config/config_manager.py | rg -n -C 6 \
>   'def _load_env_overrides|OPENSEARCH_INDEX_NAME|index_name' || true
> git show HEAD^:src/services/workspace_config_service.py | rg -n -C 6 \
>   'def load_config|OpenRAGConfig\.from_dict|index_name' || true
> ```
> 
> Length of output: 205598
> 
> `@edwinjosechittilappilly` You are correct. The intended result is one index name from `OPENSEARCH_INDEX_NAME`.
> 
> My prior reply was imprecise. There is no evidence that the user had two different environment values.
> 
> The pre-fix problem was that not every configuration load read the environment override:
> 
> * `get_index_name()` reads `get_openrag_config().knowledge.index_name`.
> * In `files` mode, `ConfigManager._load_env_overrides` could apply `OPENSEARCH_INDEX_NAME`.
> * In `db` mode, `WorkspaceConfigService.load_config()` built `OpenRAGConfig` from the persisted DB data. It did not run `_load_env_overrides`.
> * Therefore, the DB path could return the stored `index_name`, or the default `documents` when the DB value was absent, even when `OPENSEARCH_INDEX_NAME` was set.
> 
> For example, a deployment could set:
> 
> ```
> OPENSEARCH_INDEX_NAME=customer-documents
> ```
> 
> while an existing DB row still contains:
> 
> ```
> knowledge.index_name=documents
> ```
> 
> Before this PR, the DB-backed configuration object used `documents`. The environment value was not consulted on that path.
> 
> This PR makes the environment override run after either loader creates the `knowledge` data. Both paths now resolve `knowledge.index_name` to the same permitted `OPENSEARCH_INDEX_NAME` value. 🐇
> 
> _You are interacting with an AI system._

@mfortman11  @coderabbitai can you create an issue to investigate why the DB was fall back to a value that was not from env.

edwinjosechittilappilly (approved):
LGTM.

But Also have addeda thread on root cause what made this inconsistency in value in DB and in env.
@rodageve @zzzming  I guess we might need to check logs in deployment.

Wallgau (approved):
LGTM

PR #2341, opened 2026-09-08
chore: Setup UI testing framework

mfortman11:
### Testing stack

| Layer | Tool | Why |
|---|---|---|
| Runner | **Vitest 4** | Already used in `sdks/typescript`; fast transform + watch |
| Environment | **jsdom** | Node has no DOM, so React can't mount without it |
| React | **Testing Library** (`react`, `user-event`, `jest-dom`) | Vitest has no concept of React. `user-event` fires full pointer sequences, which Radix requires |
| Network | **MSW 2** | Intercepts *below* the app, so real `fetch` and real react-query run unmodified |

```bash
npm test 
npm run test:watch
npm run test:coverage
npm run test:diff-coverage
```

### Where tests go

**The file extension names the runner.**

| Pattern | Runner | Needs a backend? |
|---|---|---|
| `**/*.test.ts(x)` — co-located with source | Vitest | No |
| `tests/**/*.spec.ts` | Playwright | Yes, full Docker stack |
| `test-utils/` | shared Vitest infra | — |

### Rules

1. **Query by accessible role/label/text.** No class-name assertions, no test-only ids — tests should fail on behaviour changes, not markup refactors.
2. **Never `vi.mock` anything in `app/api/queries/` or `app/api/mutations/`.** Mock the network with MSW instead, so the URL, `response.ok` branch, payload shape and query wiring stay under test. Mocking a *context* is fine.
3. **Use `renderWithProviders`**, not `app/providers.tsx` — the latter monkey-patches `window.fetch`, and `get-query-client.ts` memoizes one client that would leak cache between tests.
4. **Type handler bodies** against the real response interfaces. This caught two fixture mismatches that every test passed straight through.
5. **No global coverage threshold.** At ~10% a floor rewards writing the cheapest tests. New code is gated per-PR instead; old gaps close by churn priority — target what git history shows is repeatedly *fixed*, not what looks complex.

### What does not belong in Vitest

jsdom has no CSS engine and no layout. Three boundaries, all documented inline where they bite:

- **CSS-enforced behaviour** — `disabled:pointer-events-none` never applies, so Radix opens a "disabled" menu. Assert the attribute, never the consequence.
- **Uploaded file names/content** — jsdom's `File` crosses into undici as an unnamed, empty blob. Text fields round-trip; file parts can only be counted.
- **Mid-stream abort** — the reader's pending `read()` never settles, so the test hangs to the suite timeout.

All three belong in Playwright. This is why the E2E suite stays.

### CI

New `test-frontend-unit.yml`, running alongside the unchanged Playwright job:

- **`vitest`** — on `frontend/**` changes. No Docker, no services.
- **`diff-coverage`** — PRs only, **80% of changed lines**. Gates changed *lines*, not changed *files*, so a one-line fix to untested legacy code isn't blocked, but new logic must carry tests.

## Summary by CodeRabbit

- **Tests**
  - Added comprehensive automated coverage for frontend search, tasks, chat streaming, uploads, onboarding, filtering, error handling, and knowledge-table behavior.
  - Added reusable component and hook testing utilities.
  - Standardized frontend tests on Vitest.

- **Chores**
  - Added automated frontend test execution for pull requests and main-branch updates.
  - Added an 80% changed-line coverage gate.

- **Documentation**
  - Documented frontend testing conventions, commands, coverage expectations, and test-selection guidance.


Wallgau (approved):
Nice work!
I think coderabbitai[bot] are accurate and needs to be addressed but approving ahead!

PR #2349, opened 2026-09-09
fix: Address 409 error for bucket connectors

mfortman11:
## Fix truncated OpenSearch errors on failed ingestion

An OpenSearch transport failure during ingestion showed up in the task view as twenty characters — `ConflictError(409, '` — and nothing else. Two independent defects stacked up to produce that.

### The backend never classified OpenSearch errors

`_infer_failure_metadata` has branches for Docling, Langflow, credentials and connectivity, but nothing for `opensearch-py` transport exceptions, so it returned `None` and the UI fell back to the raw exception string. Adds `_opensearch_failure_metadata`, which emits `component: "opensearch"` / `failure_phase: "indexing"` and a real sentence. Both values were already supported by the frontend, so no display types changed.

It's called *before* the substring heuristics, which also fixes a pre-existing misclassification: a 400 `mapper_parsing_exception` body contains the text `"failed to parse"`, so it was being reported to users as a corrupted file.

### The frontend parser was discarding the diagnosis

`formatProviderErrorMessage` is written for chat provider errors, where an embedded JSON blob is noise and the text before it is the message. For `ConflictError(409, '{...}')` that assumption inverts: the prefix is a syntactic fragment and the JSON *was* the entire diagnosis — the `failures[]` array naming the index, shard and document. It now keeps the whole string when the prefix ends on an unclosed delimiter, and still prefers a genuine prefix like `Invalid API key {...`.

### Behaviour is unchanged

A file that failed still fails. The underlying 409 is a version conflict between concurrent by-query operations — a real race, tracked separately. This PR only makes the failure legible and correctly marked `RETRYABLE`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved task failure handling for search-indexing errors, identifying retryable issues and providing clearer guidance.
  - Prevented infrastructure response details and raw JSON from appearing in user-facing error messages.
  - Improved error-message formatting when provider responses are incomplete or malformed, preserving readable text while avoiding misleading fragments.

### DIFF
--- a/frontend/lib/chat-stream-errors.test.ts
+++ b/frontend/lib/chat-stream-errors.test.ts
@@ -172,4 +172,20 @@ describe("formatProviderErrorMessage", () => {
     );
     assert.equal(formatProviderErrorMessage("Error: boom"), "Error: boom");
   });
+
+  it("keeps the whole string when the prefix is only a syntactic fragment", () => {
+    const body =
+      '{"took":2719,"timed_out":false,"total":300,"updated":19,' +
+      '"version_conflicts":1,"failures":[{"index":"documents","id":"20",' +
+      '"cause":{"type":"version_conflict_engine_exception"}}]}';
+    const raw = `ConflictError(409, '${body}')`;
+    assert.equal(formatProviderErrorMessage(raw), raw);
+  });
+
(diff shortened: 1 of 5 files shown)


edwinjosechittilappilly (approved):
code LGTM

PR #2351, opened 2026-09-10
feat: Personalized UI Experience

jason-dong03:
-conversation "+" button changes: updated to show temporary conversation button 
-display name shown in chat and chat context scenarios (basic/ initial commit) 
-settings > General > Display name change option
-onboarding includes a step to add name (optional)

## Summary by CodeRabbit

- **New Features**
  - Added optional display-name personalization during onboarding and a General settings page for updates.
  - Chat greetings now personalize to users and vary by time and day.
  - Added typewriter animations for greetings and new conversation titles.
  - Added validated file dropping with clearer errors and drag-and-drop feedback.

- **Bug Fixes**
  - Prevented unnecessary placeholder conversations during chat initialization.
  - Improved success and error toast styling.
  - Preserved display names in no-auth mode.

- **Changes**
  - Chat input file selection now uses the file picker; direct dropping there is no longer processed.

### DIFF
--- a/frontend/app/chat/page.tsx
+++ b/frontend/app/chat/page.tsx
@@ -23,6 +23,7 @@ import {
 import { FILE_CONFIRMATION, FILES_REGEX } from "@/lib/constants";
 import { buildSearchPayloadFilters } from "@/lib/filter-normalization";
 import { uploadFileForContext } from "@/lib/upload-utils";
+import { resolveDisplayName } from "@/lib/user";
 import { cn } from "@/lib/utils";
 import { useGetConversationsQuery } from "../api/queries/useGetConversationsQuery";
 import { useGetNudgesQuery } from "../api/queries/useGetNudgesQuery";
@@ -65,9 +66,7 @@ function ChatPage() {
     setChatError,
   } = useChat();
   const { user } = useAuth();
-  const displayName =
-    user?.display_name ||
-    (user?.name && user.name !== "Anonymous User" ? user.name : null);
+  const displayName = resolveDisplayName(user);
   const [messages, setMessages] = useState<Message[]>(() => [
     makeInitialMessage(displayName),
   ]);
@@ -393,7 +392,7 @@ function ChatPage() {
       window.removeEventListener("newConversation", handleNewConversation);
       window.removeEventListener("focusInput", handleFocusInput);
     };
-  }, [abortStream, setLoading]);
+  }, [abortStream, setLoading, displayName]);
 
   // Load conversation data from context
   useEffect(() => {
--- a/frontend/app/chat/page.tsx
+++ b/frontend/app/chat/page.tsx
@@ -1,14 +1,18 @@
 "use client";
 
-import { Loader2, Zap } from "lucide-react";
-import { useEffect, useMemo, useRef, useState } from "react";
+import { Loader2, Upload, Zap } from "lucide-react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { toast } from "sonner";
 import { StickToBottom, useStickToBottomContext } from "use-stick-to-bottom";
 import { ProtectedRoute } from "@/components/protected-route";
 import { Button } from "@/components/ui/button";
+import { useAuth } from "@/contexts/auth-context";
 import { useIsCloudBrand } from "@/contexts/brand-context";
 import { type EndpointType, useChat } from "@/contexts/chat-context";
 import { useTask } from "@/contexts/task-context";
+import { useFileDrag } from "@/hooks/use-file-drag";
 import { useOnboardingState } from "@/hooks/use-onboarding-state";
+import { useSupportedFileTypes } from "@/hooks/use-supported-file-types";
 import { useChatStreaming } from "@/hooks/useChatStreaming";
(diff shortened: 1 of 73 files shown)


src/api/users.py
```diff
@@ -164,3 +164,51 @@ async def set_my_dev_role(
             "permissions": sorted(perms),
         }
     )
+
+
+class DisplayNameBody(BaseModel):
+    display_name: str | None = None
+
+
+@router.patch("/me/display-name")
```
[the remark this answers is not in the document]
jason-dong03: not a public facing api, internal endpoint for users; properly scoped becuse its JWT not API key auth

Wallgau (changes requested):
The UI for custom display names is gone (no onboarding step, no Settings tab), but PATCH /users/me/display-name is still a live write API. It persists to users.display_name or workspace meta.no_auth_display_name. /auth/me still reads that value in _get_display_name, and the greeting still prefers display_name over the OAuth name. A client can change the greeting with no product surface, and no-auth still does a wasted meta lookup for a field nothing sets.

Suggested fix: Remove this route (DisplayNameBody / DisplayNameResponse / update_my_display_name) and UserRepo.update_display_name. Stop returning display_name from /auth/me (no-auth can return user: None). Keep merge_section_keys for the edited flag only. Resolve the greeting from OAuth name in resolveDisplayName, do not prefer a leftover DB display_name. Delete tests/unit/api/test_display_name_endpoint.py with the endpoint.

frontend/app/chat/page.tsx
```diff
@@ -58,12 +63,32 @@ function ChatPage() {
...
+  // so the lazy initialiser above runs before user is populated).
+  // Only replace if the chat is still at the single greeting message — never
+  // clobber an active conversation.
+  const prevDisplayNameRef = useRef<string | null | undefined>(undefined);
+  useEffect(() => {
+    if (prevDisplayNameRef.current === displayName) return;
+    prevDisplayNameRef.current = displayName;
+    setMessages((prev) => {
+      if (prev.length === 1 && prev[0].role === "assistant") {
+        return [makeInitialMessage(displayName)];
```
Wallgau, line 85: length === 1 && role === "assistant" treats any one-message assistant thread as the empty-state greeting. When auth resolves, a loaded historical reply can be replaced by makeInitialMessage(displayName).

Suggested fix: Add isGreeting: true on messages created by makeInitialMessage. Only replace when that flag is set:

if (prev.length === 1 && prev[0].isGreeting) {
  return [makeInitialMessage(displayName)];
}
Do not use role === "assistant" as the condition. Real replies are also assistant messages; that is what clobbers history. isGreeting is the only signal that this row is the generated empty-state greeting.

frontend/app/chat/page.tsx
```diff
@@ -1040,10 +1122,7 @@ function ChatPage() {
                             animate={false}
                             isInactive={index < messages.length - 1}
                             isInitialGreeting={
-                              index === 0 &&
-                              messages.length === 1 &&
-                              message.content ===
-                                INITIAL_ASSISTANT_MESSAGE.content
+                              index === 0 && messages.length === 1
```
Wallgau: Same bug on render: a stored single assistant message gets the typewriter treatment and loses feedback.
suggested fix:
`isInitialGreeting={!!message.isGreeting}`

frontend/lib/greeting.ts
```diff
@@ -0,0 +1,102 @@
...
+    ? rawFirst.charAt(0).toUpperCase() + rawFirst.slice(1)
+    : "";
+  const nameInsertion = firstName ? `, ${firstName}` : "";
+
+  const now = new Date();
+  const day = now.getDay();
+  const hour = now.getHours();
+
+  const dayVariants = DAY_OF_WEEK_MESSAGES[day];
+  if (dayVariants && Math.random() < 0.35) {
```
Wallgau, line 110: ChatPage is a client component, but Next still SSR's the first paint. useState(() => makeInitialMessage(displayName)) runs on the server and again in the browser. This function is nondeterministic: line 95 uses Math.random() to choose day-of-week vs time-of-day, and pick() uses Math.random() again for the variant. new Date() at line 90 can also differ by timezone. Server HTML and the hydrated tree can be different strings (Good morning, Jason vs Happy Tuesday, Jason). That is a hydration mismatch, not just a visual flicker.

Suggested fix: Keep getGreetingMessage() as-is. Do not call it in the useState initializer in frontend/app/chat/page.tsx. Paint a stable placeholder on first render, then generate the real greeting after mount. Reuse the existing displayName effect and gate the swap with isGreeting:
`const PLACEHOLDER_GREETING: Message = {
  role: "assistant",
  content: "How can I assist you today?",
  timestamp: new Date(0),
  isGreeting: true,
};

const [messages, setMessages] = useState<Message[]>([PLACEHOLDER_GREETING]);

useEffect(() => {
  setMessages((prev) => {
    if prev.length === 1 && prev[0].isGreeting) {
      return [makeInitialMessage(displayName)];
    }
    return prev;
  });
}, [displayName]);`

Wallgau (changes requested):

Wallgau (approved):
Great work!

PR #2356, opened 2026-09-11
fix: retry transient OpenSearch bulk failures instead of failing the …

### LINKED ISSUE #1234: [Bug]: Task list number of success and failed job seems inconsistent
### OpenRAG Version

0.4.0.dev7

### Deployment Method

uvx (uvx openrag)

### Operating System

Mac os

### Python Version

_No response_

### Affected Area

### Bug Description

Image

The Two files ingested but tasks says as failed

### Steps to Reproduce

ingest file and look at tasks.

### Expected Behavior

2 files inhgested hence the success should be 2.

### Actual Behavior

Na

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### Checklist

Thenathanb:
…whole file

A single bulk() call carries every chunk of a document, and OpenSearch sets errors: true on the response if even one item in that request failed - even when the rest of the chunks were written successfully. That RuntimeError propagates up and marks the entire file as failed (upload_task.failed_files += 1), even though the file is actually indexed and searchable. This matches issue #1234's symptom: files that show up correctly in Knowledge (built from OpenSearch aggregations, independent of the task counters) are still counted as failed in the task list.

Since every bulk op is an idempotent upsert, retrying the identical request is safe. Add DocumentIndexWriter._bulk_with_retry, which retries the whole bulk() call (up to 2 times, with backoff) only when every failing item's error type looks transient (rejected_execution, cluster event timeout, unavailable shards, timeout, version conflict). A genuine content/schema error (e.g. mapper_parsing_exception) is not retried and still fails fast.

Fixes #1234

## Summary by CodeRabbit

- **Bug Fixes**
  - Bulk document indexing retries only failed items, preserving successful results from earlier attempts.
  - Temporary failures can recover without resending the entire batch, reducing duplicate processing and improving reliability under load.
  - Non-retryable errors are reported promptly, while repeated temporary failures stop after a limited number of attempts.
  - Incomplete bulk responses and reported errors are now surfaced accurately.

- **Tests**
  - Expanded coverage for partial retries, recovery, error handling, and persistent failures.

### DIFF
--- a/src/services/document_index_writer.py
+++ b/src/services/document_index_writer.py
@@ -7,6 +7,7 @@ indexing chunks into the documents index.
 
 from __future__ import annotations
 
+import asyncio
 import datetime
 import hashlib
 import json
@@ -19,6 +20,25 @@ from utils.group_acl import unique_acl_principal_labels, unique_acl_principals
 from utils.logging_config import get_logger
 
 logger = get_logger(__name__)
+
+# A single bulk() call executes every item independently, so one item hitting
+# a transient condition (cluster under load, a shard temporarily unavailable)
+# sets `errors: true` for the whole response even though the rest of the
+# document's chunks were written successfully. That makes an entire file look
+# "failed" in the task counters even when it's actually indexed and visible.
+# Retrying the whole request is safe here because every op is an idempotent
+# upsert, so re-sending already-succeeded items has no effect.
+_BULK_RETRYABLE_ERROR_TYPES = frozenset(
+    {
+        "es_rejected_execution_exception",
+        "process_cluster_event_timeout_exception",
+        "unavailable_shards_exception",
+        "timeout_exception",
+        "version_conflict_engine_exception",
+    }
+)
+_MAX_BULK_RETRIES = 2
+_BULK_RETRY_DELAY_SECONDS = 0.5
 
 
 @dataclass
@@ -134,7 +154,7 @@ class DocumentIndexWriter:
                 )
             )
 
-        result = await client.bulk(body=bulk_body, refresh=refresh)
+        result = await self._bulk_with_retry(client, bulk_body, refresh=refresh)
         self._raise_for_bulk_errors(result)
         if final:
             await self._refresh(index_name)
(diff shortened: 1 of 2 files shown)


src/services/document_index_writer.py
```diff
@@ -315,6 +335,34 @@ def _normalized_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
...
+            if not isinstance(result, dict) or not result.get("errors"):
+                return result
+            if not self._bulk_errors_are_retryable(result):
+                return result
+            logger.warning(
+                "Retrying OpenSearch bulk write after transient error",
+                attempt=attempt + 1,
+            )
+            await asyncio.sleep(_BULK_RETRY_DELAY_SECONDS * (attempt + 1))
+            result = await client.bulk(body=bulk_body, refresh=refresh)
```
mfortman11: ### [Medium] Retry only the failed bulk items

Resending the entire bulk request discards cumulative progress. Under load, one attempt can reject chunk B after A succeeds, while the next rejects A after B succeeds. After the retry limit, the file is still marked failed even though every chunk was written successfully. It also requeues successful items during overload.

Rebuild each retry body using only the failed action/document pairs.

mfortman11 (changes requested):

src/services/document_index_writer.py
```diff
@@ -315,6 +337,85 @@ def _normalized_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
...
+
+            items = result.get("items", [])
+
+            # Guard against a bad or cut-short response — the server should
+            # return exactly one item for each request we sent. If the list
+            # is short or empty we can't safely match results back to the
+            # right slots, so we treat it as an error rather than leaving
+            # gaps filled with None.
+            if len(items) != len(pending_indices):
+                return {**result, "items": final_items, "errors": True}
```
mfortman11, line 388: Avoid passing null placeholders to the bulk error parser

When a response omits or mismatches `items`, the returned list contains `None` values. `_raise_for_bulk_errors` immediately calls `.get()` on each item, causing an unintended `AttributeError`. This currently breaks four tests in `tests/unit/test_document_index_writer_ownership.py`, whose successful fakes return `{"errors": false}`.

Either update those fakes to satisfy the new cardinality contract and raise directly for malformed responses, or return error items that `_raise_for_bulk_errors` can safely consume.

mfortman11 (changes requested):

mfortman11 (approved):

PR #2365, opened 2026-09-13
fix: file overwrite issue in COS sync

ricofurtado:
Resolve the problem where replacing a file in the COS was not reflected after syncing connectors. This update ensures that file changes are accurately synchronized.


edwinjosechittilappilly:
Please check the Base Branch.

ricofurtado:
Closing in favour of #2366, which targets `main`.

@edwinjosechittilappilly you were right about the base branch — this PR was cut from `b8dd86ba`, which is not an ancestor of `main`. #2366 is the same fix branched from `origin/main`, so please review there.

Two things worth noting for anyone following the discussion here:

- CodeRabbit skipped this PR ("auto reviews are disabled on base/target branches other than the default branch"), and React Doctor's "153 new issues in 58 files" was diffed against `0.1.19`. Both were artefacts of the wrong base; #2366 gets a real review.
- The rebase surfaced two things `main` requires that this base did not: the frontend test had to be ported from `node:test` to vitest, and `main`'s diff-coverage gate (80% of changed lines) failed at 65.5% until a component test was added for the sync dialog's new states. #2366 is at 100% (29/29).

No code review feedback from here is left unaddressed.

PR #2366, opened 2026-09-14
fix: COS file overwrite not reflected after syncing connectors

ricofurtado:
## The defect

Ingest a file from COS, overwrite it at the source, then Sync all connectors: the Knowledge copy stays stale. Only deletions were being propagated.

There were two independent causes, either of which produces exactly that symptom.

### 1. Sync silently treated shared COS files as private

`shared` (the COS-only "make documents available to all users" toggle) indexes chunks with **no `owner` field**, and nothing persists that choice — it only ever existed as a per-request flag from the ingest UI. The re-sync paths therefore ran as `shared=False`, and every owner-scoped step downstream missed those ownerless chunks:

- the replace-delete used an owner-only query, matched nothing, and returned `0`;
- `resolve_duplicate_filename` read that `0` as "nothing to replace" and **skipped the file**, leaving the stale copy indexed;
- `_reconcile_shared_owner` then stamped the syncing user onto those chunks, silently un-sharing the document;
- orphan cleanup missed them too, so deleting a shared file at source left its chunks behind.

**Fix:** `shared` becomes `bool | None`, where `None` means "no explicit intent" — each file inherits the sharing state already in the index (`ConnectorFileProcessor._resolve_shared`), falling back to private when not indexed. The UI already sends the flag only while the toggle is on screen, so an explicit choice still wins and no frontend change was needed. Deletion scopes that have already pinned a document down by a stable identity (filename alias, connector file id) now use "owned by me OR ownerless" via the new `build_owner_or_shared_filter`.

This also fixes a bug reachable without COS: a confirmed overwrite of any ownerless same-named document silently no-opped, because the delete scope was chosen from how the replacement would be *written* rather than from what was actually indexed.

### 2. Change detection failed closed, silently

`modified_time` is back-filled *after* indexing by `_update_connector_metadata`. Any gap there — a file ingested before change detection existed, or an enrichment that didn't run — left a `None` stored value, and `remote_is_newer_than_synced` returns `False` for `None`. Those files reported "unchanged" forever, with no log line and nothing visible to the user.

**Fix:** bucket connectors (COS both auth modes, S3, Azure) now report the object's **ETag** in listings and persist it at ingest as a top-level `content_etag` keyword. `classify_remote_file_change` checks it first and it is decisive in both directions: differing tags mean changed *even with no comparable timestamp*, and equal tags mean unchanged *even if the timestamp moved* (a copy or a re-upload of identical bytes no longer triggers a pointless re-ingest). Timestamps remain the fallback. Files with no comparable signal at all are now counted and named in a warning instead of quietly passing as up to date.

### 3. The confirmation dialog overstated updates

The dialog already had an "N files will be updated" section — wired to `syncedCount`, the *total* number of indexed files. A sync of 40 COS files with one modified announced "40 files will be updated", then updated one.

**Fix:** the preview endpoints return the changed files themselves (`updates` / `updates_available`), computed from the same listing the orphan pass already performs, using the same classifier sync uses — so the dialog and the sync cannot disagree. Connectors on `replace_always` detection can't predict updates (they decide per file while ingesting); they report `updates_available: false` and render as "N files will be re-checked" rather than a number that looks like a prediction. "Nothing to change" now has its own state instead of "0 files will be updated".

## Notes for review

- **`delete_document_by_filename` is widened for all processors**, not just connectors: a confirmed overwrite of `report.pdf` will now also replace an instance-shared `report.pdf`. This matches what the duplicate check (already owner-agnostic) showed the user before they confirmed.
- **`_resolve_shared` costs one extra `size:1` search per file** on sync paths — negligible next to the download and Docling conversion each file already does, but it is per-file rather than per-sync.
- **Existing documents have no stored ETag** until next ingested. They keep working on the timestamp path and now appear in the warning if they have no signal at all; re-ingesting once with overwrite records a tag permanently.

## Verification

- Backend: 1737 passed. The 27 failures are pre-existing — byte-identical to a run on clean `main` (model providers, settings, encryption, opensearch security).
- Frontend: new tests pass under vitest; 4 pre-existing failing files (`task-context`, `task-error-content`, `task-notification-menu`, `upload-utils`) fail identically on clean `main`.
- Diff coverage: **100% of changed lines (29/29)**, threshold 80%.
- `ruff check` / `ruff format` clean on all 21 changed Python files; `tsc --noEmit` and `biome check` clean.

New tests: shared-file re-sync (verified failing without the fix — the file lands SKIPPED with `deleted_count: 0`), ETag classification and per-connector plumbing, preview update reporting, and the dialog's three-way updated/re-checked/nothing distinction.

## Summary by CodeRabbit

* **New Features**
  * Sync previews identify files that will be updated, separately from newly synced and deleted files.
  * Confirmation dialogs show changed files, re-check notices, unavailable checks, and “Nothing to change” states.
  * AWS S3, Azure Blob, and IBM COS synchronization detects overwritten files using object tags when available.

* **Bug Fixes**
  * Re-syncs preserve existing shared or private visibility unless explicitly changed.
  * Timestamp fallback remains available when object tags cannot be used.
  * Ownerless shared content is protected from deletion without the required permission.


src/api/connectors.py
```diff
@@ -575,7 +731,8 @@ async def delete_orphan_documents(
             opensearch_client,
             connector_type=connector_type,
             owner_user_id=None if shared else user_id,
-            shared=shared,
+            shared=bool(shared),
+            include_shared=not shared,
```
mpawlow: ### (1a) [Critical] Orphan reconciliation deletes ownerless/shared chunks without the anonymous-delete permission check

- ℹ️ Concur with CodeRabbit finding

#### Problem

- `ConnectorSyncBody.shared` is now `bool | None`; the re-sync paths (`/connectors/{type}/sync` with no explicit file selection, `sync_all`, webhook-triggered sync) pass `shared=body.shared` (typically `False`/`None` — no explicit "shared" intent) into `_sync_existing_connector_files` → `reconcile_orphans_for_connector_type` → `delete_orphan_documents`.
- Inside `delete_orphan_documents`, `shared` falsy now sets `include_shared=not shared` → `True`, which widens the deletion filter from "owned by this user" to `build_owner_or_shared_filter(user_id)`: **owned-by-me OR ownerless**.
- The only authorization gate for touching ownerless/shared chunks is in `connector_sync`:
  ```python
  if body.shared and not await has_effective_permission(
      request, user, rbac, "knowledge:delete:anonymous"
  ):
  ```
  This only fires when `body.shared` is **truthy**. A normal sync (`shared=False`/`None`, the default and by far the common case) skips this check entirely, yet — via `include_shared=True` — can still delete ownerless chunks.
- The candidate id set (`get_synced_file_ids_for_connector`) is queried with only a `{"term": {"connector_type": connector_type}}` filter (no owner/connection scoping), executed through the **user-scoped** OpenSearch client. Ownerless chunks are visible to every authenticated user by design (`securityconfig/roles.yml`, DLS `must_not exists owner` clause), so this id set can include connector-file-ids that were ingested by a *different user's* connection.
- Orphan detection then diffs that id set against the requesting user's own remote listing (their own COS/S3/Azure connection). Any ownerless file whose id isn't present there (a different bucket/account, an unrelated connection) is classified an orphan and deleted with `include_shared=True` — without the requester ever having `knowledge:delete:anonymous`.

#### Background Information

- This is a regression introduced by the fix for the PR's own stated bug #1 ("shared files were incorrectly skipped by orphan cleanup"). The old code was too narrow (never touched ownerless chunks on regular sync); the new code overcorrects by widening the scope unconditionally instead of gating it behind the same permission that already exists for the explicit-shared path.
- Test coverage confirms the gap: `tests/unit/test_shared_flag.py` only asserts the permission check fires when `shared=True`. No test exercises the `shared=False/None` orphan-reconciliation path against ownerless chunks, so this scope-widening was never checked against the authz boundary.

#### Code References

- `src/api/connectors.py:731-739` (`delete_orphan_documents` — `include_shared=not shared`)
- `src/api/connectors.py:1452-1465` (`connector_sync` — permission check gated on `body.shared` only)
- `src/api/connectors.py:115-176` (`get_synced_file_ids_for_connector` — id set scoped only by `connector_type`, no owner/connection filter)
- `src/utils/opensearch_queries.py:77-99` (`build_owner_or_shared_filter`)
- `securityconfig/roles.yml:14-23` (ownerless chunks visible to all authenticated users via DLS)

#### Potential Solution

- Require `knowledge:delete:anonymous` whenever a deletion pass will actually run with `include_shared=True` (i.e., whenever `shared is not True`), not only when the caller explicitly requested a shared sync.
- Alternatively, scope orphan/id discovery and deletion to the *connection* (or the ingesting user), not just `connector_type`, so a user's sync can never enumerate or affect another connection's ownerless chunks.

#### Alternative Solutions

- Gate the widened `include_shared` deletion scope behind an explicit re-check of `knowledge:delete:anonymous` inside `delete_orphan_documents` itself (defense in depth, independent of caller intent), falling back to owner-only deletion when the permission is absent.

frontend/components/sync-confirm-dialog-data.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export const summarizeSyncPreview = (
+  props: SyncPreviewInput,
+): SyncPreviewSummary => {
+  const {
+    orphans,
+    orphansByType,
+    orphansAvailableByType,
+    updates,
+    updatesByType,
```
mpawlow, line 54: ### (1b) [Normal] Per-file connector sync dialog silently drops the new "updates"/"re-check" reporting, defeating one of the PR's stated fixes


#### Problem

- The PR's bug 3 fix threads `updates` / `updatesAvailableByType` through `SyncConfirmDialog` so the dialog can distinguish "N files will be updated" from "N files will be re-checked" from "nothing to change," replacing the old (admittedly wrong) behavior of showing the total synced count as "will be updated."
- `frontend/components/knowledge-actions-dropdown.tsx` — the single-connector/per-file sync entry point — is **not part of this diff** and still invokes `SyncConfirmDialog` with only `connectorType`, `orphans`, and `syncedCount`:
  ```tsx
  <SyncConfirmDialog
    ...
    connectorType={connectorType}
    orphans={syncPreview?.orphans}
    syncedCount={syncPreview?.synced_count}
  />
  ```
  It never passes `updates` or `updatesAvailableByType`, even though the backend (`GET /connectors/{type}/sync/preview`) now returns both (`SyncPreviewResponse.updates` / `.updates_available`).
- In `summarizeSyncPreview`, with `updates` and `updatesAvailableByType` both `undefined`: `normalizedUpdates` is always `{}` → `totalUpdates = 0` → `hasUpdates = false`; and `recheckedByType` is only populated when `updatesAvailableByType[type] === false`, which never happens since the map is `undefined`. So `hasRechecks` is also always `false`.
- Result: for this call site, whenever there are no orphans, the dialog **always** falls to the final `else` branch — `"Everything is already up to date."` — even when the connector actually has pending changed files that will be re-ingested. This is a regression relative to even the pre-PR behavior (which at least overstated with the synced-count total); now it understates to zero.

#### Background Information

- Confirmed by absence: `frontend/components/knowledge-actions-dropdown.tsx` does not appear in `git diff main...HEAD --stat` for this PR.
- `frontend/components/sync-confirm-dialog.test.tsx` only exercises the single-connector mode with `updates`/`updatesAvailableByType` explicitly supplied, so this integration gap has no test coverage today.

#### Code References

- `frontend/components/knowledge-actions-dropdown.tsx:247-256`
- `frontend/components/sync-confirm-dialog-data.ts:54-68` (`recheckedByType` computation depends on `updatesAvailableByType`)
- `frontend/components/sync-confirm-dialog.tsx:294-303` (description fallback chain)

#### Potential Solution

- Update `knowledge-actions-dropdown.tsx` to read `updates` and `updates_available` off its `syncPreview` (single-connector `SyncPreviewResponse`) and pass them (`updates={syncPreview?.updates}`, `updatesAvailableByType={{[connectorType]: syncPreview?.updates_available ?? false}}`) to `SyncConfirmDialog`, mirroring the two call sites already updated.

frontend/components/sync-confirm-dialog.tsx
```diff
@@ -353,14 +356,32 @@ export const SyncConfirmDialog = ({
...
+                totalRechecked={totalRechecked}
+              />
+            ) : null}
+
+            {!hasDeletes && !hasUnavailable && !hasUpdates && !hasRechecks ? (
+              <Alert>
+                <Check className="size-5" />
+                <AlertTitle>Nothing to change</AlertTitle>
+                <AlertDescription className="col-start-2 block min-w-0">
+                  <p>No files were added, changed, or removed at the source.</p>
```
mpawlow: ### (1c) [Minor] "Nothing to change" copy still claims new-file detection it can't perform

- ℹ️ Concur with CodeRabbit finding

#### Problem

- The empty-state alert still reads: *"No files were added, changed, or removed at the source."*
- The sync preview only diffs already-indexed files against the source; it never enumerates brand-new remote files. Claiming "no files were added" is not something this code path can actually verify, so the message can be shown while new files do exist at the source.

#### Code References

- `frontend/components/sync-confirm-dialog.tsx:376-383`

#### Potential Solution

- Narrow the copy to what the preview actually knows, e.g. *"No indexed files need to be updated or removed."*

mpawlow (changes requested):
## Code Review 1

- See Critical PR comment (1a) for evaluation
- Note: No functional review performed (yet)

mpawlow (approved):
## Code Review 2

- ✅ Approved / LGTM  🚀 

## Functional Review 1

- ✅ Successfully verified that the identified failing scenario (updated source file in COS not synced) is completely fixed

ricofurtado:
I understand that the CI error was fixed by Mike, and will be solved in the merge.

PR #2370, opened 2026-09-14
fix: make onboarding model catalog discoverable

edwinjosechittilappilly:
## Summary
- sort generic-provider onboarding models with the same capability/recency ranking used by the Settings picker
- preview five models and provide an explicit **Show all N models** action instead of silently truncating the flat list at 40
- keep search over the full catalogue and clarify that Azure users can type a deployment name
- add regression tests for ordering, expansion, and search beyond the preview

## Root cause
Generic onboarding passed the backend's alphabetically sorted model list directly to a flat ModelSelector, which rendered only its first 40 unsearched rows. Scrolling could not reveal rows beyond that cap, and `gpt-4.1` was hard to discover near the end of the visible slice.

## Validation
- `NODE_OPTIONS=--no-experimental-webstorage npm test` — 28 files, 314 tests passed
- `npm run typecheck`
- `npm run lint` — passed with existing repository warnings
- changed-file Biome check passed

## Summary by CodeRabbit

* **New Features**
  * Azure onboarding model selection now displays a provider-specific search hint for deployment names.
  * Azure model selection requires an explicit choice rather than automatically selecting a default.
  * Model pickers initially show five models, with an option to reveal the complete list.
  * Searching can find models beyond the initial preview without expanding the full list.
  * Available provider models are presented in version order, with newer models listed first while older Azure models remain available.


Wallgau (approved):
Nice, LGTM!

PR #2371, opened 2026-09-14
ci: propose upstream service upgrade pull requests

edwinjosechittilappilly:
OpenRAG pins OpenSearch, Docling Serve, and the Langflow base image in different places, so new upstream releases currently require manual discovery and edits. This adds a daily and manually triggered workflow that checks each service independently and opens or updates one upgrade PR per service when a newer stable version is available.

The OpenSearch check requires matching jvector, neural-search, and Prometheus exporter release assets before proposing an upgrade, then updates the Dockerfile, Dashboards pins, and related docs together. Docling Serve and Langflow checks update their pinned launcher/image versions; their upgrade PRs call out compatibility review before merge. The generated PRs never merge automatically.

Validation: `python3 scripts/ci/test_update_upstream_version.py` (5 passing tests); workflow YAML parses; `git diff --check` passes.

Repository setup: enable **Allow GitHub Actions to create and approve pull requests** under Actions → General → Workflow permissions. An optional `UPSTREAM_UPDATE_TOKEN` can be supplied if generated PRs should trigger regular CI workflows; PRs created with the default `GITHUB_TOKEN` do not trigger them.

## Summary by CodeRabbit

* **New Features**
  * Added automated daily checks for compatible upstream releases across OpenSearch, Docling Serve, and Langflow.
  * Automatically prepares or updates upgrade pull requests when stable compatible versions are available.
  * Verifies required OpenSearch plugin assets before proposing upgrades.

* **Tests**
  * Added coverage for stable version selection, release compatibility checks, and coordinated version updates.


scripts/ci/update_upstream_version.py
```diff
@@ -0,0 +1,187 @@
...
+def github_release_assets(repo: str, tag: str) -> set[str]:
+    release = get_json(f"{GITHUB_API}/repos/{repo}/releases/tags/{tag}")
+    return {asset["name"] for asset in release["assets"]}
+
+
+def opensearch_plugin_assets_exist(version: str) -> bool:
+    plugin_version = f"{version}.0"
+    expected = (
+        ("opensearch-project/opensearch-jvector", "artifacts.tar.gz"),
+        ("IBM/neural-search-jvector", f"opensearch-neural-search-{plugin_version}.zip"),
```
mpawlow, line 61: ### (1a) [Major] OpenSearch plugin asset-name pattern is stale — the automation will never propose a real OpenSearch upgrade

#### Problem

- `opensearch_plugin_assets_exist()` checks the `IBM/neural-search-jvector` repo for an asset literally named `opensearch-neural-search-{plugin_version}.zip`.
- That naming convention only applies to the `3.6.0.0` release (the version already pinned in this repo today). Every release from `3.7.0.0` onward is actually named `opensearch-neural-search-jvector-{plugin_version}.zip` (note the extra `-jvector`).
- Verified directly against the live GitHub API:
  - `IBM/neural-search-jvector` tag `3.7.0.0` → asset `opensearch-neural-search-jvector-3.7.0.0.zip`
  - `IBM/neural-search-jvector` tag `3.8.0.0` → asset `opensearch-neural-search-jvector-3.8.0.0.zip`
  - `IBM/neural-search-jvector` tag `3.6.0.0` → asset `opensearch-neural-search-3.6.0.0.zip` (old naming, matches the script)
- Because the check is a strict `asset not in github_release_assets(...)`, every candidate from `3.7.0` onward will fail the asset check and `update_opensearch()` will print "Skipping OpenSearch X: matching plugin assets are unavailable" and return `None` — forever. The OpenSearch leg of this automation is effectively dead on arrival for any version past the one already pinned.
- This fails silently (no PR, no error) rather than loudly, which makes it especially easy to miss — the workflow will report `changed=false` every day and look like it's "just waiting for a compatible release."

#### Background Information

- The other two asset checks are fine: `opensearch-project/opensearch-jvector` always ships a fixed `artifacts.tar.gz` regardless of version (matches), and `opensearch-project/opensearch-prometheus-exporter` assets are consistently named `prometheus-exporter-{plugin_version}.zip` (matches, confirmed for `3.7.0.0`/`3.8.0.0`).
- This is not covered by the test suite because `opensearch_plugin_assets_exist` is always mocked out in the tests that exercise `update_opensearch()` (see 1d) — nothing exercises the real asset-name logic against real or representative data.

#### Potential Solution

- Update the expected asset name to match the current convention: `f"opensearch-neural-search-jvector-{plugin_version}.zip"`.
- Since the naming convention itself has already changed once, prefer a check that isn't tied to one exact filename — e.g. match against a glob/regex (`opensearch-neural-search*{plugin_version}.zip`) or just check that *any* `.zip` asset exists on the release, rather than hardcoding the exact prefix.

#### Alternative Solutions

- Query the release and check for an asset whose name contains `plugin_version` and ends in `.zip`, without hardcoding the `opensearch-neural-search` / `opensearch-neural-search-jvector` prefix — more resilient to future renames at the cost of being slightly less exact.

scripts/ci/update_upstream_version.py
```diff
@@ -0,0 +1,187 @@
...
+def version_key(value: str) -> tuple[int, int, int]:
+    match = VERSION.fullmatch(value)
+    if not match:
+        raise ValueError(f"Not a stable three-part version: {value}")
+    return tuple(map(int, match.groups()))
+
+
+def get_json(url: str) -> dict | list:
+    headers = {"Accept": "application/vnd.github+json", "User-Agent": "openrag-version-updater"}
+    if url.startswith(GITHUB_API) and os.environ.get("GH_TOKEN"):
```
mpawlow, line 34: ### (1b) [Major] `GH_TOKEN` is attached to outbound requests using a spoofable prefix check

- ℹ️ Concur with CodeRabbit.ai

#### Problem

- `get_json()` decides whether to attach the `Authorization: Bearer $GH_TOKEN` header using `url.startswith(GITHUB_API)`, where `GITHUB_API = "https://api.github.com"`.
- `str.startswith` does not validate a host boundary: `"https://api.github.com.attacker.example/...".startswith("https://api.github.com")` is `True`. Any URL beginning with that literal prefix — including on an attacker-controlled subdomain-looking host — will get the token attached.
- `update_langflow()` feeds exactly this kind of externally-controlled value into `get_json()`: it follows Docker Hub's own pagination cursor, `data["next"]`, in a loop (`while url: data = get_json(url); ... url = data["next"]`), with no validation. If the Docker Hub response were ever compromised (compromised registry, cache poisoning, MITM), a crafted `next` URL could redirect the token to an attacker-controlled host.
- This is CWE-200 (Exposure of Sensitive Information). Exploitability is difficult (requires compromising the Docker Hub API response), but the token in scope (`UPSTREAM_UPDATE_TOKEN` or `github.token` with `contents: write` + `pull-requests: write`) is a real credential leak if triggered.

#### Code References

`scripts/ci/update_upstream_version.py:34` (the check), `scripts/ci/update_upstream_version.py:148-156` (the only call site where the URL is not a hardcoded/self-constructed GitHub API URL)

#### Potential Solution

```python
from urllib.parse import urlsplit
...
parsed_url = urlsplit(url)
if (
    parsed_url.scheme == "https"
    and parsed_url.hostname == "api.github.com"
    and os.environ.get("GH_TOKEN")
):
    headers["Authorization"] = f"Bearer {os.environ['GH_TOKEN']}"
```

#### Alternative Solutions

- Split `get_json()` into a GitHub-specific variant (always attaches the token, only ever called with URLs built from `GITHUB_API` locally) and a generic variant used for Docker Hub/PyPI (never attaches the token) — removes the ambiguity in the shared function's contract entirely, not just the string-match bug.

scripts/ci/update_upstream_version.py
```diff
@@ -0,0 +1,187 @@
...
+
+def update_docling() -> tuple[str, str] | None:
+    path = "src/tui/managers/docling_manager.py"
+    content = (ROOT / path).read_text()
+    match = re.search(r'"docling-serve\[ui\]==(\d+\.\d+\.\d+)"', content)
+    if not match:
+        raise ValueError("Could not find the Docling Serve pin")
+    current = match.group(1)
+    metadata = get_json(PYPI_API)
+    latest = metadata["info"]["version"]
```
mpawlow, line 132: ### (1c) [Normal] `update_docling()` will crash the whole matrix job on a PyPI prerelease

- ℹ️ Concur with CodeRabbit.io feedback

#### Problem

- `latest = metadata["info"]["version"]` reads PyPI's "latest" field directly and immediately calls `version_key(latest)` without validating it's a stable `X.Y.Z` version.
- `version_key()` raises `ValueError` on anything that doesn't match `^(?:v)?(\d+)\.(\d+)\.(\d+)$` — including a prerelease like `1.33.0rc1`. If `docling-serve`'s PyPI "latest" metadata is ever a prerelease (a real possibility depending on how/when the maintainers publish), this raises uncaught out of `update_docling()`, `main()` propagates it, and the `docling` matrix leg fails outright instead of cleanly skipping.
- Failure is contained to the `docling` leg only (`fail-fast: false` on the matrix), so it won't block the `opensearch`/`langflow` legs, but it will fail the daily scheduled run for that component with no PR and a red job every day until PyPI's latest is stable again.

#### Background Information

- The other two integrations (`update_opensearch`, `update_langflow`) both filter candidates through `VERSION.fullmatch(...)` *before* calling `version_key`, so `update_docling` is the only one of the three missing this guard — an inconsistency within the file itself, not just a one-off gap.

#### Potential Solution

```python
latest = metadata["info"]["version"]
if not VERSION.fullmatch(latest) or version_key(latest) <= version_key(current):
    return None
```

scripts/ci/test_update_upstream_version.py
```diff
@@ -0,0 +1,80 @@
...
+            "docs/docs/reference/configuration.mdx",
+            "docs/docs/support/troubleshoot.mdx",
+        )
+        with tempfile.TemporaryDirectory() as directory:
+            root = Path(directory)
+            for path in paths:
+                target = root / path
+                target.parent.mkdir(parents=True, exist_ok=True)
+                copyfile(updater.ROOT / path, target)
+            releases = [{"tag_name": "3.7.0", "draft": False, "prerelease": False}]
```
mpawlow, line 64: ### (1e) [Minor] OpenSearch update test hardcodes `3.6.0`/`3.7.0` instead of deriving them from the copied fixture

- ℹ️ Concur with CodeRabbit.ai feedback

#### Problem

- `test_opensearch_updates_image_plugins_dashboards_and_docs` copies the *actual current repo files* (`Dockerfile`, `docker-compose.yml`, Helm `values.yaml`, docs) into a temp directory, then hardcodes the expected current (`3.6.0`) and next (`3.7.0`) versions in its assertions.
- Once the real OpenSearch pin advances past `3.6.0` (which this very automation is designed to do), this test's fixture no longer matches its hardcoded expectations and it will fail — the release list `[{"tag_name": "3.7.0", ...}]` will no longer be `> current`, so `update_opensearch()` returns `None` instead of the expected tuple.
- Low urgency in practice because the test isn't wired into any automated run (see 1g), so nothing will actually notice the breakage until someone runs it by hand.

#### Potential Solution

- Read the current version from the copied `Dockerfile` after copying it, derive a synthetic "next" patch version from it (e.g. bump the patch digit), and use both dynamically in the releases fixture and all assertions instead of the literals `3.6.0`/`3.7.0`.

#### Alternative Solutions

- Use synthetic/fixture file contents instead of copying the live repo files at all, so the test is fully self-contained and immune to drift in the real pin — closer to how the other three tests in the file already work (they write synthetic fixture content rather than copying real files).

scripts/ci/update_upstream_version.py
```diff
@@ -0,0 +1,187 @@
...
+DOCKER_HUB_API = "https://hub.docker.com/v2/repositories/langflowai/langflow/tags/"
+PYPI_API = "https://pypi.org/pypi/docling-serve/json"
+VERSION = re.compile(r"^(?:v)?(\d+)\.(\d+)\.(\d+)$")
+
+
+def version_key(value: str) -> tuple[int, int, int]:
+    match = VERSION.fullmatch(value)
+    if not match:
+        raise ValueError(f"Not a stable three-part version: {value}")
+    return tuple(map(int, match.groups()))
```
mpawlow, line 29: ### (1f) [Minor] New script files sit outside CI's lint scope, and a manual `mypy` run already finds 5 type errors

#### Problem

- `lint-backend.yml` only triggers on `src/**/*.py`, `tests/**/*.py`, `pyproject.toml`, `uv.lock`. Both new files live under `scripts/ci/`, so this PR's Python code is **not** linted or type-checked by CI at all, in either direction (now or on any future edit to these files).
- Running `mypy` against `scripts/ci/update_upstream_version.py` by hand surfaces 5 errors, all stemming from `get_json()`'s loose `dict | list` return type being subscripted as a dict everywhere it's called (`release["assets"]`, `metadata["info"]["version"]`, `data["results"]`, `data["next"]`), plus one `tuple[int, ...]` vs `tuple[int, int, int]` mismatch in `version_key`.
- Functionally this is mostly benign today (a malformed response would raise `KeyError`/`TypeError`, which is consistent with the script's own "fail loud rather than silently hide updates" design intent), but it means real type errors in new, unreviewed-by-tooling code ship straight to `main` with no static check ever exercising them.

#### Code References

`scripts/ci/update_upstream_version.py:29,54,132,152,156`

#### Potential Solution

- Give `get_json` an overload or a `TypeVar`-bound signature per call site (or two thin wrappers, `get_json_dict`/`get_json_list`) so callers get real dict/list typing instead of `dict | list` everywhere.
- Consider widening `lint-backend.yml`'s path filters to include `scripts/**/*.py` (or a dedicated lint job for `scripts/`) so new CI-automation Python isn't permanently exempt from Ruff/mypy.

#### Alternative Solutions

- If `scripts/` is intentionally kept outside the enforced lint scope as a lower-bar area (consistent with other existing `scripts/ci/*.py` files, none of which are linted today either), at minimum this is worth a maintainer decision rather than an unreviewed default — it's not something this PR introduced, but this PR is the first to pair such a script with meaningful branching logic and a matching test suite.

scripts/ci/test_update_upstream_version.py
```diff
@@ -0,0 +1,80 @@
+"""Regression tests for the version proposal inputs and guarded edits."""
```
mpawlow, line 1: ### (1g) [Minor] New regression test suite isn't wired into any Make target, npm script, or CI workflow

#### Problem

- `scripts/ci/test_update_upstream_version.py` is referenced only in the PR description ("Validation: `python3 scripts/ci/test_update_upstream_version.py` (5 passing tests)") as a manual, one-time validation step.
- No workflow YAML, Makefile target, or `pytest` discovery path invokes it: `make test` runs `pytest tests/` only; `.github/workflows/*.yml` has no reference to `scripts/ci/test_update_upstream_version.py`.
- This mirrors an already-known pattern in this repo (documented in project conventions) where standalone test files exist but are never run automatically — meaning a future edit to `update_upstream_version.py` (including a fix for 1a/1b/1c above) could silently break these tests with nothing to catch it.

#### Code References

`scripts/ci/test_update_upstream_version.py:1` (whole file), `.github/workflows/upstream-version-prs.yml` (no test step)

#### Potential Solution

- Add a `python3 -m unittest scripts/ci/test_update_upstream_version.py` (or `pytest scripts/ci/`) step to `upstream-version-prs.yml` before the "Check upstream and update pins" step, or as its own lightweight CI job/workflow gated on changes to `scripts/ci/update_upstream_version.py` — so a bad edit to the updater script fails fast instead of silently corrupting or skipping upgrade PRs.

#### Alternative Solutions

- At minimum, add a Makefile target (e.g. under `help_utils`) so a human can discover and run it without reading the PR description.

mpawlow (changes requested):
## Code Review 1

- See Major PR comments (1a), (1b) for evaluation
- See Normal PR comments (1c) for consideration
- See Minor PR comments (1e), (1f), (1g) for optional consideration

## Questions

@edwinjosechittilappilly 
- (q1): Can **Renovate** or **Dependabot** be used to solve the generic pieces of this problem out of the box?
   - Concern: About rolling our own solution and maintaining it

.github/workflows/upstream-version-prs.yml
```diff
@@ -0,0 +1,61 @@
+name: Propose upstream service upgrades
+
+on:
+  schedule:
+    - cron: '17 5 * * *'
```
mpawlow, line 5: ### (1h) [Info] Cron schedule is reasonable

- Runs daily at 05:17 UTC. 
- i.e. minute 17, hour 5, every day of month, every month, every day of week

PR #2375, opened 2026-09-14
fix: include connector_file_id-only files in orphan filename map

Thenathanb:
## Summary

Fixes IBM tracker issue [lakehouse/tracker#78584](https://github.ibm.com/lakehouse/tracker/issues/78584): deleted-file names weren't shown in the sync confirmation dialog for OAuth connectors (Google Drive, SharePoint, OneDrive) — the dialog fell back to rendering the raw file id/GUID instead.

**Root cause:** `get_synced_id_to_filename_map()` in `src/api/connectors.py` built its `{id: filename}` lookup by aggregating only on `document_id`. For connectors ingested via the standard `ConnectorFileProcessor` path, the real source id is stored in `connector_file_id` and `document_id` holds an unrelated content hash — so the lookup always missed and `filename` came back empty, tripping the frontend's `filename || document_id` fallback.

This affects every connector on that ingestion path (S3, Azure Blob, IBM COS included), but was only visually obvious for OAuth connectors since their ids are opaque provider GUIDs; bucket connectors build a human-readable composite id (`bucket::key`), which happened to mask the same bug.

Traced the regression to a preexisting gap from the function's original implementation (2026-05-28), not the later sync-refactor PR that was initially suspected — confirmed by diffing that commit directly.

## Fix

Mirrors the existing dual-aggregation pattern already used by the sibling function `get_synced_id_to_modified_time_map()`: aggregate on both `connector_file_id` and `document_id` (each with a nested `top_filename` sub-agg), with the same `.keyword` fallback retry for older indices, merging so `connector_file_id` wins on key collision.

Single-function change, no signature/caller changes.

## Test plan

- [x] Added 4 unit tests to `tests/unit/api/test_connector_change_detection.py`: regression coverage for connector_file_id-only buckets, collision precedence, `.keyword` fallback, and error handling.
- [x] Full `test_connector_change_detection.py` + `test_reconcile_orphans_for_connector_type.py` suites pass (40 tests).
- [x] `ruff check` clean on changed files.
- [ ] Manual verification against a live OAuth connector (optional — logic is covered by unit tests against the real aggregation/merge code path).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved synchronization change detection across current and legacy document layouts.
  - Ensured the most specific connector file information is used when records contain overlapping identifiers.
  - Added fallback handling for older indexed data, reducing failures when searching for synchronized files.
  - Added safeguards and warnings when result limits are reached.
  - Improved resilience by returning an empty result instead of failing when synchronization searches encounter errors.


mfortman11 (approved):
LGTM

PR #2377, opened 2026-09-15
fix: changed ingest settings to have copy that doesn't reference specific providers

lucaseduoli:
This pull request updates the messaging in the `IngestSettingsSection` component to clarify model provider requirements for users. The main change is:

User messaging improvement:

* Updated the message displayed when no models are detected to instruct users to "Configure a model provider with vision-capable models first," making it clearer that vision-capable models are required. (`frontend/app/settings/_components/ingest-settings-section.tsx`)


edwinjosechittilappilly (approved):
LGTM

PR #2379, opened 2026-09-15
fix: connector sync truncates the source listing at 100 files

ricofurtado:
Found while investigating whether the COS overwrite defect (#2366) also affected AWS S3. It does not — S3 change detection works correctly on `main`, verified by driving a real `S3Connector` through the real classifier. But the investigation turned up a separate truncation bug that produces the same user-visible symptom, on **every** connector.

## The bug

`sync_connector_files` collected the source listing like this:

```python
page_size = min(max_files or 100, 1000) if max_files else 100
file_list = await connector.list_files(page_token, max_files=page_size)
...
if (max_files and len(files_to_process) >= max_files) or not file_list.get("nextPageToken"):
    break
page_token = file_list.get("nextPageToken")
```

Two mistakes compound:

1. With no `max_files`, it still asked for a page size of **100**.
2. It read the continuation token under **`nextPageToken`** — a key no connector in this repo emits. All of them return `next_page_token`.

So the token was always `None`, the loop broke after one call, and the cap of 100 had already been applied. Confirmed against the real S3 connector: 150 objects in the bucket, 100 returned, loop terminates.

Who it hits:

| Connector | Paging style | Effect |
| --- | --- | --- |
| `aws_s3`, `ibm_cos`, `azure_blob` | SDK paginates internally, returns one flat list | capped at first 100 objects |
| `google_drive` | same | capped at first 100 files |
| `sharepoint` | returns a real Graph `$skiptoken` in `next_page_token` | token discarded, stops after page 1 |

## Why it is a correctness bug, not a slow sync

The filename-fallback branch of `_sync_existing_connector_files` passes the indexed filenames as `filename_filter`. A file past the cap was never listed, so it never matched the filter and never got re-ingested — **its stale copy stays in Knowledge indefinitely**. That is the same thing a user reports as "I overwrote the file and sync didn't pick it up", reached by a different route than #2366.

Also reachable via `connector_sync` with `sync_all=true` (API-only; no frontend sender), where it silently ingests only the first 100 files.

## The fix

- Pass `max_files` straight through — `None` means no cap, so connectors that paginate internally return everything.
- Read the token under both spellings, so a connector written against either convention pages correctly.

## Also in this PR

**`OPENRAG_DEV_AWS_S3`** — S3 was the only bucket connector with no dev bypass (`is_available` returned bare `IBM_AUTH_ENABLED`, while COS has `OPENRAG_DEV_IBM_COS` and Azure has `OPENRAG_DEV_AZURE_BLOB`). It could not be exercised end to end without standing up IBM auth, which is the likeliest reason its sync path has had less real-world validation than its siblings'. With the flag, S3 can be driven against MinIO locally.

**`tests/unit/connectors/test_aws_s3_connector.py`** — S3 had no connector test file, though `test_ibm_cos_connector.py` and `test_azure_blob_connector.py` both exist. Pins S3's own change-detection signals (listing id/mtime/etag, etag persisted at download, overwrite classification across each combination of stored signals, and identical-bytes-with-a-moved-timestamp) rather than inferring them from machinery shared with COS.

## Verification

- 5 new paging tests, confirmed **failing** against the old code (3 fail: uncapped truncation, SharePoint token dropped, filter missing files past the cap).
- 8 new S3 connector tests.
- Full unit sweep: 31 failures, byte-identical to a run on clean `main` (model providers / settings / encryption — all pre-existing).
- `ruff check` and `ruff format` clean.

No frontend changes.

## Summary by CodeRabbit

- **New Features**
  - Added an optional development setting to enable AWS S3 and S3-compatible storage connectivity without IBM authentication.
  - AWS S3 synchronization now recognizes reliable change signals and skips directory placeholder objects.

- **Bug Fixes**
  - File synchronization now follows continuation tokens across all pages, including empty pages and supported token formats.
  - Uncapped synchronization no longer truncates results prematurely.
  - Explicit file limits remain enforced across multiple pages, including zero-file limits.
  - Filters apply beyond the first page.
  - AWS S3 metadata handling preserves consistent entity tags and accurately detects file changes.


src/connectors/service.py
```diff
@@ -284,13 +299,9 @@ async def sync_connector_files(
                 files_to_process.append(file_info)
 
             # Stop if we have enough files or no more pages
-            if (max_files and len(files_to_process) >= max_files) or not file_list.get(
-                "nextPageToken"
-            ):
+            if (max_files and len(files_to_process) >= max_files) or not page_token:
                 break
```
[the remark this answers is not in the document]
ricofurtado: Fixed

src/connectors/service.py
```diff
@@ -256,17 +268,24 @@ async def sync_connector_files(
...
         page_token = None
 
-        # Calculate page size to minimize API calls
-        page_size = min(max_files or 100, 1000) if max_files else 100
-
-        while True:
-            # List files from connector with limit
-            logger.debug("Calling list_files", page_size=page_size, page_token=page_token)
-            file_list = await connector.list_files(page_token, max_files=page_size)
+        # A zero cap means "sync nothing" and has to short-circuit before the
```
mpawlow, line 271: ### (2a) [Minor] Negative `max_files` silently produces an empty sync instead of being rejected

- ℹ️ Concur with CodeRabbit.ai feedback

#### Problem

- The new zero-cap fix is `while max_files != 0:`, which correctly short-circuits `0` but does nothing for negative values.
- `ConnectorSyncBody.max_files: int | None = None` (`src/api/connectors.py:950`) still has no lower-bound constraint, and `connector_sync` forwards it unchanged when `sync_all=True`.
- For `max_files=-1`:
  - The loop is entered (`-1 != 0`).
  - `connector.list_files(page_token, max_files=-1)` is still called — a full, wasted round trip against the real source (verified against the mock connector: it fetches and returns all 150 files before anything is discarded).
  - Every per-file/aggregate check (`if max_files and len(files_to_process) >= max_files: break`) evaluates true immediately (`0 >= -1`), so `files_to_process` ends up empty.
  - Net effect: the sync silently completes having "processed" zero files, with no error surfaced to the caller — indistinguishable from "the source is empty."

#### Code References

- `src/connectors/service.py:271` — `while max_files != 0:` (doesn't guard negative values)
- `src/connectors/service.py:290, 306` — `if max_files and len(files_to_process) >= max_files:` (truthy check treats negatives as an immediately-exceeded cap)
- `src/api/connectors.py:950` — `ConnectorSyncBody.max_files: int | None = None` (no `ge=0` constraint)

#### Potential Solution

- Add a `ge=0` constraint to `ConnectorSyncBody.max_files` in `src/api/connectors.py` (Pydantic `Field(default=None, ge=0)`), so the API rejects negative values with a 422 instead of accepting them and silently producing an empty result.

#### Alternative Solutions

- Guard in `sync_connector_files` itself: `while max_files is None or max_files > 0:`, treating any `max_files <= 0` as "sync nothing" — cheaper to add here but leaves the API contract silently permissive for a nonsensical input rather than rejecting it at the boundary.

src/connectors/service.py
```diff
@@ -256,17 +268,24 @@ async def sync_connector_files(
...
+        # None (no cap) and positive caps take the loop as before.
+        while max_files != 0:
+            # Pass max_files straight through — None means "no cap". Asking for a
+            # synthetic page size instead silently truncated every sync: the
+            # connectors that paginate internally (all three bucket ones, and
+            # Google Drive) honour the cap and then report next_page_token=None,
+            # so there was no token to continue with and everything past the
+            # first page was simply dropped.
+            logger.debug("Calling list_files", max_files=max_files, page_token=page_token)
+            file_list = await connector.list_files(page_token, max_files=max_files)
```
mpawlow, line 283: ### (1b) [Normal] Removing the page-size cap can send an unbounded `$top` to SharePoint/Graph for large `max_files`

#### Problem

- The old code bounded the per-page request size regardless of the caller's `max_files`: `page_size = min(max_files or 100, 1000)`.
- The new code passes `max_files` straight through to `connector.list_files(page_token, max_files=max_files)` with no upper bound.
- `SharePointConnector.list_files` sets `params["$top"] = str(max_files_value)` directly from that argument (`src/connectors/sharepoint/connector.py:482,492`) on **every** page request — not just the first.
- If a caller supplies a large explicit `max_files` (e.g. a bulk API-only sync_all request), each page request now asks Graph for `$top=<max_files>` instead of a sane page size, which risks Graph API `$top` limits/rejections or at minimum defeats the purpose of paging.

#### Background Information

- This is a secondary side effect of the same refactor that fixed the truncation bug — the PR correctly removed the *broken* page-size logic but didn't replace it with a *correct* one for token-paginated connectors, only for the internally-paginating bucket connectors (where `max_files` is a post-fetch stop condition, not a raw request parameter, so no equivalent risk exists there).

#### Code References

- `src/connectors/service.py:268-269` (removed page-size calculation) / `src/connectors/service.py:281` (`connector.list_files(page_token, max_files=max_files)`)
- `src/connectors/sharepoint/connector.py:482,492` (`$top` bound directly to the passed-in `max_files`)

#### Potential Solution

- Reintroduce a page-size cap at the call site specifically for the per-page request, independent of the aggregate `max_files`: e.g. `page_request_size = min(max_files, 1000) if max_files else None`, and pass that to `connector.list_files` while still tracking the aggregate `max_files` for the stop condition.

#### Alternative Solutions

- Push the per-page ceiling into `SharePointConnector.list_files` itself (clamp `max_files_value` to a documented maximum before building `$top`), so the service layer doesn't need to know about Graph's page-size limits.

src/config/settings.py
```diff
@@ -317,6 +317,21 @@ def is_dev_azure_blob_enabled() -> bool:
...
+
+    Allows testing the S3 connector (e.g. against MinIO) in a local environment
+    where IBM auth is not configured. Never enable in production. Requires
+    ``OPENRAG_DEV_AWS_S3=true``.
+
+    S3 was the only bucket connector with no dev bypass, so it could not be
+    exercised end to end without standing up IBM auth — which is why its sync
+    behaviour has historically had less real-world validation than COS's.
+    """
+    raw = os.getenv("OPENRAG_DEV_AWS_S3", "false").strip().lower()
```
mpawlow, line 331: ### (1c) [Minor] `OPENRAG_DEV_AWS_S3` is undocumented in `.env.example`

#### Problem

- The two sibling dev-bypass flags each have a dedicated comment block in `.env.example`: `OPENRAG_DEV_IBM_COS` (line 202-204) and `OPENRAG_DEV_AZURE_BLOB` (line 224-227).
- The new `OPENRAG_DEV_AWS_S3` flag has no equivalent entry anywhere in `.env.example`, breaking the pattern this PR itself cites as precedent ("S3 was the only bucket connector with no dev bypass").
- Per the repo's own convention (`AGENTS.md`/`CLAUDE.md`: "changes that add ... an env var usually need a docs change in the same PR"), this is the one place that convention wasn't followed.

#### Code References

- `.env.example:187-189` (AWS section — no dev-bypass block, unlike COS/Azure below it)
- `src/config/settings.py:320-321` (new `is_dev_aws_s3_enabled()`)

#### Potential Solution

- Add a short comment block after the existing `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` pair in `.env.example`, mirroring the IBM COS block: document `OPENRAG_DEV_AWS_S3=true` for local/MinIO testing, default `false`, never-in-production warning.

mpawlow (commented):
## Code Review 1/2

- See Minor PR comment (2a) (1c) for optional consideration
- Note: **Functional Review 1** is pending

mpawlow (approved):
## Functional Test 1

- ✅ Approved / LGTM  🚀 
- ✅ Successfully synced 100+ files
   - Used **Azure Blog Storage** connector

```javascript
  fetch('/api/connectors/azure_blob/sync', {
    method: 'POST',
    headers: { 'Content-Type': 'application/json' },
    body: JSON.stringify({ sync_all: true })
  }).then(r => r.json()).then(console.log)
```

Image

PR #2393, opened 2026-09-16
feat: Add multi lang support for OCR

mfortman11:
### Problem

OpenRAG never sent a language to Docling, so each OCR engine fell back to its own default — `ocrmac` (macOS) to English only, `easyocr` (Linux/container) to `["fr", "de", "es", "en"]`. Non-Latin text was silently dropped during ingest: a document with Japanese and English content came back English-only, with no error. There was no OCR language field anywhere in `KnowledgeConfig`, so there was nothing to send even if the call site had wanted to.

### What this adds

Screenshot 2026-09-16 at 4 25 30 PM

A curated 15-language OCR picker under a new **Advanced OCR Settings** section in ingest settings, stored on `KnowledgeConfig.ocr_languages` (default `["en"]`) and translated to engine-specific codes at request time.

**Backend**

- `KnowledgeConfig.ocr_languages: list[str]`, default `["en"]`. Existing config files pick up the default automatically.
- `OCR_LANGUAGE_CODES` + `resolve_ocr_languages(engine, languages)` in `docling_service.py` translate neutral codes (ISO 639-1, plus `zh-Hans`/`zh-Hant`) to `ocrmac` BCP-47 tags or `easyocr` short codes. Unknown codes pass through untranslated, so an operator can name an engine code OpenRAG does not know rather than have it dropped.
- `get_docling_preset_configs()` now emits `ocr_preset` and `ocr_lang`, omitting `ocr_lang` when the list is empty. `DoclingConfig` updated in step — it required `ocr_engine`, so the legacy preset endpoint would have 500'd on a `ValidationError` once the key was renamed.
- Switched from the deprecated `ocr_engine` field to `ocr_preset` (correctness housekeeping; `ocr_lang` works with either).
- `ocr_languages` wired through settings GET/PUT and `langflow_sync`, with a validator rejecting empty lists, blank entries, and cross-script combinations. The legacy preset endpoint now carries configured languages through instead of wiping them from the flow.

**Frontend**

- `OCR_LANGUAGE_OPTIONS` catalog in `lib/constants.ts`; `ocr_languages?: string[]` on the settings query/mutation types.
- New `lib/ocr-languages.ts` holds the family table plus pure `allowedLanguages()` / `isSelectionValid()`, kept out of the component so it is testable without rendering.
- `MultiSelect` gained `disabled`, plus `disabled`/`hint` on `Option`, so incompatible languages stay visible and explain themselves rather than vanishing.
- The picker sits in a `Collapsible` mirroring the Advanced Vision Model section: dimmed when the OCR toggle is off, fields disabled rather than hidden. Empty selection falls back to `["en"]`.

### Why the combination rules

`easyocr` loads one recognition model per job and rejects languages outside it (`ValueError: Japanese is only compatible with English`); `ocrmac` has no such restriction. Since the engine is chosen by host OS, Japanese + Korean is valid on a macOS dev machine and a hard error in a Linux container. The picker applies the stricter engine's rules everywhere — non-English selections must come from one family, English combines with anything — so a saved config is portable. The same rules are mirrored server-side in the `SettingsUpdateBody` validator, deliberately duplicated rather than shared, and tested independently on both sides.

The curated 15 are the `ocrmac ∩ easyocr` intersection. Excluded: Cantonese and Najdi Arabic (ocrmac only) and easyocr's remaining ~71 languages. The list caps the UI, not the capability — anything outside it is still reachable via raw-code pass-through.

### ⚠️ Blocked upstream — the setting is currently inert

**Do not expect this to work end-to-end on docling-serve 1.26.0.** The engine honors `ocr_lang` when called directly in-process (deterministic, 4/4: `OcrMacOptions(lang=["ja-JP","en-US"])` extracts all Japanese lines, `lang=["en-US"]` extracts none). Requests *through docling-serve* have never reproducibly honored it, on either 1.26.0 or 1.33.0, on both the multipart async endpoint and the JSON sync endpoint.

OpenRAG sends the option correctly — dumping the multipart body shows repeated `ocr_lang` parts alongside `ocr_preset=ocrmac` and `do_ocr=true`. One partial source-level finding stands for the multipart path: `FormDepends` in `docling_serve/helper_functions.py` flattens only **dict** fields to JSON strings, and `ocr_lang` is `list[str] | None`, so FastAPI leaves it `None`. That does not explain the JSON-endpoint failures.

Users can pick languages and the choice is stored, validated, and sent correctly, but ingest still extracts only the engine default. Remaining work: file the upstream bug, and decide whether to surface the limitation in the UI.

### Testing

14 new backend tests across `tests/unit/api/test_settings_ocr_languages.py`, `test_settings_endpoints.py`, `test_docling_service.py`, and `test_langflow_docling_ocr_sync.py`, plus 11 in `frontend/lib/ocr-languages.test.ts`. Full backend suite diffed against a clean-tree baseline shows zero new failures (28 pre-existing failures and 2 collection errors in `tests/unit/test_agent_*.py` are unrelated). Frontend: 337 tests pass, `tsc --noEmit` and `biome check` clean.

Unit tests prove the right options are built and sent, not that OCR works. Any end-to-end check must start a **fresh** docling-serve per configuration — reusing one server silently reuses cached converters and produces false results in both directions.

## Summary by CodeRabbit

- **New Features**
  - Added advanced OCR settings for selecting recognition languages.
  - Supports multiple languages with compatibility guidance and automatic handling of incompatible combinations.
  - OCR language preferences are saved, restored, and applied during document processing.
  - Added platform-specific language support across supported OCR engines.

- **Bug Fixes**
  - Preserved configured OCR languages when synchronizing document-processing settings.
  - Added validation to prevent empty, blank, or incompatible language selections.


Wallgau (approved):
LGTM, great work!

PR #2394, opened 2026-09-17
fix: Address custom component flow mismatch and add test validation

mfortman11:
Fixes the flow and custom component mismatch issue.

Adds regression test and make backend tests run in the ci

Adds the `test backend/unit tests` ci job

## Summary by CodeRabbit

* **Tests**
  * Added automated validation for custom components and embedded flow code, including syntax, formatting, linting, compatibility, and synchronization checks.
  * Documented known baseline failures and added safeguards to detect newly introduced failures or resolved entries.

* **Chores**
  * Added continuous integration coverage for backend unit tests on relevant pull requests and pushes to the main branch.
  * Standardized formatting in component and flow definitions without changing runtime behavior.


tests/unit/conftest.py
```diff
@@ -39,8 +40,71 @@ def _reset_db_engine_module_state(monkeypatch):
...
+    return node_ids, ignored
+
+
+_KNOWN_FAILING_NODE_IDS, _KNOWN_UNIMPORTABLE = _read_known_failures()
+
+# Modules that fail at import cannot be xfailed; skip collecting them.
+collect_ignore = list(_KNOWN_UNIMPORTABLE)
+
+
+def pytest_collection_modifyitems(config, items):
```
mpawlow, line 87: ### (1a) [Normal] `OPENRAG_ENFORCE_KNOWN_FAILURES` stale-entry check assumes a full-suite run

#### Problem

- `pytest_collection_modifyitems` in `tests/unit/conftest.py` computes `stale = _KNOWN_FAILING_NODE_IDS - collected`, where `collected` is only what **this invocation** gathered.
- It never verifies the invocation actually covered all of `tests/unit`.
- Running any partial/filtered `pytest` invocation with `OPENRAG_ENFORCE_KNOWN_FAILURES=1` set raises a `pytest.UsageError` listing almost the entire `known_failures.txt` as "stale," even though nothing is actually stale.

#### Background Information

- Verified by reproduction: temporarily appending a bogus entry and running only `tests/unit/test_custom_components.py` with the env var set reported **27 "stale"** entries — i.e. every known-failure line not in that one file — not just the injected bogus one.
- `.github/workflows/test-backend.yml` only ever runs the full `tests/unit` directory, so CI itself is safe. The risk is a developer reproducing CI locally (`OPENRAG_ENFORCE_KNOWN_FAILURES=1 uv run pytest tests/unit/some_file.py`) and getting a wall of unrelated, misleading failures.
- The comment in `conftest.py` ("In CI the whole suite is collected...") states this as an assumption but the code does not enforce or check it.

#### Code References

- `tests/unit/conftest.py:87-110` (`pytest_collection_modifyitems`)

#### Potential Solution

- Guard the enforcement to only run when the full suite was requested, e.g. skip the stale check when `config.args` names a specific file/test (`config.args not in ([], ["tests/unit"], ["tests/unit/"])`), and mention the caveat in the raised message.

#### Alternative Solutions

- Document the full-suite requirement prominently next to the env var (e.g. in the `known_failures.txt` header and a `--help`-visible warning) without changing behavior — lower effort, but leaves the footgun for anyone who doesn't read the docs first.

.github/workflows/test-backend.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Install dev dependencies
+        run: uv sync --locked --group dev
+
+      - name: Run unit tests
+        env:
+          # Every entry in known_failures.txt must be collected, so a renamed or
+          # mistyped line fails loudly instead of silently protecting nothing.
+          OPENRAG_ENFORCE_KNOWN_FAILURES: '1'
+        run: uv run --locked pytest tests/unit -q --color=yes
```
mpawlow, line 62: ### (1b) [Minor] New CI workflow omits `--continue-on-collection-errors`, unlike `make test-unit`

- ℹ️ Maybe it's desirable for the CI workflow to fail fast?

#### Problem

- `.github/workflows/test-backend.yml` runs `uv run --locked pytest tests/unit -q --color=yes` directly.
- `make test-unit` deliberately adds `--continue-on-collection-errors`, with an inline comment: *"A stale module that fails to import must not abort the whole suite; the collection error is still reported and still fails the target."*
- Without that flag, pytest aborts the **entire session** on any collection error, per pytest's own default behavior (`--continue-on-collection-errors` exists specifically to change this).

#### Background Information

- Confirmed by reading `Makefile:1008-1013` vs. the new workflow's single `run:` step — they are not the same command and diverge exactly on this flag.
- The known-failures mechanism already handles all currently-known import failures via `collect_ignore`, so today this has no practical effect. But the *next* accidental import break (in a file not yet in `known_failures.txt`) would zero out the ~2000-test CI signal instead of reporting the one broken module and letting the rest run — a worse debugging experience than what `make test-unit` already guards against for exactly this scenario.

#### Code References

- `.github/workflows/test-backend.yml:62`
- Compare: `Makefile:1008-1013`

#### Potential Solution

- Add the flag: `run: uv run --locked pytest tests/unit -q --color=yes --continue-on-collection-errors`.

#### Alternative Solutions

- Have the workflow call `make test-unit` (with `--locked` folded into the target, or exported via `UV_LOCKED=1`) so the local and CI commands can't drift apart again — more idiomatic single-source-of-truth, slightly larger diff.

mpawlow (approved):
## Code Review 1

- ✅ Approved / LGTM  🚀 
   - See Normal PR comment (1a) for consideration
   - See Minor PR comment (1b) for optional consideration
- ✅ All 117 tests pass locally.
- ✅ Full unit suite (executed as CI would) with known-failures enforcement completed successfully (exit code 0)

PR #2395, opened 2026-09-17
fix: langflow content duplicate detection

### LINKED ISSUE #2388: Re-uploading identical content under a different filename silently does nothing
Re-uploading identical content under a different filename silently does nothing — no new document, no warning, no error.

## Steps to reproduce

1. Upload any file.
2. Make a byte-identical copy under a different name.
3. Upload the copy.

**Expected:** either a second document, or a clear message that this duplicates an existing one.

**Actual:** the ingest completes without error and nothing visibly happens. No new document, no warning.

## Root cause

Document identity is derived from **file content only** — the filename is never part of the hash.

`_resolve_document_id` in `src/services/langflow_file_service.py:212` ends at:

```python
if file_tuples and len(file_tuples[0]) > 1:
    content: Any = file_tuples[0][1]
    ...
    return hash_id(io.BytesIO(content))
```

Confirmed directly against that method:

```
leave-policy.pdf       -> QFS2WHnk0oRWiSHEA83nwILw
leave-policy-COPY.pdf  -> QFS2WHnk0oRWiSHEA83nwILw
SAME DOCUMENT ID: True
```

The second ingest is not rejected. It resolves to the same document id, writes the same chunk ids, and overwrites the first copy with byte-identical content. Nothing fails, so nothing errors, and nothing changes.

## Why no existing guard catches this

Two duplicate mechanisms exist, and this case falls between them.

**Filename guard** — `resolve_duplicate_filename` (`src/models/processors.py:251`, called at 776, 1167, 1494, 1601) compares names. The names differ, so it does not fire.

**Content guard** — fires on other paths, and those paths report it:

| Path | Behaviour on identical content | Reported to user? |
| --- | --- | --- |
| Traditional (`process_document_standard`) | returns `{"status": "unchanged"}` (`processors.py:497`) | Yes |
| Connector | `mark_duplicate_skipped` (`processors.py:1176`) | Yes |
| **Langflow manual upload** | **overwrites in place** | **No** |

Langflow ingest is the default (`disable_ingest_with_langflow: False`), so this is the path most users are on.

## Is the deduplication itself intended?

Most likely yes. `hash_id` already accepts an `include_filename` parameter (`src/utils/hash_utils.py:68`, implemented at :34) that would make identity name-sensitive — and **no call site passes it**. The opt-out was built and never wired, which reads as a deliberate decision that identical bytes are one document.

On that reading the defect is the silence on this one path, not the deduplication.

## Impact

Low severity, but confusing. A user re-uploading a renamed copy gets no feedback that the system recognised it, and may conclude that ingest is broken. It also makes the duplicate invisible in the audit trail — the second upload leaves no trace.

## Proposed direction

**Preferred: keep the deduplication, surface it.** Report the second upload as a duplicate of the existing document, the way the traditional and connector paths already do. No change to document identity, so low risk.

**Alternative: treat them as two documents.** Requires including the filename in the document id. The parameter exists, but the blast radius is large — identity changes, existing documents keep old ids while new uploads get different ones, and re-ingesting becomes a create rather than an update in place. Would need a migration story.

## Open questions

1. Where does the warning surface — the upload task result (like the traditional path's `"unchanged"`), a toast, or both? The frontend has no duplicate-specific state for this path today.
2. Warn-and-skip, or block with an override?
3. `replace_duplicates` is plumbed from the router (`src/api/router.py:31`, default `"true"`) into the filename guard, but `process_document_standard` does not accept it. Should it govern content duplicates too?

## Not yet confirmed

Whether the document listing ends up showing the first or the second filename, and whether the second upload increments `successful_files`. Worth confirming before wording any message, since "already uploaded as X" needs to name the right X.

## Check the collision without running the stack

```python
from services.langflow_file_service import LangflowFileService
svc = LangflowFileService.__new__(LangflowFileService)
content = open("some.pdf", "rb").read()
a = svc._resolve_document_id([("a.pdf", content, "application/pdf")], None)
b = svc._resolve_document_id([("b.pdf", content, "application/pdf")], None)
assert a == b
```

VisveshJ:
image
image

Uploading two documents with the same exact content should now not both be uploaded, a duplicate document warning now shows up. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate uploads with different filenames from being ingested when their content matches an existing document.
  * New content continues to process normally, while filename-based duplicates remain prioritized.
  * Replacing a same-name file correctly reprocesses its content.
  * Concurrent identical uploads no longer trigger repeated ingestion.

* **User Experience**
  * Duplicate files display a “Duplicate” status with an explanatory warning.
  * Upload notifications distinguish skipped, successful, and failed files.
  * Skipped files remain visible in the knowledge table after processing.


src/models/processors.py
```diff
@@ -1608,6 +1611,38 @@ async def process_item(self, upload_task: UploadTask, item: str, file_task: File
...
+            #
+            # The guard only runs on the "proceed" path (no filename match).
+            # When duplicate_action == "replaced" the caller asked to replace an
+            # existing same-name document: the old chunks were already deleted
+            # above, so we must not short-circuit here even if the hash matches
+            # (that would leave the index empty after the delete).
+            file_hash = hash_id(item)
+            file_task.document_id = file_hash
+
+            if duplicate_action == "proceed" and await self.check_document_exists(
```
jason-dong03: duplication detection is not atomic: Given a scenario where we do 2 simultaneous uploads with the same content , it can lead to both executing "check_document_exists()" before either upload has been processed/ indexed -> both uploads sees it as missing and submits the work to docling/langflow 

test file also doesn't verify this race condition (all i see is [False, True] condition) 

There needs to be some way to introduce atomic claim keyed by the content hash: 
maybe OpenSearch's 'create' with document_id = file_hash or something

jason-dong03: This is under the assumption that we do concurrent uploads (which I believe we support(?)) someone fact check me

jason-dong03: NIT: check_document_exists() also accepts an untyped client which would lead it to treating the response as an untyped dictionary

fix: type-safe cast around this path

mfortman11 (changes requested):
### P2: Clear stale skipped overlays before later ingests

`frontend/contexts/task-context.tsx:698-702`

When a content duplicate is uploaded and a later upload with the same filename succeeds, the skipped overlay remains indefinitely. Because skipped overlays take priority over active overlays, the indexed document continues to appear as “Duplicate” until the page reloads; even refreshTasks() preserves it.

Please remove or supersede skipped overlays when a newer task or indexed row exists.

  ———

### P2: Distinguish non-duplicate skipped outcomes

`frontend/contexts/task-context.tsx:412-420`

  Not every skipped file is a duplicate. Connector synchronization emits reason: "deleted_at_source" when removing a missing source (`src/models/processors.py:1126-1138`), but this mapping discards result.reason. As a result, downstream code labels every skipped row and completion toast as a duplicate.

Please preserve the skip reason and apply duplicate-specific status and messaging only to duplicate outcomes.

VisveshJ:
change the color to a lighter yellow for readibility --> future followup

mfortman11 (approved):

PR #2397, opened 2026-09-18
chore: FE query/mutation coverage

mfortman11:
Fully unit test query/mutations

The PR adds frontend tests for onboarding mutations and provider health checks, and fixes two bugs they exposed: false success when saving the docs filter ID failed and health checks running too early. The
  requested PR summary is complete.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved onboarding completion handling so settings refreshes reliably occur and success callbacks receive completed mutation details.

* **Tests**
  * Added comprehensive coverage for connector connections, synchronization, access, visibility, onboarding, settings, models, provider health, nudges, and document refresh workflows.
  * Verified request payloads, successful responses, cache invalidation, conditional execution, cancellation, and error handling across API hooks.
  * Expanded provider health checks across pending, running, and processing task states.


frontend/app/api/mutations/useOnboardingMutation.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+    );
+
+    await waitFor(() => expect(result.current.isSuccess).toBe(true));
+    expect(capturedBody).toEqual({
+      llm_provider: "openai",
+      openai_api_key: "sk-test",
+    });
+  });
+
+  it("persists openrag_docs_filter_id via the onboarding-state endpoint on success", async () => {
```
lucaseduoli, line 41: Can we also cover this `openrag_docs_filter_id` persistence path when the caller passes an `onSuccess` option?

The production onboarding card calls `useOnboardingMutation({ onSuccess })`, and the hook currently spreads `...options` after its internal `onSuccess`, so the caller callback overrides the internal persistence side effect. This test verifies the no-options path, but misses the path the app actually uses. Adding that case would catch the issue and likely means the hook should compose `options?.onSuccess` rather than letting it replace the built-in handler.

frontend/app/api/mutations/useConnectConnectorMutation.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+    queryClient.setDefaultOptions({
+      queries: {
+        retry: false,
+        staleTime: 0,
+        gcTime: 60_000,
+        refetchOnWindowFocus: false,
+      },
+      mutations: { retry: false },
+    });
+    let refetched = false;
```
lucaseduoli: `refetched` is set but never asserted, and Biome flags it as unused. Since the comment below says invalidation should not refetch without an active observer, can we remove this variable/handler to keep the test clean?

frontend/app/api/queries/useGetNudgesQuery.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+  it("includes the chatId in the request path when given", async () => {
+    let capturedUrl = "";
+    server.use(
+      http.post("/api/nudges/chat-123", ({ request }) => {
+        capturedUrl = request.url;
+        return HttpResponse.json({ response: "" });
+      }),
+    );
+
+    const { result } = renderHook(
```
lucaseduoli: `result` is unused here, which Biome flags. This can just call `renderHook(...)` without destructuring, unless we want to assert the hook state too.

lucaseduoli (changes requested):
Just minor picks

lucaseduoli (approved):

PR #2404, opened 2026-09-21
fix: detect cloud-connector duplicates ingested from other sources

ricofurtado:
Fixes the COS/bucket-connector duplicate report. Companion backport to `release-0.7.1-patch8`: #2405.

## The bug

Selecting an entire bucket for ingestion through COS (or any bucket connector) never offered the overwrite/skip prompt when the duplicate already in the knowledge base came from a **different source** — a direct upload, or another connector. The ingest then ran and every colliding file was skipped with `A file with this name already exists`.

Repro from the report: ingest a folder normally, put the same files in a bucket, configure that bucket through the COS connector, ingest the whole bucket.

## Root cause

Two duplicate definitions had drifted apart.

The pre-check that decides whether the dialog appears, `_classify_bucket_connector_duplicates`, asked only *"did THIS connector type already ingest this blob id?"*. That runs through `get_synced_file_ids_for_connector`, a `{"term": {"connector_type": ...}}` query — chunks from a direct upload carry no `connector_type` at all, so they are invisible to it. Every blob classified as new, `duplicate_count == 0`, no dialog, and the frontend went straight to a full bucket sync.

The per-file backstop in `ConnectorFileProcessor` asks a **filename** question instead, alias-expanded and across every source. With `replace_duplicates` left at its default `False`, its answer is `skip` — the warning in the report.

The OAuth connectors (Drive/OneDrive/SharePoint) were never affected: `_classify_connector_duplicates` is already filename-based. The bucket variant diverged.

## What changed

- **The bucket pre-check asks both questions,** through one shared helper (`_bucket_blob_ids_with_indexed_filename`) that the sync also uses, so the check that fills the dialog and the sync that acts on it cannot disagree. It goes through the same `find_existing_filenames` the OAuth classifier and the processor backstop already use.
- **`connector_sync` honors `replace_duplicates` on the `bucket_filter` path.** It was read only for explicit `selected_files` syncs, so overwrite could not be expressed for a whole container. Overwrite covers the blobs whose name is indexed but which this connector never synced — nothing we store about them can say whether the source is newer, so change detection cannot decide them and they are replaced outright. Blobs this connector already synced keep the modified-time/etag gate, so confirming an overwrite re-ingests the colliding files rather than every object in the container.
- **The upload UI sends bucket names + `replace_duplicates` for overwrite** instead of POSTing the whole listing back as `selected_files`. The request is now the same size for a 10-object bucket and a 10,000-object one.
- **The bucket sync passes `file_infos`.** Without it the task has only the ids, so its file rows read `supported::sample.htm` instead of `sample.htm` until each blob is fetched — visible in the report's screenshot.
- **`find_existing_filenames` batches its candidates.** A whole-container check can now feed it one filename per blob, and a `terms` query is capped at `index.max_terms_count` (65 536 by default).

## What overwrite now touches

| Blob | Overwrite confirmed |
|---|---|
| Name indexed from another source, never synced here | replaced |
| Synced here, changed at source | replaced (as before) |
| Synced here, unchanged | skipped — its indexed copy is already the current source content |
| Not indexed anywhere | ingested |

The dialog counts an unchanged already-synced blob as a duplicate but the sync skips it. Re-ingesting it would produce byte-identical chunks, so the user-visible outcome is the same and a previously-ingested container doesn't get re-processed wholesale.

## Tests

`tests/unit/api/test_connector_bucket_duplicates.py` (new, 10 tests): a filename indexed by another source is reported as a duplicate; the id half still counts; `.txt`/`.md` aliases count; a missing index is not an error; overwrite replaces only the colliding files and leaves an unchanged synced blob alone; collisions batch with changed blobs; declining the overwrite issues no filename lookup at all; an id-less listing starts no sync. Plus batching coverage in `test_opensearch_filenames.py` and `file_infos` assertions in the existing `bucket_filter` reconciliation test.

`tests/unit` passes: 2039 passed, same 8 pre-existing failures as `origin/main` in this environment (watsonx on-prem, encryption, settings provider defaults — all unrelated).

## Note

Applies to `aws_s3`, `azure_blob` and `ibm_cos`, which share `SharedBucketView` and this endpoint.

## Summary by CodeRabbit

* **New Features**
  * Improved bucket synchronization for duplicate filenames and aliases.
  * Overwrite sync now replaces only eligible conflicts while preserving unchanged files.
  * Large filename lookups are processed in batches for improved reliability.
  * Added configurable protection for ownerless shared files during duplicate replacement.

* **Bug Fixes**
  * Prevented duplicate files from being unnecessarily re-ingested.
  * Preserved accurate file metadata across synchronization batches.
  * Improved handling of buckets with no identifiable files.
  * Ensured anonymous-delete permissions are respected during connector synchronization.


frontend/components/connectors/shared-bucket-view.tsx
mpawlow, line 189: ### (1a) [Major] `connector_sync` silently ignores the requested `connection_id`

#### Problem

- `SharedBucketView` always sends an explicit `connection_id: connector.connectionId!` on every sync request (`runBucketSync` / `runSelectedFilesSync`, `frontend/components/connectors/shared-bucket-view.tsx:189,212`) — the component is rendered per-connection, so the user is targeting one specific bucket connection.
- The backend's `ConnectorSyncBody` (`src/api/connectors.py:979-1004`) has **no `connection_id` field at all**. Pydantic's default `extra="ignore"` behavior silently drops the field from the request body.
- `connector_sync` (`src/api/connectors.py:1710-1738`) instead loops `active_connections` and picks **"the first connection that actually works"** — never consulting `body.connection_id`.
- Contrast with the sibling endpoint `connector_check_duplicates` (`ConnectorCheckDuplicatesBody`, `src/api/connectors.py:1007-1008,1409-1416`), which correctly resolves `working_connection` by the requested `connection_id` first, falling back to "first working" only when unset.

#### Background Information

- With more than one active connection of the same connector type (a realistic setup for `aws_s3`/`azure_blob`/`ibm_cos` — multiple accounts/buckets), the duplicate-check call and the sync call can resolve to **different connections**. The dialog shows duplicates computed against connection A; the user confirms overwrite; the sync then runs against connection B (whichever authenticates first in list order) — syncing the wrong bucket's contents into the index while displaying misleading UI state.
- This is a pre-existing bug in `connector_sync` unrelated to this PR's *new* logic, but this PR is the one wiring `connection_id` through the frontend duplicate-check flow, and it did not extend the same fix to the actual sync call it depends on.

#### Code References

- `frontend/components/connectors/shared-bucket-view.tsx:189, 212` — sends `connection_id`
- `src/api/connectors.py:979-1004` — `ConnectorSyncBody` (no `connection_id` field)
- `src/api/connectors.py:1710-1738` — connection selection ignores the body

#### Potential Solution

Add `connection_id: str | None = None` to `ConnectorSyncBody`, and in `connector_sync` select the matching active connection first (mirroring `connector_check_duplicates`'s existing pattern at `src/api/connectors.py:1409-1416`) before falling back to "first working connection".

src/api/connectors.py
mpawlow, line 2087: ### (1b) [Minor] `connections_synced` reports task count, not connection count

#### Problem

- The `bucket_filter` overwrite path can create **two** tasks from one connection — a "new files" batch and a "changed/replaced files" batch (`src/api/connectors.py:1950-1989`).
- The final response sets `"connections_synced": len(task_ids)` (`src/api/connectors.py:2063`) while the adjacent `message` field hardcodes `"Started syncing files from 1 {connector_type} connection"` (line 2062) — the two fields contradict each other whenever both batches run, reporting `connections_synced: 2` for a sync against one connection.

#### Code References

- `src/api/connectors.py:2058-2066`

#### Potential Solution

```
"connections_synced": 1,
```

Report the connection count directly — `connector_sync` only ever operates against a single `working_connection` — rather than deriving it from `len(task_ids)`.

mpawlow, line 1074: ### (1c) [Normal] Explicit multi-file selection + overwrite has no "one winner" guard for colliding aliases

#### Problem

- The whole-container `bucket_filter` overwrite path added by this PR correctly resolves **contested filenames** among the blobs being synced: `claimed_names` ensures only the first blob in listing order claims a given indexed name when two blobs clean to the same filename (`src/api/connectors.py:1877-1903`, e.g. `a/report.pdf` and `b/report.pdf` both cleaning to `report.pdf`).
- The **explicit file-picker path** (`_classify_connector_duplicates`, `src/api/connectors.py:1068-1134`, used by `sync_specific_files` with `replace_duplicates=True` at line 1789-1800) has no equivalent guard: it only checks each selected file's aliases against filenames **already indexed**, never against aliases of other files in the **same selection**. If a user explicitly multi-selects two files that clean to the same filename but neither is yet indexed, both are classified `non_duplicate`, and both proceed into concurrent per-file ingestion with `replace_duplicates=True`.
- Per `delete_document_by_filename`'s own docstring (`src/models/processors.py:329-336`), a stale document is deleted immediately before the replacement writes — two files racing to replace the same eventual filename can each delete/replace the other's just-written chunks mid-task, a data-loss race, not just a duplicate-content nuisance.

#### Background Information

- CodeRabbit raised a related but narrower version of this in an earlier review pass ("detect aliases shared by multiple selected files"); it was fixed for the `bucket_filter` path (commit `a32447d0`, "one winner when two blobs contest an indexed filename") but the fix was not extended to `_classify_connector_duplicates` / the `selected_files` path, which is a separate code path with separate logic.

#### Code References

- `src/api/connectors.py:1068-1134` (`_classify_connector_duplicates`, no in-selection collision check)
- `src/api/connectors.py:1877-1903` (the guard that exists only for `bucket_filter`)
- `src/models/processors.py:318-348` (`delete_document_by_filename`, confirms delete-then-replace ordering)

#### Potential Solution

- In `_classify_connector_duplicates`, after computing `cleaned_files`/aliases, also detect aliases shared by two or more files in `expanded_files_info` itself (not only collisions with `existing_filenames`) and apply the same first-in-listing-order winner rule used in the `bucket_filter` path, demoting the losers to `duplicate_files` (skipped) rather than letting them race.

#### Alternative Solutions

- Serialize replacement writes for colliding filenames within a task (lock/queue per target filename) instead of picking a winner up front — more robust but a larger change to the per-file concurrency model; not proportionate for this PR.

src/api/connectors.py
```diff
@@ -1629,6 +1796,7 @@ async def connector_sync(
                 replace_duplicates=body.replace_duplicates,
                 preview_mode=preview_mode,
                 shared=body.shared,
+                allow_anonymous_delete=allow_anonymous_delete,
             )
         elif body.sync_all or body.bucket_filter:
```
mpawlow, line 1825: ### (1d) [Minor] `connector_sync`'s `bucket_filter` branch has grown into a large, multi-concern block worth extracting

#### Problem

- The `bucket_filter` branch inside `connector_sync` (`src/api/connectors.py`, roughly lines 1801-1990) now does remote listing, synced-state retrieval, filename-collision classification, batching, and task dispatch — all inline in a route handler. Per this repo's own convention ("New services: add them to `src/app/container.py`... **no router wiring** [in `api/`]"), substantial orchestration logic is expected to live in a service (`connectors/service.py`), with the route handler limited to validation/auth/response-shaping.

#### Code References

- `src/api/connectors.py:1801-1990`

#### Potential Solution

- Extract the bucket reconciliation/classification/dispatch sequence into a method on `ConnectorService` (`src/connectors/service.py`), leaving `connector_sync` to validate, call the service, and map the result to an HTTP response — consistent with how `sync_specific_files` is already delegated.

#### Alternative Solutions

- Leave as-is if the team considers this route-handler size acceptable for now (the file is already ~3,100 lines with similar patterns elsewhere) — purely a maintainability nit, not a correctness issue.

mpawlow (changes requested):
## Code Review 1

- See Major PR comment (1a) for evaluation
- See Normal PR comment (1c) for consideration
- See Minor PR comments (1b) (1d) for optional consideration

frontend/components/connectors/shared-bucket-view.tsx
ricofurtado: The issue depends on "more than one active connection of the same connector type (a realistic setup… multiple accounts/buckets)". I went looking for a way to produce that state and couldn't:

- create_connection reuses the existing active connection for that type+user, and after inserting calls cleanup_duplicate_connections(remove_duplicates=True).
- The load path calls the same cleanup after reading connections.json, so legacy or hand-edited duplicates are destroyed at startup rather than loaded.
- activate_connection — the one route that could resurrect a second active connection has no callers
- Multiple buckets live in one connection via bucket_names/bucket_filter. That's why bucket_filter exists

So the list this code walks holds at most one entry, and the described failure: dialog computed against A, sync runs B can't occur today. This is a consistency and defense-in-depth fix, not a live wrong-bucket bug. My earlier "inert" call was right on reachability, what it got missing was that an API silently ignoring a parameter it receives is worth fixing.

src/api/connectors.py
ricofurtado: That was fixed in the new PR I've created.

frontend/components/connectors/shared-bucket-view.tsx
mpawlow: Verified

src/api/connectors.py
mpawlow: Acknowledged.

mpawlow: Finding was re-discovered during **Code Review 2**

### (2a) [Major] Explicit multi-select overwrite has no same-batch collision guard for colliding filenames

#### Problem

- `_classify_connector_duplicates` (`src/api/connectors.py:1074`) — the classifier behind the **explicit file-picker** path (`sync_specific_files` with `replace_duplicates=True`) — only checks each selected file's cleaned-filename aliases against `existing_filenames` already in the index.
- It never checks a selected file's aliases against the **other files in the same selection**. If a user explicitly multi-selects two files that clean to the same target filename (e.g. two blobs both named `report.pdf` in different folders) and neither is indexed yet, both are classified `non_duplicate` and both proceed into **concurrent** per-file ingestion with `replace_duplicates=True`.
- Per `delete_document_by_filename`'s own docstring (`src/models/processors.py:318-336`), a stale document is deleted immediately before the replacement write — two files racing for the same eventual filename can each delete/replace the other's just-written chunks mid-task. This is a data-loss race, not a duplicate-content nuisance.
- The sibling whole-container path (`bucket_filter`, added by this PR) already has this guard: `claimed_names` in `connector_sync` (`src/api/connectors.py:1902-1932`) picks a single first-in-listing-order winner per contested indexed name before dispatching. The explicit-selection path was not given the equivalent guard.

#### Background Information

- CodeRabbit raised the general form of this in its first-pass review ("Reject alias collisions within one selected bucket set", still unresolved on the PR, `src/api/connectors.py` around the classifier). It was fixed for the `bucket_filter` path (commit `a32447d0`, "one winner when two blobs contest an indexed filename") but the fix was never extended to `_classify_connector_duplicates` — a separate function with separate logic.
- This exact gap was also raised in a prior human review round on this PR (finding `1c`) and confirmed by the author (`ricofurtado`) as accurate; the author's reply states it was addressed in a **new, independent PR**.
- Verified: PR **#2406** (`fix: two files heading for one filename no longer merge into one document`, branch `fix/same-name-files-in-one-batch`, currently `OPEN`) closes this race generically — for uploads, OAuth connectors, and bucket connectors alike — via an in-process filename-claim registry (`src/utils/filename_claims.py`). Its description explicitly frames itself as "independent of #2404; no overlapping files."
- Net effect: **as merged today, PR #2404 alone does not close this race for its own explicit-selection code path.** The risk is real but is mitigated to the extent #2406 lands (independently, in the same release train). Reviewers should not assume #2404 alone is sufficient protection against this failure mode.

#### Code References

- `src/api/connectors.py:1074-1140` (`_classify_connector_duplicates` — no in-selection collision check)
- `src/api/connectors.py:1902-1932` (the equivalent guard that exists only for `bucket_filter`)
- `src/models/processors.py:318-348` (`delete_document_by_filename`, confirms delete-then-replace ordering)
- Companion fix (separate PR): [#2406](https://github.com/langflow-ai/openrag/pull/2406)

#### Potential Solution

- In `_classify_connector_duplicates`, after computing `cleaned_files` / aliases, also detect aliases shared by two or more files in `expanded_files_info` itself (not only collisions with `existing_filenames`), and apply the same first-in-listing-order winner rule used in the `bucket_filter` path — demoting the losers to `duplicate_files` (skipped) instead of letting them race.

frontend/components/connectors/shared-bucket-view.tsx
```diff
@@ -177,9 +176,12 @@ export function SharedBucketView({
   };
 
   // Full bucket sync: backend classifies new/changed/unchanged itself and
-  // re-ingests both new and changed blobs (used both when there are no
-  // duplicates and when the user chooses to overwrite them).
-  const runBucketSync = () => {
+  // re-ingests both new and changed blobs. With `replaceDuplicates`, it instead
```
mpawlow: ### (2b) [Normal] "Skip duplicates" still round-trips the full non-duplicate listing; the PR's stated request-size fix only covers the overwrite branch

#### Problem

- The PR description states the request "is now the same size for a 10-object bucket and a 10,000-object one," attributing this to sending `bucket_names` + `replace_duplicates` instead of the whole listing. This is only true for the **overwrite** path.
- `handleOverwriteDuplicates` (`frontend/components/connectors/shared-bucket-view.tsx:287-291`) does call `runBucketSync(true)`, which sends only `bucket_filter` + `replace_duplicates: true` — correctly bounded.
- The **skip-duplicates** path (dialog dismissed without confirming overwrite) still calls `runSelectedFilesSync(nonDuplicateFiles)` (`frontend/components/connectors/shared-bucket-view.tsx:294-306`), POSTing the *entire* `non_duplicate_files` array from the classify response back as `selected_files`.
- For a large container where only a handful of files collide, `non_duplicate_files` can be the vast majority of the listing (e.g. 9,995 of 10,000), so "skip" still sends a listing-sized payload — the exact scaling concern the PR's own narrative says was solved.

#### Code References

- `frontend/components/connectors/shared-bucket-view.tsx:176-183` (PR's own comment describing the size-parity claim)
- `frontend/components/connectors/shared-bucket-view.tsx:206-224` (`runSelectedFilesSync`, unchanged for the skip path)
- `frontend/components/connectors/shared-bucket-view.tsx:294-306` (`handleDuplicateDialogOpenChange`, skip branch)

#### Potential Solution

- Give the skip path the same shape as the overwrite path: send `bucket_filter` + a `replace_duplicates: false` intent and let the backend re-derive new/duplicate blobs server-side (it already computes this for the overwrite branch), rather than shipping the client-held `non_duplicate_files` list back verbatim.

#### Alternative Solutions

- If keeping the explicit-list shape for "skip" is preferred (e.g. to avoid a second server-side classification pass), cap or paginate the listing sent to `check-duplicates` and `selected_files` together so neither side ever materializes an unbounded array — a larger change, not necessary if the primary solution is adopted.

src/api/connectors.py
mpawlow: Still applicable in **Code Review 2**

### (2c) [Minor] `connections_synced` still reports task count, not connection count

#### Problem

- The `bucket_filter` overwrite path can create **two** tasks from one connection — a "new files" batch and a "changed/replaced files" batch.
- The final response sets `"connections_synced": len(task_ids)` (`src/api/connectors.py:2087`) while the adjacent `message` field hardcodes `"Started syncing files from 1 {connector_type} connection"` (`src/api/connectors.py:2085`) — the two fields contradict each other whenever both batches run, reporting `connections_synced: 2` for a sync against exactly one connection.
- This was already flagged in a prior review round on this PR (finding `1b`) and remains unfixed; the GitHub review thread is still open (unresolved, not outdated — the surrounding code has not changed).

#### Code References

- `src/api/connectors.py:2078-2088`

#### Potential Solution

- Report the connection count directly (`connector_sync` only ever operates against a single `working_connection`) rather than deriving it from `len(task_ids)`:
  ```
  "connections_synced": 1,
  ```

src/api/connectors.py
mpawlow: Still applicable in **Code Review 2**

### (2d) [Minor] `connector_sync`'s `bucket_filter` branch remains a large, multi-concern inline block

#### Problem

- The `bucket_filter` branch inside `connector_sync` (`src/api/connectors.py:1825` onward, now materially larger after this PR's additions) performs remote listing, synced-state retrieval, filename-collision classification, batching, and task dispatch — all inline in a route handler.
- This repo's own conventions (`AGENTS.md` / `CLAUDE.md`, "New services... **no router wiring** [in `api/`]"; CodeRabbit's path instructions: "No business logic in route handlers") expect substantial orchestration logic to live in a service (`src/connectors/service.py`), with the route limited to validation/auth/response-shaping — the pattern `sync_specific_files` on `ConnectorService` already follows.
- Both CodeRabbit and a prior human review round flagged this (finding `1d`); it remains unaddressed. The review thread is open and not outdated, confirming no code movement occurred here since it was raised.

#### Code References

- `src/api/connectors.py:1825-2050` (approximate span of the `bucket_filter` branch)

#### Potential Solution

- Extract the bucket reconciliation/classification/dispatch sequence into a method on `ConnectorService` (`src/connectors/service.py`), leaving `connector_sync` to validate, call the service, and map the result to an HTTP response.

#### Alternative Solutions

- Leave as-is if the team considers this route-handler size acceptable for now (the file is already ~2,800+ lines with similar patterns elsewhere) — a maintainability nit, not a correctness issue, and the lowest-priority item in this review.

mpawlow (approved):
## Code Review 2

- ✅ (Conditional) Approval / LGTM  🚀 
- See PR comments:
   - 🔴 Major: (2a)
      - Will be addressed in the follow-on PR: https://github.com/langflow-ai/openrag/pull/2406
   - 🟠 Normal: (2b)
   - 🟡 Minor: (2c) (2d)

PR #2406, opened 2026-09-22
fix: two files heading for one filename no longer merge into one document

ricofurtado:
Follow-up to #2404, which fixed the same class of race for one bucket-connector path. This closes it everywhere else — uploads and the OAuth connectors — at the seam all ingestion shares. Independent of #2404; no overlapping files.

## The bug

Files within a task are ingested **concurrently** (`TaskService` runs them under a worker semaphore, up to 4). The duplicate gate's question : *"is this name already indexed?"* is answered by OpenSearch, which cannot see a file that has not been written yet. So two files destined for the same indexed filename both pass the gate:

- an upload of two `report.pdf` from different folders (`list[UploadFile]`, no name de-dup);
- a Drive/OneDrive/SharePoint selection spanning folders;
- two keys under one bucket prefix — a bucket connector names a blob after its key's basename, so `a/report.pdf` and `b/report.pdf` are both `report.pdf`.

What happened next was silent, and depended on `replace_duplicates`:

| | Before |
|---|---|
| `replace_duplicates=false` | Chunk ids are content-derived (`_scoped_chunk_id`), so neither file overwrote the other: **one filename holding two documents' chunks**. The knowledge view shows a single row; deleting it takes both down. |
| `replace_duplicates=true` (the upload default) | The second file **deleted the first one's chunks mid-flight** and indexed itself over them. The first file's task still reported success. |

Neither is reachable by inspecting the index afterwards in the usual way, because at the moment each file checked, the other had not written yet.

## The fix

`src/utils/filename_claims.py` — a small in-process registry that holds the name a file is about to be indexed under, for the duration of that file's ingestion. The second file then resolves as an ordinary duplicate and takes the existing `A file with this name already exists` skip, counted successful, exactly as a duplicate-against-the-index does today.

Placement matters more than the mechanism:

- **Taken** in `TaskProcessor.resolve_duplicate_filename` — the one duplicate policy all four processors (document, connector, S3, Langflow) already share, so no path can forget it.
- **Released** in the `finally` of `TaskService.background_custom_processor`'s per-file wrapper — the single point every file passes through, terminal state or not, retries included (retry re-enters the same driver with the same holder key).
- **Aliases** are claimed together, so `notes.txt` and `notes.md` contest one claim — the same pair `get_filename_aliases` already unifies for duplicate checks and deletes.
- **Scoped by ownership**, mirroring `_scoped_chunk_id` and the DLS rules the duplicate check runs under: another user's private document with the same name is not a duplicate and is invisible to the check anyway.

## Scope and limits

- **In-process only.** Same single-worker assumption the RBAC and identity caches already make (AGENTS.md); with multiple workers, two files landing in different processes fall back to today's behaviour. Redis would be the fix, alongside the caches.
- **Which file wins is still scheduling-dependent** — the first to claim. Making the winner deterministic would mean de-duplicating at each entry point (upload, connector selection); this change makes the *outcome* correct and visible everywhere instead, in one place.
- A private and a shared ingest of the same name by one user remain in different claim scopes. The index check treats them as colliding, so that pair keeps its existing behaviour; untangling it is a separate question about what shared-vs-private duplicates should mean.

## Tests

- `tests/unit/test_same_name_files_in_one_batch.py` — two same-named files ingested concurrently: one completes, one is SKIPPED with the duplicate warning and no failure; with replace on, the loser no longer deletes the winner's chunks; aliases behave the same; differently-named files still both ingest (the claim must not serialize an ordinary batch). The first three fail without the claim and pass with it — verified by reverting it.
- `tests/unit/test_filename_claims.py` — registry semantics: first holder wins, release frees the name, a holder can re-claim its own name (retries), aliases contest one claim, other owners don't.
- `tests/unit/conftest.py` — autouse fixture clearing the registry per test, since tests drive `process_item` directly and never pass through TaskService's release.

`tests/unit`: 2046 passed, same 8 pre-existing failures as clean `origin/main` in this environment (watsonx on-prem, encryption, bulk retry, processor mapping, settings provider defaults).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrently processed files with the same filename, including recognized aliases, from being ingested more than once.
  * Conflicting files are now skipped as duplicate filenames with appropriate status and warning details.
  * Improved duplicate handling when replacement is enabled, preventing the later file from removing the earlier file’s content.
  * Files with different names continue to process concurrently.


src/models/processors.py
```diff
@@ -268,7 +270,21 @@ async def resolve_duplicate_filename(
...
                              ingestion can continue.
+
+        ``claim_holder`` identifies this file's run (see ``utils.filename_claims``)
+        and holds the name for its duration, so a second file heading for the
+        same name inside the same batch resolves as a duplicate instead of
+        racing the index check. TaskService releases the claim when the file
+        reaches a terminal state. Callers that pass None get the OpenSearch
+        policy alone — every processor passes one.
         """
+        if claim_holder is not None and not filename_claims.claim(
```
mpawlow, line 281: ### (1a) [Major] Winner-fails-after-claim leaves the loser stranded as a false "success"

#### Problem

- The claim is taken, and the loser's fate is decided, **before the winner's ingestion outcome is known**.
- If the winning file later fails (corrupt PDF, extraction error, timeout), the losing file has already been finalized as `SKIPPED` and counted in `successful_files` — but nothing was ever indexed under that filename.
- `SKIPPED` is excluded from `retry_failed_files`'s candidate filter (`file_task.status != TaskStatus.FAILED` → rejected as `"not_failed"`), so the loser can never be retried automatically.
- Net result: a task can report "1 failed, 1 successful" for a `report.pdf` pair while **zero documents were actually ingested**, with no automated path to recover — the user must notice and manually re-upload.

#### Background Information

- Pre-PR, this scenario had no such failure mode: each file independently raced OpenSearch, so a corrupt file's failure never blocked the other file's success (that was the bug this PR fixes for the *opposite* case — both succeeding).
- The PR's own "Scope and limits" section addresses *which* file wins ("still scheduling-dependent... this change makes the outcome correct and visible everywhere instead") but does not address *what happens when the winner subsequently fails* — a materially different, unaddressed failure mode introduced by resolving the race eagerly at claim time rather than at completion time.
- `tests/unit/test_same_name_files_in_one_batch.py` only exercises the happy path where the winner succeeds; there is no test where the winner fails after claiming.

#### Code References

- `src/models/processors.py:281-286` — claim taken and loser's outcome (`"skip"`) decided synchronously, before either file's processing has run.
- `src/models/processors.py:323-335` (`mark_duplicate_skipped`) — the loser is finalized `TaskStatus.SKIPPED` and `upload_task.successful_files += 1`, independent of the winner's eventual outcome.
- `src/services/task_service.py:1015`, `:1029` (`retry_failed_files`) — retry candidates are filtered to `TaskStatus.FAILED` only; a `SKIPPED` file is rejected with `"reason": "not_failed"`.

#### Potential Solution

- On the winner's terminal failure, have `TaskService`'s per-file `finally` block (`task_service.py:757-762`, alongside the claim release) check whether any other file in the same task was `SKIPPED` due to a claim conflict on this exact filename/scope, and requeue **one** of them for processing instead of leaving it as a dead-end success.

#### Alternative Solutions

- Simpler and more consistent with the rest of the codebase's terminal-state model: don't finalize the loser immediately. Have it block on the winner's outcome (e.g. an `asyncio.Event` per claimed name, stored alongside the holder) and only resolve to `"skip"` (real duplicate) or retry-as-primary (winner failed) once the winner reaches a terminal state.
- Cheapest fix if the above is judged out of scope for this PR: surface the risk explicitly — log a warning distinguishing "skipped: lost an in-flight race" from "skipped: pre-existing duplicate" in `file_task.result`, and mention the gap under "Scope and limits" in the PR description so it's a known, documented trade-off rather than a silent one.

src/models/processors.py
mpawlow, line 995: ### (1b) [Major] `_reconcile_shared_owner` can apply the wrong `shared`/`owner` state to the winner's chunks when the "skip" is a claim race, not a real duplicate

#### Problem

- `ConnectorFileProcessor.process_item` calls `_reconcile_shared_owner(file_task.filename, shared)` whenever `resolve_duplicate_filename` returns `"skip"` (`processors.py:1198`, `:1236`) — a call designed for the case "this file's content and name haven't changed since a prior sync," where `shared` reflects *this same document's* previously-indexed state.
- This PR adds a **new** reason to reach `"skip"`: an in-flight claim conflict with a *different* connector document that happens to target the same filename (`processors.py:281-286`).
- On the re-sync path (`self.shared is None`), `_resolve_shared` falls back to `_indexed_shared_state(file_id, ...)`, keyed by **this file's own** connector `file_id` — a different value per racing file. Two different remote objects that happen to collide on filename can have genuinely different previously-recorded `shared` states.
- `_reconcile_shared_owner`'s `update_by_query` is scoped only by **filename** + "owned by this user or ownerless" (`build_replace_filename_query`, `opensearch_queries.py:102-117`) — not by document id — so if the winner has already written chunks by the time the loser's reconcile call runs, the loser's independently-resolved `shared` value can flip the *winner's* `owner`/`owner_email`/`owner_name` fields to the wrong value.

#### Background Information

- Because DLS treats an ownerless chunk as visible instance-wide, an incorrect flip in either direction is a real access-control consequence, not just cosmetic metadata drift — clearing `owner` on a chunk that should stay private exposes it; setting an owner on a chunk that should stay shared un-shares it.
- The window is narrow (requires the reconcile call to land after the winner has written at least one chunk) and the precondition specific (re-sync path, two connector items colliding on filename, differing resolved `shared` state) — hence Major rather than Critical, and confidence pinned at the 60 threshold rather than higher.
- Only `ConnectorFileProcessor` calls `_reconcile_shared_owner` on skip; `DocumentFileProcessor`, `S3FileProcessor`, and the Langflow-ingest processor do not, so this is scoped to the connector-sync path.

#### Code References

- `src/models/processors.py:995-1050` (`_reconcile_shared_owner`) — its docstring's assumption ("a file's content and name haven't changed since a prior sync") no longer covers every caller of this method.
- `src/models/processors.py:1187-1198`, `:1236` — call sites, immediately following the new claim-based `"skip"`.
- `src/models/processors.py:971-993` (`_resolve_shared`) — per-file, per-`file_id` resolution that can diverge between two racing documents.
- `src/utils/opensearch_queries.py:102-117` (`build_replace_filename_query`) — the filename-scoped (not document-scoped) match used by the reconcile query.

#### Potential Solution

- Skip the `_reconcile_shared_owner` call when the `"skip"` was caused by a claim conflict rather than a real OpenSearch duplicate. `resolve_duplicate_filename` already knows which branch produced `"skip"` (`processors.py:281-286` vs. `:291`) — thread that distinction out (e.g. a more granular return value, or a separate `Literal["proceed", "skip_duplicate", "skip_claimed", "replaced"]`) so callers can choose not to reconcile ownership for a file that was never actually indexed.

#### Alternative Solutions

- Scope `_reconcile_shared_owner`'s `update_by_query` to the specific document instead of by filename alone, if a document identifier is available at that point — narrower but requires plumbing more context through.

tests/unit/test_same_name_files_in_one_batch.py
```diff
@@ -0,0 +1,119 @@
...
+import asyncio
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from models.processors import DUPLICATE_FILENAME_WARNING, DocumentFileProcessor
+from models.tasks import FileTask, TaskStatus, UploadTask
+
+
+def _processor(replace_duplicates: bool) -> DocumentFileProcessor:
```
mpawlow, line 25: ### (1d) [Minor] New concurrent-race regression test covers only `DocumentFileProcessor`

#### Problem

- `tests/unit/test_same_name_files_in_one_batch.py` — the integration-style test proving the fix actually works under concurrency — only exercises `DocumentFileProcessor` (the upload path).
- The PR description explicitly claims the fix "closes it everywhere else — uploads **and the OAuth connectors**" and calls out "two keys under one bucket prefix" (S3) as a motivating scenario, but `ConnectorFileProcessor`, `S3FileProcessor`, and the Langflow-ingest processor — which all wire `claim_holder=self._claim_holder(...)` into `resolve_duplicate_filename` — have no equivalent concurrent-race test.
- `tests/unit/test_filename_claims.py` tests the registry primitive in isolation, which is necessary but not sufficient to prove each processor is wired correctly end-to-end.

#### Code References

- `tests/unit/test_same_name_files_in_one_batch.py:25-119` — all four tests instantiate only `DocumentFileProcessor`.
- `src/models/processors.py:1195` (connector), `:1522` (S3), `:1630` (Langflow) — the other three call sites with no matching concurrency test.

#### Potential Solution

- Parametrize `test_same_name_files_in_one_batch.py`'s tests over the four processor classes (or add one representative concurrent-race test per processor), matching the breadth the PR description claims.

src/utils/filename_claims.py
```diff
@@ -0,0 +1,123 @@
...
+# One registry per process: ingestion is driven by the singleton TaskService.
+filename_claims = FilenameClaimRegistry()
+
+
+def claim_holder(task_id: str, file_key: str) -> str:
+    """Stable identity for one file's run within one task."""
+    return f"{task_id}:{file_key}"
+
+
+def release_claims(task_id: str, file_keys: Iterable[str]) -> None:
```
mpawlow, line 121: ### (1e) [Minor] `release_claims()` is dead code

#### Problem

- `release_claims(task_id, file_keys)` is defined in `src/utils/filename_claims.py` but has zero callers anywhere in the repository.
- The actual release path in `TaskService` calls `filename_claims.release(claim_holder(task_id, item_key))` directly, inline, duplicating what `release_claims` already does for a single key.
- It adds an untested, unused public entry point that a future reader may assume is the wired-in cleanup path when it isn't, and pulls in the `Iterable` import solely to support its own signature.

#### Code References

- `src/utils/filename_claims.py:121-123` (`release_claims`) — unused.
- `src/services/task_service.py:762` — the actual (duplicate, inline) release call.

#### Potential Solution

- Delete `release_claims()` and the now-unnecessary `Iterable` import, or replace `task_service.py:762`'s inline call with `release_claims(task_id, [item_key])` so the helper has a real caller instead of being dead code.

src/utils/filename_claims.py
```diff
@@ -0,0 +1,123 @@
...
+
+from collections.abc import Iterable
+
+from utils.file_utils import get_filename_aliases
+from utils.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+
+def claim_scope(owner_user_id: str | None, shared: bool) -> str:
```
mpawlow, line 42: ### (1f) [Minor] `claim_scope()` reimplements a scope-string format that already exists elsewhere

#### Problem

- `claim_scope()` (`filename_claims.py:42-48`) computes the same `"shared"` / `f"owner:{owner_user_id}"` scope-string shape that `DocumentIndexWriter._scoped_chunk_id` already computes for chunk ids and DLS scoping — the PR's own docstrings say this function is deliberately "mirroring" that format rather than reusing it.
- Because the two are independent copies, a future change to the sentinel format in one (e.g. to avoid `"shared"` colliding with a literal owner id of `"shared"`) is easy to make without updating the other, silently desyncing claim scoping from the chunk-id/DLS scoping it's supposed to mirror — reintroducing exactly the kind of cross-user collision or missed-collision bug this PR exists to prevent.

#### Code References

- `src/utils/filename_claims.py:42-48` (`claim_scope`).
- `src/services/document_index_writer.py:179-183` (`_scoped_chunk_id`, referenced by name in `filename_claims.py`'s own module docstring, line 25).

#### Potential Solution

- Extract the scope-string format into one small shared helper (e.g. in a module both `filename_claims.py` and `document_index_writer.py` can import) and have both call it, rather than maintaining two docstring-linked but independently-implemented copies.

#### Alternative Solutions

- If a shared import isn't feasible for layering reasons, at minimum add a unit test asserting `claim_scope(...)` and `_scoped_chunk_id`'s scope portion produce identical strings for the same inputs, so a future divergence fails loudly instead of silently.

mpawlow (changes requested):
## Code Review 1

- See PR comments:
  - 🔴 Major: (1a) (1b)
  - 🟡 Minor: (1d) (1e) (1f)
    - (optional consideration)

